Issue #348, opened 2022-03-20
Add "cost" hook in tap base


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/350

Originally created by @LaurentS on 2022-03-20 17:19:25

---

## Summary
As a user of a tap, I would like to know how much API "cost" was caused by a tap run.
For instance, the github API has per-hour usage limits, where 1 REST API call costs 1, or a graphql API call has a cost that depends on the number of nodes returned.
Other APIs might charge per call.

At the end of a tap run, I would like to know how much of the resource the tap consumed.

## Proposed benefits

With `tap-github` running in production, we have found it hard to track the reasons behind sudden surges in "quota exceeded" errors. Being able to identify which runs use what would help in understanding the cause of such issues.

For billable APIs/resources, such a feature could also help track actual dollar costs.

Making these values retrievable would allow tracking them in monitoring systems, etc...

## Proposal details
As it's not really possible to define how this cost is calculated at the SDK level, it would be great if the SDK provided some method that a tap could overload to calculate and accumulate said cost, something like:

```python
class Tap:
    def calculate_request_cost(self, request, response) -> Integer:
        # return whatever cost we want to track in arbitrary units
```
This could be called by the SDK after each request returns (as the cost might depend on the content of the response). The SDK would simply keep a sum of all these results, and at the end of the run, the tap would at a minimum log a line like:
```
Total cost for this run: NNN
```
or possibly export this value in the state as another metric?

Again, the final method could simply be a no-op by default that each tap can override to implement appropriate behaviour.

## Best reasons not to build
By default, these methods can simply do a `pass`, and therefore have only negligible performance impact. There is no behaviour change related to this, unless the result is exported in state, which might cause issues downstream depending on the target used.

MeltyBot:
[View 4 previous comments from the original issue on GitLab](https://gitlab.com/meltano/sdk/-/issues/350)

laurentS:
Just noting that I'm going to work on this

aaronsteers:
@laurentS - Sounds great! Are you okay with a simple log line printed or were you thinking about emitting another machine-readable manner? 

In theory, Singer metrics could also be used and could be more machine-readable. https://hub.meltano.com/singer/spec#metrics

aaronsteers:
@laurentS - The one spec consideration is to perhaps introduce a domain label for the cost and to make this support `n` number of cost domains.

[...]

Issue #383, opened 2022-05-18
Reduce noise from 'ignored properties' warnings


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/386

Originally created by @aaronsteers on 2022-05-18 03:42:17

---

The SDK prints warnings for every field that the API returns but is not mapped to the stream's `schema`. We can do a couple things to reduce the noisiness of logs.

## Consolidate warnings into a single message per stream

Instead of:

```
2022-05-18T03:26:54.596749Z [info     ] time=2022-05-18 03:26:54 name=tap-gitlab level=WARNING message=Property 'closed_by' was present in the 'issues' stream but not found in catalog schema. Ignoring. cmd_type=extractor job_id=gitlab-to-postgres name=tap-gitlab run_id=8308d041-475f-4404-be2c-e4012413e93a stdio=stderr
2022-05-18T03:26:54.597507Z [info     ] time=2022-05-18 03:26:54 name=tap-gitlab level=WARNING message=Property 'milestone' was present in the 'issues' stream but not found in catalog schema. Ignoring. cmd_type=extractor job_id=gitlab-to-postgres name=tap-gitlab run_id=8308d041-475f-4404-be2c-e4012413e93a stdio=stderr
2022-05-18T03:26:54.598140Z [info     ] time=2022-05-18 03:26:54 name=tap-gitlab level=WARNING message=Property 'author' was present in the 'issues' stream but not found in catalog schema. Ignoring. cmd_type=extractor job_id=gitlab-to-postgres name=tap-gitlab run_id=8308d041-475f-4404-be2c-e4012413e93a stdio=stderr
2022-05-18T03:26:54.598749Z [info     ] time=2022-05-18 03:26:54 name=tap-gitlab level=WARNING message=Property 'type' was present in the 'issues' stream but not found in catalog schema. Ignoring. cmd_type=extractor job_id=gitlab-to-postgres name=tap-gitlab run_id=8308d041-475f-4404-be2c-e4012413e93a stdio=stderr
2022-05-18T03:26:54.599366Z [info     ] time=2022-05-18 03:26:54 name=tap-gitlab level=WARNING message=Property 'assignee' was present in the 'issues' stream but not found in catalog schema. Ignoring. cmd_type=extractor job_id=gitlab-to-postgres name=tap-gitlab run_id=8308d041-475f-4404-be2c-e4012413e93a stdio=stderr
[...]

PR #704, opened 2022-06-08
feat(taps): Add api costs hook

laurentS:
This PR addresses #348 and adds a simple callback mechanism to the sdk that is:
- totally optional: taps won't see any difference in behaviour (not even logging output) until they implement `calculate_api_request_cost` in their streams.
- flexible: each tap can aggregate cost along any number of dimensions/domains. The only constraint is to provide all dimensions on each request (even if associated cost is 0)
- easy to implement: see https://github.com/MeltanoLabs/tap-github/pull/142
- untested :sweat_smile: I'm not sure what the best way to test it is (the implementation in tap-github kind of does that, but it would be good to cover it in the sdk itself).
- undocumented :see_no_evil: I'll add that once we're all happy with the code

Here is an example of output (at the very end of the tap run) for `tap-github` with the implementation linked above:
```
time=2022-06-08 12:24:42 name=tap-github level=INFO message=Total API costs for stream languages: {'rest': 2, 'graphql': 0, 'search': 0}
time=2022-06-08 12:24:42 name=tap-github level=INFO message=Total API costs for stream repositories: {'rest': 2, 'graphql': 0, 'search': 0}
time=2022-06-08 12:24:42 name=tap-github level=INFO message=Total API costs for stream stargazers: {'rest': 0, 'graphql': 7, 'search': 0}
```

On the downside, the logging above could be improved for better machine handling. `log_api_costs` could be overridden by tap developers for this. And the way I call the logging method at the end feels a bit hacky, but I couldn't think of a better way to do it.

### DIFF
--- a/tests/core/test_streams.py
+++ b/tests/core/test_streams.py
@@ -1,5 +1,6 @@
 """Stream tests."""
 
+import logging
 from typing import Any, Dict, Iterable, List, Optional, cast
 
 import pendulum


laurentS:
@aaronsteers @edgarrmondragon it looks like the CI setup is not fully there yet for PRs, the job on python 3.10 (py, external) complains about missing credentials for `tap-google-analytics`, and the `pre-commit` one cannot find poetry. I'm not sure how to fix that last one.

edgarrmondragon:
@laurentS Don't worry about the pre-commit check. The external tests can be retried by an admin (I'll look into making it wait for approval instead of just failing)

singer_sdk/streams/core.py
```diff
@@ -854,6 +857,18 @@ def _write_request_duration_log(
...
+        """Log a summary of API costs.
+
+        The costs are calculated via `calculate_api_request_cost`.
+        This method can be overridden to log results in a custom
+        format. It is only called once at the end of the life of
+        the stream.
+        """
+        if len(self._api_costs) > 0:
+            msg = f"Total API costs for stream {self.name}: {self._api_costs}"
+            self.logger.info(msg)
```
edgarrmondragon, line 870:
```suggestion
    def log_sync_costs(self) -> None:
        """Log a summary of Sync costs.

        The costs are calculated via `calculate_sync_cost`.
        This method can be overridden to log results in a custom
        format. It is only called once at the end of the life of
        the stream.
        """
        if len(self._sync_costs) > 0:
            msg = f"Total Sync costs for stream {self.name}: {self._sync_costs}"
            self.logger.info(msg)
```

singer_sdk/streams/core.py
```diff
@@ -76,6 +76,9 @@ class Stream(metaclass=abc.ABCMeta):
     parent_stream_type: Optional[Type["Stream"]] = None
     ignore_parent_replication_key: bool = False
 
+    # Internal API cost aggregator
+    _api_costs: Dict[str, int] = {}
```
edgarrmondragon, line 80:
```suggestion
    _sync_costs: Dict[str, int] = {}
```

singer_sdk/streams/rest.py
```diff
@@ -320,6 +320,7 @@ def request_records(self, context: Optional[dict]) -> Iterable[dict]:
                 context, next_page_token=next_page_token
             )
             resp = decorated_request(prepared_request, context)
+            self.update_api_costs(prepared_request, resp, context)
```
edgarrmondragon, line 323:
```suggestion
            self.update_sync_costs(prepared_request, resp, context)
```

singer_sdk/streams/rest.py
```diff
@@ -333,8 +334,64 @@ def request_records(self, context: Optional[dict]) -> Iterable[dict]:
...
+        Returns:
+            A dict of costs (for the single request) whose keys are
+            the "cost domains". See `calculate_api_request_cost` for details.
+        """
+        call_costs = self.calculate_api_request_cost(request, response, context)
+        self._api_costs = {
+            k: self._api_costs.get(k, 0) + call_costs.get(k, 0)
+            for k in call_costs.keys()
+        }
+        return self._api_costs
```
edgarrmondragon, line 359:
```suggestion
        call_costs = self.calculate_api_request_cost(request, response, context)
        self._sync_costs = {
            k: self._sync_costs.get(k, 0) + call_costs.get(k, 0)
            for k in call_costs.keys()
        }
        return self._sync_costs
```

singer_sdk/tap_base.py
```diff
@@ -379,6 +379,9 @@ def sync_all(self) -> None:
             stream.sync()
             stream.finalize_state_progress_markers()
 
+        for stream in self.streams.values():
+            stream.log_api_costs()
```
edgarrmondragon, line 383:
```suggestion
            stream.log_sync_costs()
```

edgarrmondragon (changes requested):
@laurentS I've added a few suggestions to rename `api` to `sync` in the base class, since I can imagine a more generic cost calculation for files, databases, etc.

Other points that might be good to start a discussion on:

1. **Which logger should be used**. This is using the tap/stream logger, but I wonder if it makes sense to use a different one so metrics and tap behavior are not mixed in logs (releated to https://github.com/meltano/sdk/issues/383#issuecomment-1154701837)

1. Adding some tests to validate cost calculation in the general case.

laurentS:
> @laurentS I've added a few suggestions to rename `api` to `sync` in the base class, since I can imagine a more generic cost calculation for files, databases, etc.
Ok, makes sense, I've applied your suggestions. Thanks for writing them! I had gone with `api` since it was all network requests, but it's true the requests could be to anything.
 
> Other points that might be good to start a discussion on:
> 
>     1. **Which logger should be used**. This is using the tap/stream logger, but I wonder if it makes sense to use a different one so metrics and tap behavior are not mixed in logs (releated to [#383 (comment)](https://github.com/meltano/sdk/issues/383#issuecomment-1154701837))

From my perspective, this would be much better. Right now, the costs are logged at `info` level, but there is a lot of logging that I have no interest in at that same level. I hesitated to make the costs `warning`s but semantically, that feels wrong. I like your approach of splitting behaviour logging from metrics (though in general I have no use for the general metrics telling me how long each network request took, or the number of records for each of them). I guess we could add something like the `_metric_logging_function` to allow users to set the level they want for various parts of logging (behaviour, metrics, costs?).

>     2. Adding some tests to validate cost calculation in the general case.

I will try to write something for this.

laurentS:
@edgarrmondragon I've updated the code to match your suggestions, and added some testing to validate the summing code.
What do you suggest we do about the logging? If we go with something different as per the [comment](https://github.com/meltano/sdk/issues/383#issuecomment-1154701837) you linked to, should we do this as part of a separate PR?

singer_sdk/tap_base.py
```diff
@@ -379,6 +379,9 @@ def sync_all(self) -> None:
             stream.sync()
             stream.finalize_state_progress_markers()
 
+        for stream in self.streams.values():
+            stream.log_sync_costs()
+
```
edgarrmondragon, line 384: Wdyt about moving this to the first loop?

```suggestion
            stream.sync()
            stream.finalize_state_progress_markers()
            stream.log_sync_costs()
```

tests/core/test_streams.py
```diff
@@ -357,3 +357,23 @@ def test_cached_jsonpath():
 
     # cached objects should point to the same memory location
     assert recompiled is compiled
+
+
+def test_sync_costs_calculation(tap: SimpleTestTap):
```
edgarrmondragon, line 362: Thanks for adding this!

edgarrmondragon:
> @edgarrmondragon I've updated the code to match your suggestions, and added some testing to validate the summing code. What do you suggest we do about the logging? 

@laurentS thanks for the updates! I left a single suggestion. Other than that, this looks good to merge 😄.

> If we go with something different as per the [comment](https://github.com/meltano/sdk/issues/383#issuecomment-1154701837) you linked to, should we do this as part of a separate PR?

Let's leave that for another PR, another day 👍

singer_sdk/tap_base.py
laurentS: The only reason I did it this way was so all logging would be grouped at the end of the tap run, and not sprinkled around.
Since I mostly grep logs, it doesn't actually matter too much, now that I think of it.
Code would look cleaner without this. Your call :)

edgarrmondragon: @laurentS Yeah, if you don't mind too much and still works for the use case you have in mind, let's keep the single loop 🙂

tests/core/test_streams.py
```diff
@@ -357,3 +357,23 @@ def test_cached_jsonpath():
...
+        request: requests.PreparedRequest,
+        response: requests.Response,
+        context: Optional[Dict],
+    ):
+        return {"dim1": 1, "dim2": 2}
+
+    stream.calculate_sync_cost = calculate_test_cost
+    stream.update_sync_costs(fake_request, fake_response, None)
+    stream.update_sync_costs(fake_request, fake_response, None)
+    assert stream._sync_costs == {"dim1": 2, "dim2": 4}
```
edgarrmondragon, line 379: @laurentS Just to make sure we cover the `log_sync_costs` method:

```suggestion
def test_sync_costs_calculation(tap: SimpleTestTap, caplog):
    """Test sync costs are added up correctly."""
    fake_request = requests.PreparedRequest()
    fake_response = requests.Response()

    stream = RestTestStream(tap)

    def calculate_test_cost(
        request: requests.PreparedRequest,
        response: requests.Response,
        context: Optional[Dict],
    ):
        return {"dim1": 1, "dim2": 2}

    stream.calculate_sync_cost = calculate_test_cost
    stream.update_sync_costs(fake_request, fake_response, None)
    stream.update_sync_costs(fake_request, fake_response, None)
    assert stream._sync_costs == {"dim1": 2, "dim2": 4}

    with caplog.at_level(logging.INFO, logger=tap.name):
        stream.log_sync_costs()

    assert len(caplog.records) == 1

    for record in caplog.records:
        assert record.levelname == "INFO"
        assert f"Total Sync costs for stream {stream.name}" in record.message
```

edgarrmondragon: @laurentS file is missing an `import logging` line. That's the last hoop 🙏

laurentS: Fixed. Sorry about this. I need to be more careful with the github suggestions system :sweat_smile:

singer_sdk/tap_base.py
laurentS: Well, I messed up here by replying without properly looking at the code. There was another reason that I used 2 loops: the first loop skips all child streams, so the logging only happens with the top level stream (this is what I'm seeing in our logs now).
My bad. I will open a separate PR to revert this change, we do need 2 loops in the end.

laurentS: The fix is in #752. @edgarrmondragon sorry for messing this up.

PR #730, opened 2022-06-17
feat(taps): Allow sort checking to be disabled

### LINKED ISSUE #729: Support for truly sorted API's where replication_key is not an incremental one
My need comes from trying to build a tap for the UserFlow API.
Documentation: https://userflow.com/docs/api#pagination

In short, the API allows to order by a created timestamp, but does not allow filtering based on that.
The API supports pagination based on a uuid - which isn't incremental.

The current SDK behavior is that if I define the id as the replication_key, something like **ff31e109-f873-41b6-9f1c-2d88b5bf6a2a** will be tracked as the "latest record", which isn't true. It just happens to be the "highest" because of the 'ff...' in front.

So, what I would instead want to have is an API where `is_sorted = True`, but that would not check for the `replication_key` to be an increment, and instead trust that the last one seen is the most recent one.

ilkkapeltola:
This change adds a new boolean property to core: check_sorted
By default, `check_sorted = True`, which is the behavior currently when `is_sorted = True`.
However, there are API's that are sorted, but not by `replication_key`.

This change allows for an API that is truly ordered, but where the `replication_key` is not an incremental one.

Closes #729

### DIFF
--- a/singer_sdk/helpers/_state.py
+++ b/singer_sdk/helpers/_state.py
@@ -202,10 +202,11 @@ def increment_state(
     latest_record: dict,
     replication_key: str,
     is_sorted: bool,
+    check_sorted: bool,
 ) -> None:
     """Update the state using data from the latest record.
 
-    Raises InvalidStreamSortException if is_sorted=True and unsorted
+    Raises InvalidStreamSortException if is_sorted=True, check_sorted=True and unsorted
     data is detected in the stream.
     """
     progress_dict = stream_or_partition_state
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):
@ilkkapeltola This looks good to me! We'll ship this in the next release 😄

singer_sdk/helpers/_state.py
```diff
@@ -217,7 +218,7 @@ def increment_state(
         progress_dict = stream_or_partition_state[PROGRESS_MARKERS]
     old_rk_value = to_json_compatible(progress_dict.get("replication_key_value"))
     new_rk_value = to_json_compatible(latest_record[replication_key])
-    if old_rk_value is None or new_rk_value >= old_rk_value:
+    if old_rk_value is None or new_rk_value >= old_rk_value or not check_sorted:
```
aaronsteers, line 221: @edgarrmondragon - Just as a small matter, and probably a nit, what do you think of us putting the `or not check_sorted` ahead of the `>=` comparison. The difference would be small but this would prevent comparison errors in case that for some reason the values do not support comparison.

edgarrmondragon: @aaronsteers ah that makes sense, do start a PR. Or, I'll do it after lunch 🥪 .

edgarrmondragon: Tracked here: https://github.com/meltano/sdk/issues/736

aaronsteers: Thanks, @edgarrmondragon !

PR #732, opened 2022-06-17
feat(taps): Implement reference paginators

### LINKED ISSUE #316: Add reference paginator implementations
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/318

Originally created by @edgarrmondragon on 2022-01-29 01:34:13

---

## Summary

[//]: # (Concisely summarize the feature you are proposing.)

Add reference and common pagination implementations in a similar fashion to _authenticators_.

## Proposed benefits

[//]: # (Concisely summarize the benefits this feature would bring to yourself and other users.)

There is a limited number of pagination standards, which may only differ in minute details, much like authentication headers (`Authorization: Bearer <token>`, `Authorization: Token <>`), so having a sensible set of pre-built implementations may simplify things further for users by allowing them to pick one off-the-shelf.

Another benefit is that having dedicated pagination classes makes unit-testing them much easier.

Yet another benefit of moving to a dedicated class for pagination, is that the paginator state doesn't need to be limited to the previous value but can include arbitrary attributes, like the last-seen record (https://gitlab.com/meltano/sdk/-/issues/124+).

## Proposal details

[//]: # (In as much detail as you are able, describe the feature you'd like to build or would like to see built.)

I have a reference implementation [here](https://github.com/edgarrmondragon/tap-readthedocs/pull/13/files).

**TL;DR**

```python
class APIPaginator:
    """An API paginator object."""

    @property
    def current_value(self) -> TPageToken:
        """Get the current pagination value."""
        ...

    @property
    def finished(self) -> bool:
        """Get a flag that indicates if the last page of data has been reached."""
        ...

    @property
    def count(self) -> int:
        """Count the number of pages traversed so far."""
        ...

    def advance(self, response: Response) -> None:
        """Get a new page value and advance the current one."""
        ...

    def has_more(self, response: Response) -> bool:
        """Override this method to check if the endpoint has any pages left."""
        ...

    @abstractmethod
    def get_next(self, response: Response) -> Optional[TPageToken]:
        """Get the next pagination token or index from the API response."""
        ...
```

## Best reasons not to build

[//]: # (Will this negatively affect any existing functionality? Do you anticipate any breaking changes versus what may already be working today? Make the counter-argument to your proposal here.)

Can't think of any. The current `RESTStream.get_next_page_token` can be slowly deprecated with the introduction of a paginator that wraps the stream (as in [tap-readthedocs/client.py at 09dca8c653cd73e51ce265e239c94c68479481b1 · edgarrmondragon/tap-readthedocs · GitHub](https://github.com/edgarrmondragon/tap-readthedocs/blob/09dca8c653cd73e51ce265e239c94c68479481b1/tap_readthedocs/client.py#L17-L48)).

edgarrmondragon:
Docs: https://meltano-sdk--732.org.readthedocs.build/en/732

Closes #316


docs/porting.md
```diff
@@ -103,10 +103,76 @@ _Important: If you've gotten this far, this is a good time to commit your code b
...
 When you think you have it right, run `poetry run tap-mysource` again, and debug until you are confident the result is including multiple pages back from the API.
 
+You can also add unit tests for your pagination implementation for additional confidence:
+
+```python
+from singer_sdk.pagination import BaseHATEOASPaginator, first
+
+
+class CustomHATEOASPaginator(BaseHATEOASPaginator):
+   def get_next_url(self, response: Response) -> str | None:
```
aaronsteers, line 117: Since you've already created the reference docs pages, I think class docstrings that explain the pattern that each expects would be super helpful. And since many data people won't be familiar with the official pattern names, using simplified explanations and/or references to external explainers will be super helpful, I think.

```suggestion
class CustomHATEOASPaginator(BaseHATEOASPaginator):
    """Paginator for HATEOAS APIs - or "Hypermedia as the Engine of Application State".
    
    This paginator expects responses to have a key "next" with a value
    like "https://api.com/link/to/next-item".
    """"

   def get_next_url(self, response: Response) -> str | None:
```

singer_sdk/streams/rest.py
```diff
@@ -366,33 +352,24 @@ def prepare_request_payload(
...
+        if hasattr(self, "get_next_page_token"):
+            warn(
+                "`RESTStream.get_next_page_token` is deprecated and will not be used "
+                + "in a future version of the Meltano SDK. "
+                + "Override `RESTStream.get_new_paginator` instead.",
+                DeprecationWarning,
             )
-            first_match = next(iter(all_matches), None)
-            next_page_token = first_match
+            return LegacyStreamPaginator(self)  # type: ignore
```
aaronsteers, line 368: Nice balance of compatibility and forward-looking.

Should we start a 'version upgrade guide' doc that would like this as a change developers should make when updating to the newer version of the SDK?

aaronsteers (commented):
@edgarrmondragon - This is so cool!

My comments are only cosmetic, more about getting this ready for a prod release.

I think it's worth discussing if we should warn (as you have it) or just fail if they are using the legacy approach. And if we fail, would be nice to do so in a unit test (nested in get_standard_tests or similar) rather than at runtime. 😄

singer_sdk/streams/rest.py
edgarrmondragon: @aaronsteers I think we should, as we target the `1.0.0` version at some point.

singer_sdk/pagination.py
```diff
@@ -0,0 +1,417 @@
...
+        Returns:
+            The next page token or index. Return `None` from this method to indicate
+                the end of pagination.
+        """
+        return None
+
+
+class BaseHATEOASPaginator(BaseAPIPaginator[Optional[ParseResult]], metaclass=ABCMeta):
+    """Paginator class for APIs supporting HATEOAS links in their responses."""
+
```
edgarrmondragon, line 179:
```suggestion
class BaseHATEOASPaginator(BaseAPIPaginator[Optional[ParseResult]], metaclass=ABCMeta):
    """Paginator class for APIs supporting HATEOAS links in their responses.
    
    HATEOAS stands for "Hypermedia as the Engine of Application State". See https://en.wikipedia.org/wiki/HATEOAS.
    """

```

singer_sdk/streams/rest.py
```diff
@@ -24,31 +16,38 @@
...
+    SimpleHeaderPaginator,
+)
 from singer_sdk.plugin_base import PluginBase as TapBaseClass
 from singer_sdk.streams.core import Stream
 
 DEFAULT_PAGE_SIZE = 1000
 DEFAULT_REQUEST_TIMEOUT = 300  # 5 minutes
 
 _TToken = TypeVar("_TToken")
+# _TPaginator = TypeVar("_TPaginator", bound=BaseAPIPaginator)
```
aaronsteers, line 32: Should this commented code be removed?

tests/core/rest/test_pagination.py
```diff
@@ -0,0 +1,339 @@
...
+    assert not paginator.finished
+    assert paginator.current_value == "abc"
+    assert paginator.count == 1
+
+    response._content = b'{"nextPageToken": "abc"}'
+    with pytest.raises(RuntimeError, match="Loop detected in pagination"):
+        paginator.advance(response)
+
+
+def test_paginator_page_number():
```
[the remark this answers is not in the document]
aaronsteers: Fantastic! 😄

aaronsteers (approved):
Nice work, @edgarrmondragon !

Just one small question (nit).

singer_sdk/streams/rest.py
edgarrmondragon: Yup, good call :)

edgarrmondragon:
> I think it's worth discussing if we should warn (as you have it) or just fail if they are using the legacy approach. And if we fail, would be nice to do so in a unit test (nested in get_standard_tests or similar) rather than at runtime. 😄

@aaronsteers I think we could warn that it will _fail_ in the future. Wdyt?

PR #733, opened 2022-06-17
ci: Disable external checks in forks

edgarrmondragon:


edgarrmondragon:
@aaronsteers This disables the external tests on PRs coming from forks. That's how it was originally setup in GitLab and is a bit a annoying for contributors to see a failing test that they can't fix.

Here's how it looks: https://github.com/meltano/sdk/pull/734/checks

PR #737, opened 2022-06-17
fix: Check if RK is sortable before comparing

### LINKED ISSUE #736: Check if RK is sortable before comparing
> Just as a small matter, and probably a nit, what do you think of us putting the `or not check_sorted` ahead of the `>=` comparison. The difference would be small but this would prevent comparison errors in case that for some reason the values do not support comparison.
> 
> (Sorry for not seeing this earlier. I understand this is merged already and would be a small update after-the-fact.)

_Originally posted by @aaronsteers in https://github.com/meltano/sdk/pull/730#discussion_r900416920_

edgarrmondragon:
Closes #736


aaronsteers (approved):
@edgarrmondragon - Looks great. Thanks for tackling! 

Feel free to merge when ready.

PR #738, opened 2022-06-18
chore(deps): Bump pipelinewise-singer-python to 2.0.1

edgarrmondragon:
Risks:

## `orjson`

The new version of `pipelinewise-singer-python` uses [`orjson`](https://github.com/ijl/orjson) to Serialize/Deserialize Singer messages. This is meant to speed up SerDe, but comes with some caveats:

- Fails by default when fed non-string dict keys. Some taps in the wild might be relying on the JSON SerDe library not caring about the data types of dict keys, so tap maintainers would need to adapt their keys to be strings.
- Fails even when dict keys are a subclass of `str`. This is the case for SQLAlchemy results returning column names of type `sqlalchemy.sql.elements.quoted_name`, so they have to be explicitly converted to strings.
- Built on Rust. This means they publish wheels, and although the list of systems seems exhaustive, there's always the risk of developers ending up building the SDK dependencies.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -53,11 +53,11 @@ pytz = ">=2015.7"
 
 [[package]]
 name = "backoff"
-version = "1.8.0"
+version = "1.11.1"
 description = "Function decoration for backoff and retry"
 category = "main"
 optional = false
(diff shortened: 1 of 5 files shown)


edgarrmondragon:
Made irrelevant by #979

PR #740, opened 2022-06-21
Poetry-dev-deps

aaronsteers:
This adds back dev dependencies that may not be _technically_ required now that we have a remote githook. However, they are still helpful at dev-time and I think they make sense to remain as proper dev-dependencies. 

I've also alpha sorted this section of pyproject.toml so dupes as a general best practice, applied as 2 separate commits to aid in review.

The basis for keeping _all_ dev dependencies included is for IDEs that use the poetry environment for dev-time linting. The ideal scenario is that the remote linter basically has 'nothing to do' and the contributor or team member has (as closely as possible) a local dev experience that closely mirrors what will end up in the PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,29 +54,28 @@ docs = ["sphinx", "sphinx-rtd-theme", "sphinx-copybutton", "myst-parser"]
 
 [tool.poetry.dev-dependencies]
 # snowflake-connector-python = "2.0.4" # Removed: Too many version conflicts!
-pytest = "^7.1.2"
 black = "^22.3"
-mypy = "^0.961"
 cookiecutter = "^2.1.1"
+coverage = {extras = ["toml"], version = "^6.4"}
 darglint = "^1.8.0"
-PyYAML = "^6.0"
-flake8-docstrings = "^1.6.0"
 flake8 = "^3.9.0"
-pyarrow = "^8.0.0"
-pandas = ">=1.2.0"
-numpy = ">=1.20.0"
-tox = "^3.25.0"
-freezegun = "^1.2.1"
-viztracer = "^0.15.3"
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
@aaronsteers I should've documented this when I removed project dev dependencies in favor of pre-commit, but I've used [a VSCode extension](https://marketplace.visualstudio.com/items?itemName=MarkLarah.pre-commit-vscode) instead of VSCode's built-in `Format Document`, which helps linting files other than `.py` and runs checks not available to VSCode like `pyupgrade` and `isort`.

Regardless, I'm not opposed to have those dependencies added back in `pyproject.toml`.

aaronsteers:
@edgarrmondragon - do you know if https://pre-commit.ci should/could be used for `poetry lock`?

edgarrmondragon:
> @edgarrmondragon - do you know if https://pre-commit.ci should/could be used for `poetry lock`?
> 
> related:
> 
> * https://python-poetry.org/docs/master/pre-commit-hooks/#poetry-lock

Yeah, it's possible and [officially maintained by the poetry team](https://github.com/python-poetry/poetry/blob/master/.pre-commit-hooks.yaml).

Are you also suggesting that pre-commit.ci not just _checks_ PRs as it's currently doing, but also fixes them? Sorry if you're not.

aaronsteers:
@edgarrmondragon - re:

> Are you also suggesting that pre-commit.ci not just checks PRs as it's currently doing, but also fixes them? Sorry if you're not.

Yeah - I guess I incorrectly assumed we were applying changes via that tool. More discussion here:

- https://github.com/meltano/internal-general/discussions/416

aaronsteers:
Closing as stale. We can still add back any needed dev libraries in future.

Issue #742, opened 2022-06-21
Slash command (or pre-commit.ci hook) for `poetry lock` on PRs


aaronsteers:
Following from the patterns in

- https://github.com/meltano/hub/pull/613

We could add a `/poetry-lock` slash command that simply runs `poetry lock` in CI and commits the results back to any open PR.

This would be handy for version bumps and other PRs such as 

- https://github.com/meltano/sdk/pull/740

...where the code changes themselves are rather simple but can't othewise be merged back unless `poetry lock` is applied to update the lock file.

edgarrmondragon:
@aaronsteers I did a quick run to add the poetry hooks, but there seems to be some problems: https://github.com/meltano/sdk/pull/746#issuecomment-1161231250

PR #746, opened 2022-06-21
ci: Add Poetry pre-commit hooks

edgarrmondragon:
Closes https://github.com/meltano/sdk/issues/742


edgarrmondragon:
This seems to require poetry `1.2.x` (or the `master` branch), and it might produce an incompatible `poetry.lock`.

PR #750, opened 2022-06-22
[feat] Add end of pipe clean up hook to Sinks

z3z1ma:
This adds a hook to the drain all method which is flagged at endofpipe drain. I can go on to list the ways I think this is useful. From ensuring futures are completed, to tearing down resources, **clearing buckets or stages used in a load**, to cleanup in general.

### DIFF
--- a/singer_sdk/sinks/core.py
+++ b/singer_sdk/sinks/core.py
@@ -401,3 +401,7 @@ class Sink(metaclass=abc.ABCMeta):
             "ACTIVATE_VERSION message received but not implemented by this target. "
             "Ignoring."
         )
+
+    def clean_up(self) -> None:
+        """Perform any clean up actions required at end of a stream."""
+        pass
--- a/singer_sdk/target_base.py
+++ b/singer_sdk/target_base.py
@@ -273,7 +273,7 @@ class Target(PluginBase, SingerReader, metaclass=abc.ABCMeta):
 
     def _process_endofpipe(self) -> None:
         """Called after all input lines have been read."""
-        self.drain_all()
+        self.drain_all(is_endofpipe=True)
 
     def _process_record_message(self, message_dict: dict) -> None:
         """Process a RECORD message.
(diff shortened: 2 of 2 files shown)


aaronsteers:
Thanks, @z3z1ma ! 

One thing I realized afterwards is that the relationship of streams and sinks is not always obvious - for instance a new sink may be created if a schema declaration is modified after another sink for that stream is already in flight.

So, the `clean_up()` operation should just be sure to not modify or clean up the other sinks' resources, even for sinks that might have the same stream name.

Do you think that is okay in this context? I wonder if a uuid per sink, or a dedicated temp dir would help reduce chance of sinks accidentally stepping on each other during clean_up.

What do you think?

z3z1ma:
I think a sink can store arbitrary sink specific data structures and objects in a subclass of Sink since we all inherit it when using the SDK. So the user implemented clean up can inherently be sink specific. Still it's a good call out. A sink uuid prop on init might be universally ok. But not the make or break since it's easy in user land too.

edgarrmondragon:
> I think a sink can store arbitrary sink specific data structures and objects in a subclass of Sink since we all inherit it when using the SDK. So the user implemented clean up can inherently be sink specific. Still it's a good call out.

Agree 👍

singer_sdk/sinks/core.py
```diff
@@ -401,3 +401,12 @@ def activate_version(self, new_version: int) -> None:
...
         )
+
+    @abc.abstractmethod
+    def clean_up(self) -> None:
+        """Perform any clean up actions required at end of a stream.
+
+        Implementations should ensure that clean up does not affect resources
+        that may be in use from other instances of the same sink. Stream name alone
+        should not be relied on, it's recommended to use a uuid as well."""
+        pass
```
edgarrmondragon, line 412: @z3z1ma tests are complaining because existing subclasses don't implement this method:

```
TypeError: Can't instantiate abstract class SampleCSVTargetSink with abstract method clean_up
```

The base case (noop) is probably OK for most use cases, so there'd be no need to force implementations of this method. Otherwise this will introduce a breaking change for SDK-based targets in the wild. If that's what we want _eventually_, we could emit a warning for targets not implementing this method, e.g.

  ```suggestion
    def clean_up(self) -> None:
        """Perform any clean up actions required at end of a stream.

        Implementations should ensure that clean up does not affect resources
        that may be in use from other instances of the same sink. Stream name alone
        should not be relied on, it's recommended to use a uuid as well."""
        warnings.warn(
            "Future versions of the MeltanoSDK will expect custom target sinks to "
            + "implement the `clean_up` method.",
            category=FutureWarning,
        )
```

edgarrmondragon (changes requested):
@z3z1ma I left a comment re not making `clean_up` an abstract method.

PS: can you address the flake8 errors? https://results.pre-commit.ci/run/github/379008980/1656003341.A3LTOHxoRfK1mp3bSNBLSw

singer_sdk/sinks/core.py
z3z1ma: Agree the simplicity in keeping this optional is preferred. Updated.

aaronsteers:
> > I think a sink can store arbitrary sink specific data structures and objects in a subclass of Sink since we all inherit it when using the SDK. So the user implemented clean up can inherently be sink specific. Still it's a good call out.
> 
> Agree 👍

Double-agreed 👍 

Thanks.

edgarrmondragon (approved):
All checks 🟢, [docs](https://meltano-sdk--750.org.readthedocs.build/en/750/classes/singer_sdk.Sink.html#singer_sdk.Sink.clean_up) look good. Merging!

Thank you so much for your work, @z3z1ma!

PR #759, opened 2022-06-27
Use recent start_date as starting_replication_value

ericboucher:
When both a start_date and a replication_start_date are available, I believe that we should take the most recent one.

### DIFF
--- a/singer_sdk/streams/core.py
+++ b/singer_sdk/streams/core.py
@@ -320,8 +320,18 @@ class Stream(metaclass=abc.ABCMeta):
             ):
                 value = replication_key_value
 
-            elif "start_date" in self.config:
-                value = self.config["start_date"]
+            # Use start_date if it is more recent than the replication_key state
+            if "start_date" in self.config:
+                if not value:
+                    value = self.config["start_date"]
+                else:
+                    value_datetime = pendulum.parse(value)
+                    start_datetime = pendulum.parse(self.config["start_date"])
+                    value = (
+                        value
+                        if value_datetime > start_datetime
+                        else self.config["start_date"]
+                    )
 
         write_starting_replication_value(state, value)
 


singer_sdk/streams/core.py
```diff
@@ -320,8 +320,20 @@ def _write_starting_replication_value(self, context: Optional[dict]) -> None:
...
-                value = self.config["start_date"]
+            # Use start_date if it is more recent than the replication_key state
+            if "start_date" in self.config:
+                if not value:
+                    value = self.config["start_date"]
+                else:
+                    value_datetime = cast(datetime.datetime, pendulum.parse(value))
+                    start_datetime = cast(
+                        datetime.datetime, pendulum.parse(self.config["start_date"])
+                    )
```
edgarrmondragon, line 331: @ericboucher what would happen here if either of these values isn't date-parseable? e.g.  if they are UNIX timestamps.

ericboucher: I re-used what was in `get_starting_timestamp` but I can raise a more explicit error if that's useful. Do you prefer that I:
1. raise a more explicit error
2. handle unix timestamps explicitely

edgarrmondragon (changes requested):

ericboucher:
@edgarrmondragon I added a check for `is_timestamp_replication_key` to make it a bit safer and explcit. wdyt?

Since we are in the subcase where a start_date is given as an input, I think this should be enough, but we could also add a try/except around the parsing.

As an alternative, we can indeed make a simple string comparison, although we are not sure what would happen when comparing a uuid and a date from "start_date".

I couldn't find if the format of `start_date` is actually enforced anywhere.

edgarrmondragon:
@ericboucher ok, that makes sense and might be good for most use cases. Would you mind adding a test case, perhaps somewhere in https://github.com/meltano/sdk/blob/b0a3c497cae2d4e33d18fc1d5415a37d2a7defb5/tests/core/test_streams.py#L132?

ericboucher:
@edgarrmondragon done! Not sure why codecov is not passing though... Any clue? The UI is not very readable... 🤷

edgarrmondragon:
@ericboucher it seems that codecov doesn't like that there isn't an `else` branch

Screen Shot 2022-07-15 at 18 23 22

I think I have a quick fix for that, and if that works I might push it to this branch 🙂

edgarrmondragon (approved):

singer_sdk/streams/core.py
```diff
@@ -320,8 +320,15 @@ def _write_starting_replication_value(self, context: Optional[dict]) -> None:
...
-                value = self.config["start_date"]
+            # Use start_date if it is more recent than the replication_key state
+            if "start_date" in self.config:
+                start_date_value = self.config["start_date"]
+                if not value:
+                    value = start_date_value
+                elif self.is_timestamp_replication_key:
+                    value = max(value, start_date_value, key=pendulum.parse)
+                else:
+                    value = max(value, start_date_value)
```
aaronsteers, line 331: Hi, @ericboucher.

If the replication key is not a timestamp, I think we should probably skip this entire block, since in that case `replication_key_value` is not expected to be comparable with `start_date_value`.

```
            if "start_date" in self.config and self.is_timestamp_replication_key:
               ...
```

I think this also simplifies the logic in the `if` block and makes the last `else` case unnecessary.
 
What do you think?

tests/core/test_streams.py
```diff
@@ -48,6 +48,16 @@ def get_records(self, context: Optional[dict]) -> Iterable[Dict[str, Any]]:
...
 
 
+class UnixTimestampIncrementalStream(SimpleTestStream):
+    name = "unix_ts"
+    schema = PropertiesList(
+        Property("id", IntegerType, required=True),
+        Property("value", StringType, required=True),
+        Property("updatedAt", IntegerType, required=True),
+    ).to_dict()
+    replication_key = "updatedAt"
```
aaronsteers, line 58: As of now, I believe we only accept ISO timestamp strings and not Unix timestamps (integers). (Someone please correct me if that's not correct.)

Would it be okay to keep Unix timestamps out of scope for this and to discuss those in another issue/PR?

aaronsteers (dismissed):

singer_sdk/streams/core.py
edgarrmondragon: @aaronsteers we don't impose any data type on `start_date` and it's up to the tap developer. See for example https://github.com/MeltanoLabs/tap-stackexchange/blob/4661caa7f35c310f45beb8222e1320d573624555/tap_stackexchange/tap.py#L45-L49

That means `value` and `start_date_value` would be comparable if a UNIX timestamp is used for both. I don't think we want to restrict `start_date` to date strings alone. Wdyt?

tests/core/test_streams.py
edgarrmondragon: @aaronsteers ditto my other comment

singer_sdk/streams/core.py
aaronsteers: In our [Singer Spec docs](https://hub.meltano.com/singer/spec#config-files), we note that the field should be in [RFC3339 format](https://medium.com/easyread/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a#:~:text=There%20are%20already%20so%20many,org%2Frfc%2Frfc3339.txt).

Here's the corresponding section in the Singer docs:

https://github.com/singer-io/getting-started/blob/ca5c56f16e67e0c2da49e9bb94d6f52fb19c9de6/docs/CONFIG_AND_STATE.md#special-fields

Both `start_date` and `user_agent` are special fields according to the singer spec, and the Singer docs confirm that the date should be in RFC3339.

edgarrmondragon: @aaronsteers IMO _should_ is the key word here, so I'd say it's only the _recommended_ format if we look at [RFC best practices][1] on requirement levels.

Also, it'd be weird to scratch non-datetime `start_date` values from this fix. It doesn't make sense to me to ship a partial fix when we can be more thorough, and already have a test case and implementation in this PR.

[1]: https://datatracker.ietf.org/doc/html/rfc2119

aaronsteers: I read "should" as a hard requirement and I have serious concerns about allowing integer datetime values. I did not know we had any cases in the wild already where start_date was epoch/unix timestamp based and my preference is to unwind these or implement the custom fixes/overrides there in those taps.

Again, because `start_date` format is actually dictated by the spec (according to my reading, at least), orchestrators and users would reasonably expect to be able to send a consistent stringified date value to all taps in a project. If we diverge on this point and encourage tap developers to diverge, then we break assumptions in a way that will create silent failures and buggy behavior for our end users. For instance, trying to do an alpha greater-than-or-less-than comparison between a stringified integer and a stringified datetime will not fail but it will more likely just produce a wrong result. Many non-SDK taps which we cannot control are simply running a naive string comparison, which would work for the suggested standard and fail silently when mixed and matched with an epoch/integer format.

I'm not going to say tap developers can't go against the standard, but we should keep with the safe prescriptive guidance here, along a path that provides the most consistent experience for users.


Lastly, there may be intuitions to tightly couple the tap config interface to the upstream API's behavior and I would strongly discourage us from that approach. For one, it implies that the API behaviors are monolithic, where in fact each stream in a tap (or even future-added streams to the tray) may each have a completely different internal data type expectation. By keeping with a standard recommendation, we keep the external user interface the same regardless of internals of the upstream APIs.

Secondly, the tap users (and downstream targets!) should be reasonably isolated from internal storage preferences of the distinct quirks and idioms of every upstream API. If something should arrive and be stored as a datetime, we should send it to the target as such. 

Lastly, to be clear, I'm not saying the standard stringified date format is better or more preferable to epoch or Unix time - but given that the choice is already made, and given there's a high cost of supporting different taps on different standards, I do not think we should accommodate or support a divergence in this area.

(In the morning, I'll move this to a separate discussion since it's becoming a larger and more in-depth discussion that I think is warranted for this sort order fix PR.)

edgarrmondragon: @aaronsteers I pushed the UNIX timestamp code branch, I can revert/undo the changes and apply the conditional you suggest.

ericboucher:
@edgarrmondragon your changes make sense to me! Let me know if you any other changes from me :)

edgarrmondragon (approved):

edgarrmondragon:
@ericboucher thanks for getting this started! It prompted a useful discussion on our side 😄

PR #768, opened 2022-06-30
feat: Better error messages when config validation fails

### LINKED ISSUE #753: feature: Validate tap/target config options
I just found that I've been running a tap with a misnamed option in the config file for almost 3 months. Despite this fact, `name=tap-github level=INFO message=Config validation passed with 0 warnings.`
It would be helpful if the sdk, as part of the config validation, actually raised an error when encountering unexpected options. For instance, `mypy` has an [option](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-configs) for this, which is very helpful to catch typos in config and such.

The following test clarifies the bug (inserted in `tests/core/test_plugin_config.py`):

```python
SAMPLE_CONFIG_WITH_ADDITIONAL_OPTION: Dict[str, Any] = {
    "default_true": False,
    "option_not_in_schema": 1,
}
def test_tap_config_validation_fails_on_invalid_option():
    """Verify that config validation fails on invalid option."""
    with pytest.raises(ValidationError):
        TapConfigTest(
            config=SAMPLE_CONFIG_WITH_ADDITIONAL_OPTION, parse_env_config=True
        )
    # this test fails because ValidationError is not raised
```

edgarrmondragon:
## Summary of changes

1. Writes each validation error message to stderr and aborts the app execution:

   ```console
   $ poetry run python samples/sample_tap_sqlite
   2023-11-17 16:09:23,020 | ERROR    | tap-sqlite-sample    | Config validation error: 'path_to_db' is a required property
   ```

2. Adds tests that validate the tap and target constructors and CLI output.


tests/core/test_streams.py
```diff
@@ -78,9 +81,14 @@ class SimpleTestTap(Tap):
     """Test tap class."""
 
     name = "test-tap"
-    settings_jsonschema = PropertiesList(Property("start_date", DateTimeType)).to_dict()
+    config_jsonschema = PropertiesList(
+        Property("username", StringType, required=True),
+        Property("password", StringType, required=True),
+        Property("start_date", DateTimeType),
+        additional_properties=False,
```
laurentS, line 88: I think this fixes my use case. I'd almost be in favour of making the stricter behaviour the default though, as I find the current config validation a bit misleading: the tap says everything is fine, even though it isn't. Or should the `Config validation passed with 0 warnings.` message include a warning that it did not check for some types of errors, such as additional properties, unless this is set?
I'm conscious that this would be a "breaking change" in the sense that it would break apparently valid configs, but it would probably surface problems to tap users that they might not be aware of, which I think is a win. In my case, I was using `tap-github` with `skip_parent_stream` (with the final `s` missing), believing that I was skipping a whole bunch of API calls. This change would have made the problem more obvious, and easily fixed. Are there cases where the user would want to run with a "broken" config?

edgarrmondragon: @laurentS I'm also inclined to make this strict validation the default. A slightly more complicated approach (for us), that is backwards compatible, is to iterate over the individual validators and catch the one for `additionalProperties`. Wdyt?

laurentS: I don't really understand the difference between the 2 behaviours. But if there are more validators (there seems to be a long list in the code), is your second approach more flexible, possibly making it possible to perform more sorts of validations that would help prevent errors? If so, I'd lean towards that approach as a user of the sdk, but I also don't know how much extra work it would be to switch to that...

singer_sdk/exceptions.py
```diff
@@ -11,6 +11,21 @@
 class ConfigValidationError(Exception):
     """Raised when a user's config settings fail validation."""
 
+    def __init__(
+        self,
+        message: str,
+        *,
+        errors: list[str] | None = None,
```
laurentS, line 18: splitting hairs here, but can't this be
```suggestion
        errors: list[str] = [],
```
so that types are a tiny bit more constrained?

laurentS (approved):
This seems very helpful!

singer_sdk/exceptions.py
WillDaSilva: Though that would be nice, assigning a mutable object like a list as a default value for a function parameter means that every call shares that default object. If the function every modifies the list, then those modifications will apply to every future call.

In a future version of Python this will hopefully be fixed by being able to write `errors: list[str] => []`, which will result in the default value being recomputed on every call, instead of when the function is defined. See [PEP 671](https://peps.python.org/pep-0671/).

laurentS: This is a very good point! If forget this point way too often :( Thanks for pointing it out!

PR #781, opened 2022-07-01
fix: Allow lists and dictionaries as types for default JSON values

### LINKED ISSUE #782: Typing error when using a list as a default value in `singer_sdk.typing.Property`
mypy is complaining on taps that use a list as the default value:

```
error: Argument "default" to "Property" has incompatible type "List[<nothing>]"; expected "Union[str, float, None]"
```

edgarrmondragon:
Closes #782

### DIFF
--- a/singer_sdk/typing.py
+++ b/singer_sdk/typing.py
@@ -86,7 +86,15 @@ __all__ = [
     "PropertiesList",
 ]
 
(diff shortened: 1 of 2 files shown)


edgarrmondragon:
Pretty safe change here, so I'll go ahead merge.

PR #785, opened 2022-07-02
Output full path in error messages

ericboucher:
Output actual full path in `response_error_message` instead of `"{self.path}"`

### DIFF
--- a/singer_sdk/streams/rest.py
+++ b/singer_sdk/streams/rest.py
@@ -16,6 +16,7 @@ from typing import (
     TypeVar,
     Union,
 )
+from urllib.parse import urlparse
 
 import backoff
 import requests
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
Hi @ericboucher, I'm generally in favor of this change, but do you know if there are any security concerns with logging the formatted path, which may contain contextual information (entity IDs, etc.)?

ericboucher:
I was thinking about this. I think it's generally fine since it is not logging URL parameters where you could have a token or sensitive data. An option could be to add a line to obfuscate secrets but I am not sure it's necessary.

edgarrmondragon:
> I was thinking about this. I think it's generally fine since it is not logging URL parameters where you could have a token or sensitive data. An option could be to add a line to obfuscate secrets but I am not sure it's necessary.

@ericboucher alright 👍. Can you then just add a note to the method's docstring indicating the cases in which it may be desirable to override the method (e.g. when the URL path contains secrets or PII)? That'll render in the docs page.

edgarrmondragon (approved):
@ericboucher lgtm, thanks!

PR #802, opened 2022-07-10
ci: Test on Windows and macOS

edgarrmondragon:


pyproject.toml
```diff
@@ -104,8 +104,19 @@ version_files = [
...
+    "*/singer_sdk",
+]
+tests = [
+    "tests/",
+    "*/tests",
+]
+
 [tool.coverage.run]
-source = ["singer_sdk"]
+branch = true
```
edgarrmondragon, line 118: This makes coverage a bit more strict by checking branches in `if` statements. This means our overall coverage will seem to have reduced, but it's only because we're now moving the bar slightly up.

See https://coverage.readthedocs.io/en/6.4.1/branch.html#branch

Issue #744, opened 2022-06-21
License scanning for SDK libraries


aaronsteers:
This is not-yet-done and was flagged as a follow-up action from 

- https://github.com/meltano/sdk/issues/743

cc @edgarrmondragon, @kgpayne

edgarrmondragon:
Some third-party options:

- [**snik.io**](https://snyk.io/). I've never used it but dbt-labs folks do, and it can automatically create PRs for vulnerability fixes.
- [**fossa.com**](https://fossa.com/). I've used it before and it runs in PRs that touch `pyproject.toml`/`poetry.lock`.

  
  <img width="832" alt="Screen Shot 2022-06-22 at 12 07 00" src="https://user-images.githubusercontent.com/16805946/175096768-a2bb91c4-2023-43fc-ab8f-5bb54c273c47.png">

WillDaSilva:
I had previously considered Snyk (Snik?), but I'm hesitant to use them because we would be only be able to run a limited number of tests per month unless we [purchase one of their expensive plans](https://snyk.io/plans/). It would be unfortunate for our pipelines to begin failing just because we had run too many that month.

Fossa seems great. Thanks for the recommendation @edgarrmondragon.

I don't believe licensebat gives us anything that it's provided by `actions/dependency-review-action` (which is being added to the Meltano repository by https://github.com/meltano/meltano/pull/6403). It's probably worth using the same approach for both repositories for their Python dependencies and security analysis. They'll only differ since Meltano also needs container scanning, but the SDK does not.

edgarrmondragon:
thanks @WillDaSilva, I agree in using the same approach for both repos so I started https://github.com/meltano/sdk/pull/808.

PR #808, opened 2022-07-13
ci: Add dependency security review GitHub workflow

edgarrmondragon:


WillDaSilva (approved):
Thanks @edgarrmondragon! This PR implements vulnerability scanning for dependencies, but until we've decided on a list of permitted licenses it does not close #744. I've updated the description to prevent #744 from being closed by this PR.

I will update both Meltano and the SDK with the list of permitted licenses later today.

edgarrmondragon:
@WillDaSilva awesome, thanks!

PR #816, opened 2022-07-14
fix(templates): Generate correct SQL target project from cookiecutter

### LINKED ISSUE #815: target cookie cutter SQL option doesn't use SQL
1. `cookiecutter https://github.com/meltano/sdk --directory="cookiecutter/target-template"`
2. cat sinks.py

```python
# sinks.py
"""testermctestface target sink class, which handles writing streams."""


from singer_sdk.sinks import RecordSink


class testermctestfaceSink(RecordSink):
    """testermctestface target sink class."""

    def process_record(self, record: dict, context: dict) -> None:
        """Process the record."""
        # Sample:
        # ------
        # client.write(record)
```

edgarrmondragon:
- Create connector class
- Sink inherits from `SQLSink`
- Sink declares `connector_class` attribute
- Target class inherits from `SQLTarget`
- Target declares `sqlalchemy_url` setting.

Closes #815

### DIFF
--- a/cookiecutter/target-template/{{cookiecutter.target_id}}/{{cookiecutter.library_name}}/sinks.py
+++ b/cookiecutter/target-template/{{cookiecutter.target_id}}/{{cookiecutter.library_name}}/sinks.py
@@ -1,7 +1,42 @@
 """{{ cookiecutter.destination_name }} target sink class, which handles writing streams."""
 
-{% set sinkclass = "BatchSink" if cookiecutter.serialization_method == "Per batch" else "RecordSink" %}
-from singer_sdk.sinks import {{ sinkclass }}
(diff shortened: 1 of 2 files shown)


visch:
@edgarrmondragon  I'm not sure if we want a  `connector.py`  in the template or not, this seems much better than what's currently there so 👍 from me

visch:
> @edgarrmondragon I'm not sure if we want a `connector.py` in the template or not, this seems much better than what's currently there so 👍 from me

https://github.com/meltano/sdk/pull/816/files#diff-f17282cce4b3d36d42160b741e62750beb82699aa577e2b8c5af99175532988cR20 you already handled it 👏

PR #826, opened 2022-07-16
chore(templates): Remove Poetry version pin for GitHub Actions

visch:
Hit an issue in CI with one of these pinned versions. Probably best to grab the latest version with the template.

Pinning is ok, but we should probably add a #TODO to have the developer pull the latest version of Poetry (Or we could do it programmatically with cookiecutter when a new template is made?)

Probably easiest not to pin!


cookiecutter/tap-template/{{cookiecutter.tap_id}}/.github/workflows/{% if cookiecutter.include_cicd_sample_template == 'GitHub' %}ci_workflow.yml{%endif%}
```diff
@@ -22,8 +22,6 @@ jobs:
         python-version: {{ '${{ matrix.python-version }}' }}
     - name: Install Poetry
       uses: snok/install-poetry@v1
-      with:
-        version: 1.1.8
```
edgarrmondragon, line 26: @visch Can a compatible version be used instead, i.e. something like `1.1`?

Future versions of Poetry could have breaking changes.

cookiecutter/tap-template/{{cookiecutter.tap_id}}/.github/workflows/{% if cookiecutter.include_cicd_sample_template == 'GitHub' %}ci_workflow.yml{%endif%}
```diff
@@ -48,8 +46,6 @@ jobs:
         python-version: {{ '${{ matrix.python-version }}' }}
     - name: Install Poetry
       uses: snok/install-poetry@v1
-      with:
-        version: 1.1.11
```
edgarrmondragon, line 52: Ditto compatible version

edgarrmondragon (changes requested):

cookiecutter/tap-template/{{cookiecutter.tap_id}}/.github/workflows/{% if cookiecutter.include_cicd_sample_template == 'GitHub' %}ci_workflow.yml{%endif%}
visch: Tried that here https://github.com/MeltanoLabs/target-postgres/pull/15/files , no good https://github.com/MeltanoLabs/target-postgres/actions/runs/2685763367 

Going to implement this slightly different, but same result as sem ver to 1.1 and simpler imo

visch: Done, what do you think about not using the action at all? Same number of lines, but we mess with the system site-packages

edgarrmondragon (approved):

visch:
Thanks for doing the work here @edgarrmondragon  :)

cookiecutter/tap-template/{{cookiecutter.tap_id}}/.github/workflows/{% if cookiecutter.include_cicd_sample_template == 'GitHub' %}ci_workflow.yml{%endif%}
```diff
@@ -21,12 +21,9 @@ jobs:
...
         python-version: {{ '${{ matrix.python-version }}' }}
     - name: Install Poetry
-      uses: snok/install-poetry@v1
-      with:
-        version: 1.1.8
-    - name: Install dependencies
       run: |
-        poetry install
+        python -m pip install --upgrade pip
+        pip install poetry==1.1.*
```
edgarrmondragon, line 26:
```suggestion
        pip install poetry==1.2.*
```

cookiecutter/tap-template/{{cookiecutter.tap_id}}/.github/workflows/{% if cookiecutter.include_cicd_sample_template == 'GitHub' %}ci_workflow.yml{%endif%}
```diff
@@ -47,9 +44,9 @@ jobs:
       with:
         python-version: {{ '${{ matrix.python-version }}' }}
     - name: Install Poetry
-      uses: snok/install-poetry@v1
-      with:
-        version: 1.1.11
+      run: |
+        python -m pip install --upgrade pip
+        pip install poetry==1.1.*
```
edgarrmondragon, line 49:
```suggestion
        pip install poetry==1.2.*
```

Issue #825, opened 2022-07-16
Cookie Cutter flake8-docstrings instead of pydocstyle


visch:
`tox.ini` has a check for `poetry run pydocstyle target_postgres` in both the tap and target cookie cutters.

We could replace that with `flake8-docstrings` as that's how the sdk and meltano handle it. 🤷  took me away from coding a bit today as I"m trying to understand `pydocstyle` and saw this

edgarrmondragon:
Closing in favor of https://github.com/meltano/sdk/issues/1596

PR #828, opened 2022-07-17
chore(templates): Add dependabot updates for Python dependencies on cookiecutters

visch:
While looking at https://github.com/meltano/sdk/issues/825 I noticed the `pyproject.toml` files were out of sync between the two cookiecutter projects (tap and target). Thought about a way to fix this without copy pasting over and over again

Doesn't work yet, need to run a cookiecutter for a test project right before checking with Dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,3 +22,15 @@ updates:
       interval: weekly
     reviewers: [meltano/engineering]
     labels: [deps]
+  - package-ecosystem: pip
+    directory: "./cookiecutter/tap-template/{{cookiecutter.tap_id}}"
+    schedule:
+      interval: weekly
+    reviewers: [meltano/engineering]
+    labels: [deps]
+  - package-ecosystem: pip
+    directory: "./cookiecutter/target-template/{{cookiecutter.tap_id}}"
+    schedule:
+      interval: weekly
+    reviewers: [meltano/engineering]
+    labels: [deps]


edgarrmondragon:
@visch I've thought about this but didn't start any work because dependabot will not work with the existing `pyproject.toml` files, since they aren't really valid TOML. See https://github.com/meltano/sdk/blob/c82c71c96ad9060617ae59c4815ccb7990a203dd/cookiecutter/tap-template/%7B%7Bcookiecutter.tap_id%7D%7D/pyproject.toml#L38

We could put that key in quotes and the file may pass as valid TOML, but I worry that in the future we may want to add more templating stuff to the files (think conditionals for extras, etc.) and then break dependabot updates.

Wdyt?

visch:
> @visch I've thought about this but didn't start any PRs because dependabot will not work with the existing `pyproject.toml` files, since they aren't really valid TOML. See
> 
> https://github.com/meltano/sdk/blob/c82c71c96ad9060617ae59c4815ccb7990a203dd/cookiecutter/tap-template/%7B%7Bcookiecutter.tap_id%7D%7D/pyproject.toml#L38
> 
> We could put that key in quotes and the file may pass as valid TOML, but I worry that in the future we may want to add more templating stuff to the files (think conditionals for extras, etc.) and then break dependabot updates.
> 
> Wdyt?

I was hopeful there'd be someway to run some kind of script on your repo before Dependabot scanned them but it looks like that isn't an option. 

1. We could do something like a github action that's triggered based on new changes in `main`. When detected we go generate a proper `pyproject.toml` and place it into a folder, say 

2. Another less "risky" option but would take more manual work is to create a repo that generates tap/target projects on changes to `main` from the sdk. This would automatically merge the new `pyproject.toml` files into something like `meltano/sdk-cookiecutter-tests/tap-dependabot`  and ``meltano/sdk-cookiecutter-tests/target-dependabot`  Then we'd just scan those, when a PR comes in we'd have to manually go get a PR to the `sdk` (could probably automate this too via some convoluted scripts!). 

2 is probably better as it'd also offer a good place to test that each combination of cookie cutters is generated properly with some kind of tests to automatically build a sample tap via the cookie cutters that exist

edgarrmondragon:
Our `pyproject.toml` templates are more complex now and will definitely not work with dependabot. Still open to having this, but we need a different approach.

PR #829, opened 2022-07-18
feat(taps): Add checks for primary keys, replication keys and state partitioning keys to standard tap tests

laurentS:
This PR adds 2 tests to the standard ones that all sdk-based taps run:
- 1 checks that PKeys are actually in the stream schema
- 1 does the same for replication keys

Without these tests, it is easy to write invalid streams that will only fail after a certain amount of processing (usually because the target will complain). See [this thread](https://github.com/MeltanoLabs/tap-gitlab/pull/71#discussion_r884035701) in a separate PR for context.
These tests help catch coding errors much sooner, as some errors might live in rarely used streams (or untested ones).

Ideally we could add a third test that validates that all path variables are available in the stream context. This might be best addressed in a separate PR, as this is a bit harder to test, given the dynamic nature of the stream context.


laurentS:
I've tested this with the code in https://github.com/MeltanoLabs/tap-gitlab/pull/71 and it does correctly flag problems.

visch:
@laurentS  another trick I've been doing is baking a jsonl test into the github actions test. Requires a full data pull but 🤷  just an idea not really for this PR but I happened to see this while cruising through things today!

```yaml
  jsonl_test:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: Set up Python
      uses: actions/setup-python@v2
      with:
        python-version: "3.8"
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install meltano
        meltano install
    - name: Schema test
      run: |
        meltano run tap-docusign target-jsonl
```

visch:
@laurentS  another trick I've been doing is baking a jsonl test into the github actions test. Requires a full data pull but 🤷  just an idea not really for this PR but I happened to see this while cruising through things today!

```yaml
  jsonl_test:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: Set up Python
      uses: actions/setup-python@v2
      with:
        python-version: "3.8"
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install meltano
        meltano install
    - name: Schema test
      run: |
        meltano run tap-name target-jsonl
```

laurentS:
> Hi and thank you @laurentS! I see in the PR description that is meant to add tests for replication keys, but I only see them for primary keys and state partitioning keys. Let me know if you're planning to add them. Otherwise I can review (and merge) it as is :smile:.

Sorry, this was my mistake, but I added an extra test, since it seems like a pretty useful one for correctness. I think this is ready for review now. There are some failures in CI that I'm not sure how to fix.

singer_sdk/testing.py
```diff
@@ -40,7 +40,50 @@ def _test_stream_connections() -> None:
         tap1: Tap = tap_class(config=config, parse_env_config=True)
         tap1.run_connection_test()
 
-    return [_test_cli_prints, _test_discovery, _test_stream_connections]
+    def _test_pkeys_in_schema():
```
edgarrmondragon, line 43:
```suggestion
    def _test_pkeys_in_schema() -> None:
```

singer_sdk/testing.py
```diff
@@ -40,7 +40,50 @@ def _test_stream_connections() -> None:
...
+        tap = tap_class(config=config, parse_env_config=True)
+        for name, stream in tap.streams.items():
+            pkeys = stream.primary_keys or []
+            schema_props = set(stream.schema["properties"].keys())
+            for pkey in pkeys:
+                assert (
+                    pkey in schema_props
+                ), f"Coding error in stream '{name}': primary_key '{pkey}' is missing in schema"
+
+    def _test_state_partitioning_keys_in_schema():
```
edgarrmondragon, line 54:
```suggestion
    def _test_state_partitioning_keys_in_schema() -> None:
```

edgarrmondragon (changes requested):

edgarrmondragon:
> There are some failures in CI that I'm not sure how to fix.

@laurentS Some of those are linter errors that I've suggested fixes for. The coverage failures I think we can ignore.

edgarrmondragon:
pre-commit.ci autofix

singer_sdk/testing.py
```diff
@@ -40,7 +40,52 @@ def _test_stream_connections() -> None:
...
+    def _test_pkeys_in_schema() -> None:
+        """Verify that primary keys are actually in the stream's schema."""
+        tap = tap_class(config=config, parse_env_config=True)
+        for name, stream in tap.streams.items():
+            pkeys = stream.primary_keys or []
+            schema_props = set(stream.schema["properties"].keys())
+            for pkey in pkeys:
+                error_message = (
+                    f"Coding error in stream '{name}': " +
+                    f"primary_key '{pkey}' is missing in schema"
```
edgarrmondragon, line 52:
```suggestion
                    f"Coding error in stream '{name}': "
                    + f"primary_key '{pkey}' is missing in schema"
```

edgarrmondragon: @laurentS I think this is the only thing missing for `pre-commit` to pass 🙂

edgarrmondragon (changes requested):

singer_sdk/testing.py
laurentS: @edgarrmondragon sorry I've been incredibly slow with this. I've fixed the formatting issue above, and rebased on latest `main`. I think it's good to go, except for the codecov checks

edgarrmondragon (approved):
@laurentS codecov is not deal breaker since at least replication key checks should be covered by external GitLab tests when this lands on `main`. The state partitioning keys test requires a sample that uses those, so that's not on you really.

cc @aaronsteers in case you want to groom the title

PR #830, opened 2022-07-18
docs: Easy to use samples in `pyproject.toml` with steps

visch:


edgarrmondragon:
@visch thanks! This requires users to edit a file that they'd need to be careful not to commit if they happen to also be contributing to the sdk repo . Wdyt of adding a __main__.py to each of the sample sub-packages to make them executable?

visch:
> @visch thanks! This requires users to edit a file that they'd need to be careful not to commit if they happen to also be contributing to the sdk repo . Wdyt of adding a **main**.py to each of the sample sub-packages to make them executable?
> 
> For example, as
> 
> ```shell
> poetry run samples/sample_tap_sqlite
> ```
> 
> Also, can you add a brief description or link to an issue?

I was hopeful there'd be someway to run some kind of script on your repo before Dependabot scanned them but it looks like that isn't an option. 

1. We could do something like a github action that's triggered based on new changes in `main`. When detected we go generate a proper `pyproject.toml` and place it into a folder, say 

2. Another less "risky" option but would take more manual work is to create a repo that generates tap/target projects on changes to `main` from the sdk. This would automatically merge the new `pyproject.toml` files into something like `meltano/sdk-cookiecutter-tests/tap-dependabot`  and ``meltano/sdk-cookiecutter-tests/target-dependabot`  Then we'd just scan those, when a PR comes in we'd have to manually go get a PR to the `sdk` (could probably automate this too via some convoluted scripts!). 

2 is probably better as it'd also offer a good place to test that each combination of cookie cutters is generated properly with some kind of tests to automatically build a sample tap via the cookie cutters that exist

edgarrmondragon:
Better addressed by #915

PR #842, opened 2022-07-20
feat: Allow authenticating more generic requests

### LINKED ISSUE #170: Allow Authenticator to mutate outgoing request
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/171

Originally created by @edgarrmondragon on 2021-07-19 20:05:11

---

## Summary
[//]: # (Concisely summarize the feature you are proposing.)
Some APIs use a form of [digest authentication](https://en.wikipedia.org/wiki/Digest_access_authentication) where the URL, query parameters, body, API secrets and a [_nonce_](https://en.wikipedia.org/wiki/Cryptographic_nonce) are required to sign a request. An example of this sort of authentication is used by the [Bitso API](https://bitso.com/api_info?python#generating-api-keys).

At the moment, adding this auth method to a tap implies overriding `RESTStream.prepare_request` with an almost identical implementation, only adding the lines that pertain to the calculation of the right auth headers using the other request arguments.

I propose instead, allowing authenticator class to mutate the outgoing request, giving it access to the raw, unprepared, request object so it can be "signed", e.g. headers updated appropriately.

## Proposed benefits
[//]: # (Concisely summarize the benefits this feature would bring to yourself and other users.)

1. The base `RESTStream.prepare_request` would then rarely need to be overridden even for idiosyncratic authentication mechanisms.
2. Authentication is further decoupled from the client/stream implementation.
3. Support other "dynamic" authentication methods that need access to the request.

## Proposal details
[//]: # (In as much detail as you are able, describe the feature you'd like to build or would like to see built.)

Make the authenticator interface ` APIAuthenticatorBase` have an abstract method `authenticate_request(self, request: requests.Request)` and use this inside of `RESTStream.prepare_request`:

```python
class RESTStream(Stream, metaclass=abc.ABCMeta):
    def prepare_request(
        self, context: Optional[dict], next_page_token: Optional[Any]
    ) -> requests.PreparedRequest:
        """Prepare a request object.

        If partitioning is supported, the `context` object will contain the partition
        definitions. Pagination information can be parsed from `next_page_token` if
        `next_page_token` is not None.
        """
        http_method = self.rest_method
        url: str = self.get_url(context)
        params: dict = self.get_url_params(context, next_page_token)
        request_data = self.prepare_request_payload(context, next_page_token)
        headers = self.http_headers

        request = requests.Request(
            method=http_method,
            url=url,
            headers=headers,
            params=params,
            data=request_data,
        )
        # Mutate the request to include any auth headers, parameters, etc.
        self.authenticator.authenticate_request(request)

        prepared_request: requests.PreparedRequest = (
            self.requests_session.prepare_request(request)
        )
        return prepared_request
```

For the API that motivated this issue, the authenticator class would look like this:

```python
"""Bitso Authentication."""

import hashlib
import hmac
import json
import time
from urllib.parse import urlencode, urlparse

from requests import Request
from singer_sdk.authenticators import APIAuthenticatorBase


class BitsoAuthenticator(APIAuthenticatorBase):
    """Authenticator class for Bitso."""

    @classmethod
    def create_for_stream(cls, stream) -> "BitsoAuthenticator":
        """Create the authenticator for the stream."""
        return cls(stream=stream)

    def authenticate_request(self, request: Request) -> None:
        """Mutate outgoing request with authentication data.

        See: https://bitso.com/api_info?python#creating-and-signing-requests
        """
        bitso_key: str = self.config["key"]
        bitso_secret: str = self.config["secret"]
        nonce = str(int(round(time.time() * 1000)))

        _, path = request.url.split(self.config["base_url"])
        message = nonce + request.method + path

        if request.method.lower() == "post":
            message += json.dumps(request.data)

        if request.params:
            message += "?" + urlencode(request.params)

        signature = hmac.new(
            bitso_secret.encode("utf-8"),
            message.encode("utf-8"),
            hashlib.sha256,
        ).hexdigest()
        auth_header = "Bitso %s:%s:%s" % (bitso_key, nonce, signature)

        # Update request with Bitso auth
        request.headers.update({"Authorization": auth_header})
```

Other authentication methods would require simpler `authenticate_request` implementations that perhaps only add a header, given a token.

## Best reasons not to build
[//]: # (Will this negatively affect any existing functionality? Do you anticipate any breaking changes versus what may already be working today? Make the counter-argument to your proposal here.)

1. Arguably, overriding `prepare_request` is not too much of a burden.

edgarrmondragon:
Closes #170


singer_sdk/authenticators.py
```diff
@@ -95,6 +95,18 @@ def auth_params(self) -> dict:
...
+        """Authenticate a request.
+
+        Args:
+            request: A `request object`_.
+
+        .. _request object:
+            https://requests.readthedocs.io/en/latest/api/#requests.Request
+        """
+        request.headers.update(self.auth_headers or {})
+        request.params.update(self.auth_params or {})
```
WillDaSilva, line 108: The properties already default to `{}`, so there's no need to do it here too. The properties themselves also seem redundant - why not just use the attributes `auth_headers` and `auth_params` directly, and change them into properties in the future if they need special handling?

```suggestion
        request.headers.update(self.auth_headers)
        request.params.update(self.auth_params)
```

edgarrmondragon: > why not just use the attributes `auth_headers` and `auth_params` directly, and change them into properties in the future if they need special handling?

@WillDaSilva I agree. `_auth_headers` should just be `auth_headers` in some of the constructors. Do you think it's worth trying that in this PR?

WillDaSilva (approved):

singer_sdk/authenticators.py
WillDaSilva: @edgarrmondragon Let's focus on getting this PR merged in today in time for the release. That change is just cleanup that can wait and/or be done in a separate PR.

PR #868, opened 2022-07-28
fix(taps): Update return type for `backoff_max_tries` to reflect it accepts a callable that returns an integer

### LINKED ISSUE #784: Unclear documentation - backoff_max_tries
https://sdk.meltano.com/en/latest/classes/singer_sdk.RESTStream.html#singer_sdk.RESTStream.backoff_max_tries

Here, annotations show that the method should return int. But, description says, it can be set to `None` for retrying indefinitely. 
Can it be updated to include None in the return types or maybe use a -1 for indefinite retries?

rawwar:
closes #784

### DIFF
--- a/singer_sdk/streams/rest.py
+++ b/singer_sdk/streams/rest.py
@@ -19,8 +19,8 @@ from typing import (
 from urllib.parse import urlparse
 
 import backoff
-from backoff._typing import _MaybeCallable
 import requests
+from backoff._typing import _MaybeCallable
 from singer.schema import Schema
 
 from singer_sdk.authenticators import APIAuthenticatorBase, SimpleAuthenticator
--- a/singer_sdk/streams/rest.py
+++ b/singer_sdk/streams/rest.py
@@ -20,7 +20,6 @@ from urllib.parse import urlparse
 
 import backoff
 import requests
-from backoff._typing import _MaybeCallable
 from singer.schema import Schema
 
 from singer_sdk.authenticators import APIAuthenticatorBase, SimpleAuthenticator
@@ -33,6 +32,8 @@ DEFAULT_PAGE_SIZE = 1000
 DEFAULT_REQUEST_TIMEOUT = 300  # 5 minutes
 
 _TToken = TypeVar("_TToken")
+_MaybeCallable = Union[T, Callable[[], T]]
+
 
 
 class RESTStream(Stream, Generic[_TToken], metaclass=abc.ABCMeta):
--- a/singer_sdk/streams/rest.py
+++ b/singer_sdk/streams/rest.py
@@ -32,7 +32,7 @@ DEFAULT_PAGE_SIZE = 1000
 DEFAULT_REQUEST_TIMEOUT = 300  # 5 minutes
 
 _TToken = TypeVar("_TToken")
-_MaybeCallable = Union[T, Callable[[], T]]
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
Hi @rawwar, thanks for contributing!

It looks like the version of `backoff` used by the Singer SDK (`1.8.0`) doesn't have the `_typing` module 🤷‍♂️.

It's still easy to recreate `_MaybeCallable` from the typing primitives, so you can do that instead.

rawwar:
I am not sure what to update in the docstring. @edgarrmondragon , should I leave it as `int` or change it to `Optional[_MaybeCallable[int]]` or `int or Callable`

singer_sdk/streams/rest.py
```diff
@@ -32,6 +32,7 @@
 DEFAULT_REQUEST_TIMEOUT = 300  # 5 minutes
 
 _TToken = TypeVar("_TToken")
+_MaybeCallable = Union[_TToken, Callable[[], _TToken]]
```
edgarrmondragon, line 35:
```suggestion
_T = TypeVar("_T")
_MaybeCallable = Union[_T, Callable[[], _T]]
```

singer_sdk/streams/rest.py
```diff
@@ -542,7 +543,7 @@ def backoff_wait_generator(self) -> Callable[..., Generator[int, Any, None]]:
         """
         return backoff.expo(factor=2)  # type: ignore # ignore 'Returning Any'
 
-    def backoff_max_tries(self) -> int:
+    def backoff_max_tries(self) -> Optional[_MaybeCallable[int]]:
         """The number of attempts before giving up when retrying requests.
 
         Setting to None will retry indefinitely.
```
edgarrmondragon, line 549:
```suggestion
        Can be an integer, a zero-argument callable that returns an integer,
        or ``None`` to retry indefinitely.
```

edgarrmondragon (commented):
@rawwar this looks good!

> should I leave it as int or change it to Optional[_MaybeCallable[int]] or int or Callable

I'd just remove the type hint from the docstring altogether:

```
        """
        ...
        Returns:
            Number of max retries, callable or ``None``.
        """
```

rawwar:
i'm a bit confused here. I removed the return from docstring and i didn't get anything in the documentation. it didn't auto-generate.

edgarrmondragon:
> i'm a bit confused here. I removed the return from docstring and i didn't get anything in the documentation. it didn't auto-generate.

@rawwar sorry, I probably wasn't too clear. I meant getting rid of the type hint inside the docstring, and just rely on the actual static type annotation you already have in there. You're right that it removes the `Return Type`, though it would still be documented in the signature:

Screen Shot 2022-08-01 at 20 32 55

If you think it's better to keep the (slightly redundant imho) `Return Type`, you can make the following change:

```diff
      def backoff_max_tries(self) -> Optional[_MaybeCallable[int]]:
          """The number of attempts before giving up when retrying requests.
  
          Can be an integer, a zero-argument callable that returns an integer,
          or ``None`` to retry indefinitely.
  
-         Returns:
-             int: limit
+         Returns:
+             Optional[Union[int, Callable[[], int]]]: Number of max retries, callable or
+             ``None``.
          """
          return 5
```

edgarrmondragon:
@rawwar thank you!

rawwar:
Sorry. Was a bit occupied since couple of weeks. Thanks for updating and merging!

PR #887, opened 2022-08-05
fix: Make `conform_record_data_types` work on nested objects and arrays

### LINKED ISSUE #800: `conform_record_data_types` is not recursive
_typing.py has the function `conform_record_data_types`.
But it only iterates over the top level properties of the object, not any nested properties.

This has two negative consequences;
- Types are not normalised for nested values (For example, I am outputting lists of objects with dates on, and those dates are not normalised to strings by the sdk. This causes an error when it later tries to convert them to json).
- The check that a property exists in the schema would not be performed for any nested fields either

Jack-Burnett:
Closes #800

conform_record_data_types is responsible for converting values into json-legal form and removing fields that are not in the catalog.

Previously conform_record_data_types only worked on primitive fields on the given record.
This PR expands it to also work on lists of primitives, nested objects, and lists of objects.

Since the options available for a list are pretty wide in jsonschema, this PR only addresses lists of uniform type (e.g. a single schema applies to the whole list, as opposed to different schemas for different indices), as this is the most common type in my experience.

When logging a list of fields it has removed, it will log them with their path, e.g. "base_obect_name.child_object.primitive_value".


Jack-Burnett:
Dependency review is failing with;
`
An error occurred when accessing the FOSSA API.

      Error message from API:

          Sorry, we've encountered an unexpected error. We've recorded it. If you continue to see this, contact support at support@fossa.com and include the unique reference code: 9ddc1de8-2f38-4ccc-8fd4-c1e98e4d871f
`
Not sure if there is anything I need to do, or if this is due to the service it's calling having an issue?

aaronsteers:
@Jack-Burnett - Thanks very much for this contribution! @edgarrmondragon is primary approver for this. Since this is a non-trivial change, it may require a few days and up to a week for review/discussion.

@edgarrmondragon - Let me know if I can assist in any way or if you have questions.

edgarrmondragon (approved):
@dependabot merge

Jack-Burnett:
Hey @edgarrmondragon, I saw you approved this but didn't merge it - just wanted to check if you wanted any changes, or is it still being reviewed? No worries either way

tests/core/test_typing.py
```diff
@@ -0,0 +1,198 @@
...
+
+    record = {"object": {"value": b"\x01"}}
+
+    expected_output = {"object": {"value": True}}
+
+    actual_output = conform_record_data_types("test_stream", record, schema, logger)
+    assert actual_output == expected_output
+
+
+class TestSimpleEval(unittest.TestCase):
```
edgarrmondragon, line 104: @Jack-Burnett is there a reason to nest tests inside this class?

singer_sdk/helpers/_typing.py
```diff
@@ -275,46 +289,101 @@ def conform_record_data_types(  # noqa: C901
...
-                boolean_representation = False
-            else:
-                boolean_representation = True
-            rec[property_name] = boolean_representation
+        if isinstance(elem, list) and is_uniform_list(property_schema):
+            item_schema = property_schema["items"]
+            output = []
+            for item in elem:
+                if is_object_type(item_schema) and isinstance(item, dict):
+                    output_item, sub_unmapped_properties = _conform_record_data_types(
```
edgarrmondragon, line 334: Do you think this recursiveness may cause performance issues for deeply nested records or large numbers of records?

If so, what would you say about limiting conforming via a parameter set by the tap developer, perhaps at the tap level?

For example:

```
class MyStream(Stream):
   conform_records_max_depth = 10
```

Jack-Burnett: Type conformance is probably the slowest part of the codebase, even for non-recursive schemas.
So if people were relying on the behaviour of it not conforming nested types for performance reasons, I suppose that could cause performance to go down in some cases.

Still, I don't think you would want to address that by tweaking the max depth, since you probably only want none, root or all.
Perhaps it should instead be an enum;
- Only conform root fields
- Conform all fields recursively
- Do not conform any fields
?

One of the reasons I want this merged to master was actaully because I'd like to start experimenting more in that area of the codebase to see if conformance performance can be improved 😁

tests/core/test_typing.py
Jack-Burnett: Yep, it's required to be able to check logging via
self.assertLogs("log", level="WARN")

singer_sdk/helpers/_typing.py
edgarrmondragon: > Perhaps it should instead be an enum;
> 
> * Only conform root fields
> * Conform all fields recursively
> * Do not conform any fields
>   ?

@Jack-Burnett I like that. Go for it!

Jack-Burnett: Done!

PR #898, opened 2022-08-09
fix: Fix SQL type merging for pre-existing target tables

### LINKED ISSUE #374: SQL Targets: Processing fails when target table already exists
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/377

Originally created by @tbriggs2 on 2022-05-05 16:43:59

---

## Summary
When SQLSink.process_batch writes to a table that already exists it ensures that the column definition is compatible with the latest schema and alters the table's columns if necessary. That logic is failing, at least when using [my MySQL target](https://gitlab.com/tbriggs2/target-mysql/). I believe the error is reproducible with [target-sqlite](https://github.com/MeltanoLabs/target-sqlite) as well (because my MySQL target is based on that...)

## Steps to reproduce
1. Set up a pipeline that loads data into a MySQL table
2. Execute the pipeline (this will create the table and succeed)
3. Execute the pipeline again

## What is the current bug behavior?
Processing fails with an exception coming from the SDK.

## What is the expected correct behavior?
Processing should succeed without error.

## Relevant logs and/or screenshots
```
time=2022-05-05 12:26:14 name=sqlalchemy.engine.Engine level=INFO message=SHOW CREATE TABLE `targetTest`
time=2022-05-05 12:26:14 name=sqlalchemy.engine.Engine level=INFO message=[raw sql] ()
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/target_base.py", line 535, in cli
    target.listen(file_input)
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/io_base.py", line 43, in listen
    self._process_endofpipe()
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/target_base.py", line 275, in _process_endofpipe
    self.drain_all()
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/target_base.py", line 413, in drain_all
    self._drain_all(list(self._sinks_active.values()), self.max_parallelism)
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/target_base.py", line 436, in _drain_all
    self.drain_one(sink)
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/target_base.py", line 430, in drain_one
    sink.process_batch(draining_status)
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/sinks/sql.py", line 106, in process_batch
    self.connector.prepare_table(
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/streams/sql.py", line 596, in prepare_table
    self.prepare_column(
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/streams/sql.py", line 621, in prepare_column
    self._adapt_column_type(
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/streams/sql.py", line 774, in _adapt_column_type
    compatible_sql_type = self.merge_sql_types([current_type, sql_type])
  File "/home/tom/.cache/pypoetry/virtualenvs/singer-sqlite-oMbmf0Lm-py3.8/lib/python3.8/site-packages/singer_sdk/streams/sql.py", line 689, in merge_sql_types
    raise ValueError(
ValueError: Unable to merge sql types: INTEGER, INTEGER
```

## Possible fixes
I believe the error is [here](https://gitlab.com/meltano/sdk/-/blob/main/singer_sdk/streams/sql.py#L675). The two types are obviously the same but it's concluding that they're not. I'm not actually sure how to fix it though.

BuzzCutNorman:
This is a attempt at fixing issue 374.  The solution looks to see if the current table column type can accommodate the new type and if it can't it calls the `merge_sql_type()` function and alters the table as before.

closes #374


edgarrmondragon:
Hi @BuzzCutNorman, this came up while working on https://github.com/meltano/sdk/pull/904. This seems like an important bug so thanks for tackling it!

Your patch does seem to work, so let me know when it's ready for the team to review 😄

cc @aaronsteers

BuzzCutNorman:
Hi @edgarrmondragon, Yay!!! I am glad to hear it worked for you.  🥳 I found after some testing there is some code I need to remove.  I won't be able to get to it today but am planning to clean it up Monday.  

cc @aaronsteers

visch:
> Hi @edgarrmondragon, Yay!!! I am glad to hear it worked for you. 🥳 I found after some testing there is some code I need to remove. I won't be able to get to it today but am planning to clean it up Monday.
> 
> cc @aaronsteers

I like what you did better than my idea! Took me a second to understand it. I'll play with this fix in `target-postgres` I'm a little concerned about the type checking mechanism but I need to play with it a little, generally it looks good (It's hard for me to say yes lets do this without a few examples in the wild working, so I may just be a bit timid)

BuzzCutNorman:
Yes, please test a ton.  I did some checks with the StackOverflow database but that doesn't have that many column types.  I just removed the `issubclass` check since that one never got used and if I forced it the conversion didn't work.  I think I misunderstood how a TypeEngine subclass worked.

singer_sdk/streams/sql.py
```diff
@@ -825,9 +830,33 @@ def _adapt_column_type(
         Raises:
             NotImplementedError: if altering columns is not supported.
         """
-        current_type = self._get_column_type(full_table_name, column_name)
+        current_type: sqlalchemy.types.TypeEngine = self._get_column_type(
+            full_table_name, column_name
+        )
+
+        # Check if the existing column type and the sql type are the same
+        if isinstance(sql_type, type(current_type)):
```
visch, line 838: Played with this a little bit

```
 if isinstance(sql_type, type(current_type)):
 ```
 
 ```
 (Pdb) print(f"{sql_type=},{str(sql_type)=}. {current_type=}, {str(current_type)=}")
sql_type=VARCHAR(),str(sql_type)='VARCHAR'. current_type=VARCHAR(), str(current_type)='VARCHAR'
(Pdb) print(type(sql_type))
<class 'sqlalchemy.sql.sqltypes.VARCHAR'>
```
This will work for our "generic" types of VARCHAR. If we get to caring about varchar(6) vs varchar(1000) and needing to change this will present an issue but for the generic SDK this might be alright! 

This is where `str(sql_type)` might be better. Ideally `__eq__` in this case would work for us, but SQLAlchemy overrides that for some other purposes (I'm not well versed enough in SQLAlchemy to understand it all yet)

singer_sdk/streams/sql.py
```diff
@@ -825,9 +830,33 @@ def _adapt_column_type(
...
+            # The current column and sql type are the same
+            # Nothing to do
+            return
+        # Check if the existing column generic type
+        # and sql generic type are the same
+        elif isinstance(sql_type.as_generic(), type(current_type.as_generic())):
```
visch, line 844: I'm not super clear on when `as_generic` would be useful. With something like NVARCHAR in postgres you get

```
(Pdb) abca=NVARCHAR(length=100)
(Pdb) abca
NVARCHAR(national=True, length=100, _warn_on_bytestring=True, _expect_unicode=True)
(Pdb) str(abca)
'NVARCHAR(100)'
(Pdb) abca.as_generic()
Unicode(length=100)
```

Not sure when this comes in handy for type checking? Maybe when comparing to json types, but in the unicode context it doesn't make a lot of sense

singer_sdk/streams/sql.py
BuzzCutNorman: I did have `type(sql_type) == type(current_type)` initally but flake8 during pre-commit checks asked me to use `isinstance` when comparing class types.  I didn't think to convert it to a str().  I am not sure about varchar(6) vs varchar(1000).  I will have to work with those a little to see.

singer_sdk/streams/sql.py
```diff
@@ -825,9 +830,33 @@ def _adapt_column_type(
...
+            # a conversion on insert will happen
+            # Nothing to do
+            return
+
+        # Not the same type, generic type or compatible types
+        # calling merge_sql_types for assistnace
         compatible_sql_type = self.merge_sql_types([current_type, sql_type])
-        if current_type == compatible_sql_type:
+
+        if compatible_sql_type == current_type:
```
visch, line 853: I don't understand SQL Alchemies `__eq__` override on Types enough yet for this.

singer_sdk/streams/sql.py
BuzzCutNorman: I am using this to remove dialect specifics.  There were instances where the target `current_type` was `postgres` specific.  The `as_generic()` gives back the base TypeEngine the dialect specific version is based on.  The thought is if you have a `INT` column in `mysql` it will have the same generic class base as a `postgres`.  That is my very basic understanding of how it works at least.

visch: I understand and I'm not sure why that matters when we're only comparing data types from a single target with each target that's written. 🤷  I saw it in other places too I just don't understand why I guess there may be a great reason!

visch:
Tried running this with a test on `target-postgres` And I hit an odd error @BuzzCutNorman   see https://github.com/MeltanoLabs/target-postgres/issues/9#issuecomment-1215591628

singer_sdk/streams/sql.py
BuzzCutNorman: I think I see why.  The `_get_column_type()` is based on `get_table_columns()' which gives back a dictionary that has a `column name` and the `sqlalchemy.Column` .  It then returns the a `TypeEngine` cast as that `sqlalchemy.Column` which would be dialect specific.  Then you need to use `as_generic()` to bring them back to the base TypeEngine to compare.

singer_sdk/streams/sql.py
BuzzCutNorman: If you are asking why I swapped them.  It was purely to keep things on the same side.  I had been putting the `current_type` on the righthand side above and just wanted to keep the visual continuity.

BuzzCutNorman:
@edgarrmondragon @aaronsteers  Looks like there are still issues with this approach.  I am going to put this back to draft until I can look at the areas @visch highlighted.

BuzzCutNorman:
### Reviewing ###

> 
>  Played with this a little bit
>  
> ```
>   if isinstance(sql_type, type(current_type)):
>  (Pdb) print(f"{sql_type=},{str(sql_type)=}. {current_type=}, {str(current_type)=}")
>  sql_type=VARCHAR(),str(sql_type)='VARCHAR'. current_type=VARCHAR(), str(current_type)='VARCHAR'
>  (Pdb) print(type(sql_type))
>  <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
> ```
>  This will work for our "generic" types of VARCHAR. If we get to caring about varchar(6) vs varchar(1000) and needing to change this will present an issue but for the generic SDK this might be alright!
>  
>  This is where str(sql_type) might be better. Ideally __eq__ in this case would work for us, but SQLAlchemy overrides that for some other purposes (I'm not well versed enough in SQLAlchemy to understand it all yet)

### Setup ###
My source is 5000 row version of the User table from the StackOverflow database on a mssql server 2019 server
My target is a blank datawarehouse database on a postgres14 server

I ran  a `meltano run` job that tapped the `UserTest` table on my source  and created a `UserTest` table on the target.  In mssql the table looks like this.

```
CREATE TABLE [dbo].[UsersTest](
	[Id] [int] IDENTITY(1,1) NOT NULL,
	[AboutMe] [nvarchar](max) NULL,
	[Age] [int] NULL,
	[CreationDate] [datetime] NOT NULL,
	[DisplayName] [nvarchar](40) NOT NULL,
	[DownVotes] [int] NOT NULL,
	[EmailHash] [nvarchar](40) NULL,
	[LastAccessDate] [datetime] NOT NULL,
	[Location] [nvarchar](100) NULL,
	[Reputation] [int] NOT NULL,
	[UpVotes] [int] NOT NULL,
	[Views] [int] NOT NULL,
	[WebsiteUrl] [nvarchar](200) NULL,
	[AccountId] [int] NULL
) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY]
GO
```
The newly created table in postgres looks like this.
```
CREATE TABLE IF NOT EXISTS public."UsersTest"
(
    "Id" integer,
    "AboutMe" character varying COLLATE pg_catalog."default",
    "Age" integer,
    "CreationDate" timestamp without time zone,
    "DisplayName" character varying COLLATE pg_catalog."default",
    "DownVotes" integer,
    "EmailHash" character varying COLLATE pg_catalog."default",
    "LastAccessDate" timestamp without time zone,
    "Location" character varying COLLATE pg_catalog."default",
    "Reputation" integer,
    "UpVotes" integer,
    "Views" integer,
    "WebsiteUrl" character varying COLLATE pg_catalog."default",
    "AccountId" integer
)

```
I added the following logging in the `_adapt_column_type` function to assist in reviewing the difference between `str()` and `type()`
```
       if isinstance(sql_type, type(current_type)):
            # The current column and sql type are the same
            # Nothing to do
            self.logger.info("!!! I found a same to same match. !!!")
            self.logger.info(f"sql_type: {str(sql_type)} current_type class {str(type(current_type))} ")
            self.logger.info(f"sql_type: {str(sql_type)} current_type class {str(current_type)} ")
            return
``` 
```
        elif isinstance(sql_type.as_generic(), type(current_type.as_generic())):
            # The current column and sql generic types are the same
            # Nothing to do
            self.logger.info("!!! I found a generic to generic match. !!!")
            self.logger.info(f"sql_type: {str(sql_type.as_generic())} current_type class {str(type(current_type.as_generic()))} ")
            self.logger.info(f"sql_type: {str(sql_type)} current_type class {str(type(current_type))} ")
            self.logger.info(f"sql_type: {str(sql_type)} current_type class {str(current_type)} ")
```
```
        compatible_sql_type = self.merge_sql_types([current_type, sql_type])

        self.logger.info("!!! I asked merge_sql_type for a merge match. !!!")
        self.logger.info(f"sql_type: {str(sql_type)} current_type class {str(type(current_type))} compatible_sql_type {str(type(compatible_sql_type))} ")
        self.logger.info(f"sql_type: {str(sql_type)} current_type class {str(current_type)} compatible_sql_type {str(compatible_sql_type)}")

        if compatible_sql_type == current_type:
            # Nothing to do
            self.logger.info("!!! I believe these to be equal types !!!")
            self.logger.info(f"current_type class {str(current_type)} compatible_sql_type {str(compatible_sql_type)}")
```
### Testing ###
First thing I noticed is the table when created is not copying over the length of varchar columns present in the source to the target.  

I will truncate the `UsersTest` table on the target side and run the `meltano run` again to get and review the logging info that is returned.
```
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER
level=INFO message=!!! I found a generic to generic match. !!!
level=INFO message=sql_type: DATETIME current_type class <class 'sqlalchemy.sql.sqltypes.DateTime'>
level=INFO message=sql_type: TIMESTAMP current_type class <class 'sqlalchemy.dialects.postgresql.base.TIMESTAMP'>
level=INFO message=sql_type: TIMESTAMP current_type class TIMESTAMP
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'> 
level=INFO message=sql_type: INTEGER current_type class INTEGER 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'> 
level=INFO message=sql_type: VARCHAR current_type class VARCHAR 
level=INFO message=!!! I found a generic to generic match. !!! 
level=INFO message=sql_type: DATETIME current_type class <class 'sqlalchemy.sql.sqltypes.DateTime'>
level=INFO message=sql_type: TIMESTAMP current_type class <class 'sqlalchemy.dialects.postgresql.base.TIMESTAMP'>
level=INFO message=sql_type: TIMESTAMP current_type class TIMESTAMP
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'> 
level=INFO message=sql_type: INTEGER current_type class INTEGER 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'> 
level=INFO message=sql_type: VARCHAR current_type class VARCHAR 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'> 
level=INFO message=sql_type: INTEGER current_type class INTEGER 
```
The majority of matches where `same to same` except the `TIMESTAMP`  columns which were `generic to generic` matches.  Looking to the fourth logging line we can see that evaluating `str()` values would have give us a `same to same match` up front.

Lets pick on the `DisplayName` column which is a `nvarchar(40)` on the source.  On the target I am going to truncate the table and change it to be a `varchar(40)` just like the source and do a `meltano run`.
```
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR(40)
```
They were a `same to same` match but would have failed if evaluating string to string.  On the target I am going to truncate the table and change it to be a `varchar(5)` which is to small for the source data and do a `meltano run`.
```
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'> 
level=INFO message=sql_type: VARCHAR current_type class VARCHAR(5) 
```
It is still seen as a `same to same` match which allowed `meltano run` to run head long into a write failure.  The string to string evaluation would have failed this out.  I am going to switch to the following string to string check and comment out the `generic to generic` check.

```
       # Check if the existing column type and the sql type are the same
        if str(sql_type) == str(current_type):
```
On the target I dropped the `UsersTest` table ran the `meltano run` and truncated the table. 
```
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: TIMESTAMP current_type class <class 'sqlalchemy.dialects.postgresql.base.TIMESTAMP'>
level=INFO message=sql_type: TIMESTAMP current_type class TIMESTAMP
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: TIMESTAMP current_type class <class 'sqlalchemy.dialects.postgresql.base.TIMESTAMP'>
level=INFO message=sql_type: TIMESTAMP current_type class TIMESTAMP 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'> 
level=INFO message=sql_type: VARCHAR current_type class VARCHAR 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'> 
level=INFO message=sql_type: INTEGER current_type class INTEGER 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'> 
level=INFO message=sql_type: INTEGER current_type class INTEGER 
level=INFO message=!!! I found a same to same match. !!! 
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'> 
level=INFO message=sql_type: INTEGER current_type class INTEGER 
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR
level=INFO message=!!! I found a same to same match. !!!
level=INFO message=sql_type: INTEGER current_type class <class 'sqlalchemy.sql.sqltypes.INTEGER'>
level=INFO message=sql_type: INTEGER current_type class INTEGER
```
Everything was a `same to same` match which is nice.  the `TIMESTAMP` got evaluated as the same right off the bat and didn't need a second check.   Lets truncate the `UsersTest` tabele and change `DisplayName` to a `varchar(40)` .
```
level=INFO message=sql_type: TIMESTAMP current_type class TIMESTAMP
level=INFO message=!!! I asked merge_sql_type for a merge match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'> compatible_sql_type <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR(40) compatible_sql_type VARCHAR(40)
level=INFO message=!!! I believe these to be equal types !!!
level=INFO message=current_type class VARCHAR(40) compatible_sql_type VARCHAR(40)
```
Since it failed the `same to same` match it called the `merge_sql_types` function which is great this make more sense than the previous `isinstance()` evaluation flow.  The `current_type` and `compatible_sql_type` pass the `same to same` evaluation which we see in the last two lines.  Last let truncated `UsersTest` again and change `DisplayName` to a `varchar(5)`.
```
level=INFO message=!!! I asked merge_sql_type for a merge match. !!!
level=INFO message=sql_type: VARCHAR current_type class <class 'sqlalchemy.sql.sqltypes.VARCHAR'> compatible_sql_type <class 'sqlalchemy.sql.sqltypes.VARCHAR'>
level=INFO message=sql_type: VARCHAR current_type class VARCHAR(5) compatible_sql_type VARCHAR(5)
level=INFO message=!!! I believe these to be equal types !!!
level=INFO message=current_type class VARCHAR(5) compatible_sql_type VARCHAR(5)
```
Again the initial `same to same` check fails and the `merge_sql_types` gets called.  I don't agree with the `merge_sql_type` solution but I like how I got to this failure.

I am going to change to a single string to string check ( `if str(sql_type) == str(current_type):` ).  The logging of these checks I found handy and was wonder if I should leave a terse version of them in at a `debug` level.

singer_sdk/streams/sql.py
BuzzCutNorman: I switched this to a str() to st() evaluation to keep consistent.  The added benefit is I believe we are steersing clear of the `__eq__` override that sqlalchemy has in place.

BuzzCutNorman:
@edgarrmondragon I have completed reviewing the comments @visch made during his review and made changes to address them.  I also attempted a refresh of `merge_sql_types`.  I am figuring that `merge_sql_types` will only be called when the column types are different.  It does a loop to see if any of types in the sorted type list `sql_types` will work.  If there is a string or byte type class or subclass and it the length is None or 0 it will return it.  It will also return if the type is a similar class type to the `current_type`. The main senario for the similar class type testing on postgres was `TIMESTAMP` vs `DATETIME`.   Please review this when you get a chance.  

I also have a couple of questions: 

1. What should are the minimum column adaption should the SDK do by default?  Which is a question form of  @visch comment from his review. 

>  If we get to caring about varchar(6) vs varchar(1000) and needing to change this will present an issue but for the generic SDK this might be alright!

2. Is the `merge_column_type` a function developers should be directed to override to match merge behavior of the SQL RDMS they are using?

aaronsteers:
@BuzzCutNorman - 

> @edgarrmondragon I have completed reviewing the comments @visch made during his review and made changes to address them. I also attempted a refresh of `merge_sql_types`. I am figuring that `merge_sql_types` will only be called when the column types are different. It does a loop to see if any of types in the sorted type list `sql_types` will work. If there is a string or byte type class or subclass and its length is None or 0 it will return it. It will also return if the type is a similar class type to the `current_type`. The main senario for the similar class type testing on postgres was `TIMESTAMP` vs `DATETIME`. Please review this when you get a chance.
> 
> I also have a couple of questions:
> 
> 1. What are the minimum column adaptions the SDK do should do by default?  Which is a question form of  @visch comment from his review.
> 
> > If we get to caring about varchar(6) vs varchar(1000) and needing to change this will present an issue but for the generic SDK this might be alright!

I'm not sure I fully understand the question, but in case it helps, I _do_ think the SDK should (eventually) know how to auto-expand from `varchar(6)` to `varchar(1000)`. Does that address your question?

A bit of a tangent, but it could also nice if the developer can choose to _always_ bias towards larger data types if they prefer, something like `varchar(MAX)` to prevent future resizing on columnar database targets where storage consumption is identical anyway.

> 2. Is the `merge_column_type` a function developers should be directed to override to match merge behavior of the SQL RDMS they are using?

I think the goal here with `merge_column_type()` would be that (over time) we tackle as many generic cases as possible. Since we inevitably will not get perfect coverage over night, and since some databases will have non-standard types and conversion rules, I have imagined this as the developer pretty much always having a `merge_column_types()` implementation and that they'd handle zero or more special cases before then invoking the base class's implementation for everything else. 

What do you think?

edgarrmondragon:
> I think the goal here with `merge_column_type()` would be that (over time) we tackle as many generic cases as possible. Since we inevitably will not get perfect coverage over night, and since some databases will have non-standard types and conversion rules, I have imagined this as the developer pretty much always having a `merge_column_types()` implementation and that they'd handle zero or more special cases before then invoking the base class's implementation for everything else.

@aaronsteers I agree. Once developers start filing bugs and feature requests, we can start figuring what parts make sense to abstract away from them and which to expose in the "public" API. I think that's been the general approach for the trickier or more niche features in the SDK. For now, as rule of thumb, we could be contempt with supporting SQLite and Postgres DDL.

BuzzCutNorman:
@aaronsteers,  @edgarrmondragon

Thank you for answering my questions.  Yes, you did help me understand better how you want to shape target table columns to match a source, how developers should interact and extend this process, and the dialects you initially want to support.  This all sounds great to me. 

The only item not touch on was sql enginge conversions.  Examples are integers inserted into a varchar column or integers inserted into a numeric column.  The conversions that happen automagically when working with sql.  

Please let me know if you want to move forward with the code as it is now and work on the items above in subsequent PRs or would you like me to attempt to incorporate all or some of the items you mentioned into this PR?

aaronsteers:
> @aaronsteers, @edgarrmondragon
> 
> Thank you for answering my questions. Yes, you did help me understand better how you want to shape target table columns to match a source, how developers should interact and extend this process, and the dialects you initially want to support. This all sounds great to me.

🙌 🙏 

> The only item not touch on was sql enginge conversions. Examples are integers inserted into a varchar column or integers inserted into a numeric column. The conversions that happen automagically when working with sql.

Agreed, this part is pretty hard. Generally, I think of this as `varchar` being "able" to hold an `int` (or `datetime`, etc.) but the reverse is not necessarily true. This means, if we take a pure "ALTER" approach, we may end up with strings if a column morphs across reasonable type barriers. Other implementations will sometimes "rename and stash" the old column in order to avoid this. So, if a `last_activity` column was previously `datetime` type and is becoming `int`, we could rename the current column to `last_activity__bak_20220801` or similar and then add a new column `last_activity` as the `int` type.

This more advanced accommodation though, is not something we have to tackle right away. And arguably, there are strong benefits to keeping all historic data in the same column even if it makes for some awkward type casing.

> Please let me know if you want to move forward with the code as it is now and work on the items above in subsequent PRs or would you like me to attempt to incorporate all or some of the items you mentioned into this PR?

I love what you have so far and I don't think we can/should try to solve every case at once. @edgarrmondragon - I'll look to you for the code signoff/approval side. If this is working and provides and improvement over what we had previously, it seems to be to be is a valuable + stable increment.

edgarrmondragon (approved):
@BuzzCutNorman this looks great! Thanks for taking the time to contribute 😄. Codecov's complaining about a couple of branches not being covered but they're alright to me.

PR #900, opened 2022-08-09
docs: Generate documentation for constructor parameters

edgarrmondragon:
- Document method parameters.
- Add `nox` task for live-reloading docs.

Current state: https://sdk.meltano.com/en/latest/classes/singer_sdk.Stream.html#singer_sdk.Stream

vs

Preview: https://meltano-sdk--900.org.readthedocs.build/en/900/classes/singer_sdk.Stream.html#singer_sdk.Stream

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -104,14 +104,10 @@ ReadtheDocs.org. When a push is detected by readthedocs.org, they automatically 
 and republish the docs. ReadtheDocs is also version aware, so it retains prior and unreleased
 versions of the docs for us.
 
-To build the docs:
+To build the docs and live-reload them locally:
 
 ```bash
-# Build docs
-nox -rs docs
-
-# Open in the local browser:
-open build/index.html
+nox -rs docs-serve
 ```
 
 Sphinx will automatically generate class stubs, so be sure to `git add` them.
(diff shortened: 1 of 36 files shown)


WillDaSilva (approved):
Is it possible to render both the signatures and the detailed parameter / return value content? If so I think that would be ideal. In either case, this looks good to me.

edgarrmondragon:
> Is it possible to render both the signatures and the detailed parameter / return value content? If so I think that would be ideal. In either case, this looks good to me.

@WillDaSilva good idea! I think I've solved that now by explicitly requesting `__init__` gets the same treatment as other methods using `:special-members: __init__`.

WillDaSilva:
@edgarrmondragon Could you please provide an updated preview link? I'm not sure how you generated it - I don't see any CI actions for it.

edgarrmondragon:
> @edgarrmondragon Could you please provide an updated preview link? I'm not sure how you generated it - I don't see any CI actions for it.

@WillDaSilva It's up-to-date with the changes in this PR. Builds are triggered in [readthedocs](https://readthedocs.org/projects/meltano-sdk/builds/) but it's failing to report back to GitHub in the PR checks.

WillDaSilva (approved):

PR #914, opened 2022-08-17
docs: Bump Pygments and update dbt example

edgarrmondragon:
Check with `nox -rs docs-serve`


edgarrmondragon:
Old:

Screen Shot 2022-08-17 at 16 32 46

---

New:

Screen Shot 2022-08-17 at 16 32 41

Issue #901, opened 2022-08-10
[Bug]: It is not possible to make a field "required" in a catalog schema.


Jack-Burnett:
### Singer SDK Version

0.6.1

### Python Version

3.10

### Bug scope

Taps (catalog, state, stream maps, etc.)

### Operating System

MacOS

### Description

I override the;
```
@property
    def schema(self) -> dict:
```
function in a tap stream. I return a json schema directly (e.g. I am not using the PropertiesList utility to help me).

Ideally I would like to make every field on my schema required (aka guaranteed) so downstreams can depend on them.
In fact, this is even something that is done in an example in the singer spec: https://hub.meltano.com/singer/spec#messages

However, when returning that schema, the required attribute is stripped from it internally, making it impossible to have fields be required.

I believe the root cause is that this class forbids it: https://github.com/transferwise/pipelinewise-singer-python/blob/da64a10cdbcad48ab373d4dab3d9e6dd6f58556b/singer/schema.py#L108.
as was the case with this ticket: https://github.com/meltano/sdk/issues/332

edgarrmondragon:
Fix might be to subclass `singer.schema.Schema` in https://github.com/meltano/sdk/blob/eb024bd910537d6bfb2bddf031633ce273bd5fe7/singer_sdk/helpers/_singer.py#L1-L8

and add the `required` field so SerDe respects it.

Related to #332, which suffers from the same problem.

Jack-Burnett:
Yeah that seems like probably the best way 👍

Issue #332, opened 2022-02-23
Support enumerated values in SDK JSON schema helpers


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/334

Originally created by @edgarrmondragon on 2022-02-23 17:36:37

---

## Summary
[//]: # (Concisely summarize the feature you are proposing.)

Support `enum` in JSON Schema helpers.

## Proposed benefits
[//]: # (Concisely summarize the benefits this feature would bring to yourself and other users.)

Validation of enumerated values in records and package configurations.

## Proposal details
[//]: # (In as much detail as you are able, describe the feature you'd like to build or would like to see built.)

Add an `enum` keyword argument to the `Property` constructor.

Docs: https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

## Best reasons not to build
[//]: # (Will this negatively affect any existing functionality? Do you anticipate any breaking changes versus what may already be working today? Make the counter-argument to your proposal here.)

`CustomType` is a way to do this if users don't mind typing a raw JSON Schema dictionary.

MeltyBot:
[View 7 previous comments from the original issue on GitLab](https://gitlab.com/meltano/sdk/-/issues/334)

ReubenFrankel:
https://gitlab.com/meltano/sdk/-/issues/334#note_959265111

> Hey, I was wanting to have a catalog field of type enum for a tap I'm working on; The OP here states "`CustomType` is a way to do this if users don't mind typing a raw JSON Schema dictionary." but I do not think that is correct; because `schema.py` does not support enum in it's list of STANDARD_KEYS, it does not seem to be actually possible to workaround this by using a CustomType :disappointed:
>
> Since the comments here seem to be more focussed on expanding the list of formats, I was wondering what the status was of the original issue, regarding adding enum support?
>
> Thanks

I have successfully added [enum support to a custom type extending from `StringType`](https://github.com/Matatika/tap-auth0/blob/3b184d078a4f012e39c8f4f277a890dd1c87ad07/tap_auth0/types/refresh_token.py#L10) in the past. Not sure how this compares to your implementation with `CustomType`, but it shouldn't be that dissimilar.

edgarrmondragon:
@ReubenFrankel @aaronsteers is the limitation rather coming from

[...]

PR #917, opened 2022-08-19
fix: Allow Singer schemas to include the `required` and `enum` fields

Jack-Burnett:
Closes [#901](https://github.com/meltano/sdk/issues/901)

The use of the Schema class from a dependency was limiting schemas to only include certain declared json schema fields, since it would remove properties if they were not whitelisted.  Since this is from a dependency, we can't simply expand the whitelist so instead we override the class and expand the list there. An alternative implementation would be to simply replace the class outright (no inheritance) - this could break existing taps that pass a Schema to Stream.init though.

It helps with but does not close [#332](https://github.com/meltano/sdk/issues/332), since 332 is asking for helpers, which this does not add - it makes it possible to add them though.

I think there are lots more improvements that could be made to SchemaPlus in follow-up PRs (adding many more legitimate properties to the whitelist and having better support for some non-simple properties.

Also, I removed the 'selected' and 'inclusion' fields from the new class - because they are part of the singer spec and not actually valid fields to see on a json schema.

### DIFF
--- a/singer_sdk/helpers/_schema.py
+++ b/singer_sdk/helpers/_schema.py
@@ -1,4 +1,3 @@
-# pylint: disable=redefined-builtin, too-many-arguments, invalid-name
 """Provides an object model for JSON Schema."""
 
 from dataclasses import dataclass
--- a/singer_sdk/helpers/_schema.py
+++ b/singer_sdk/helpers/_schema.py
@@ -66,10 +66,7 @@ class SchemaPlus(Schema):
         result = {}
 
         if self.properties is not None:
-            result["properties"] = {
-                k: v.to_dict()
-                for k, v in self.properties.items()
-            }
+            result["properties"] = {k: v.to_dict() for k, v in self.properties.items()}
 
         if self.items is not None:
             result["items"] = self.items.to_dict()
--- a/singer_sdk/helpers/_schema.py
+++ b/singer_sdk/helpers/_schema.py
@@ -1,7 +1,8 @@
 # pylint: disable=redefined-builtin, too-many-arguments, invalid-name
 """Provides an object model for JSON Schema."""
 
-import json
+from dataclasses import dataclass
+from typing import Any, List, Optional, Union
 
 from singer import Schema
 
@@ -29,6 +30,7 @@ STANDARD_KEYS = [
 ]
 
 
+@dataclass
 class SchemaPlus(Schema):  # pylint: disable=too-many-instance-attributes
     """Object model for JSON Schema.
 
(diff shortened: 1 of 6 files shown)


edgarrmondragon:
Hi @Jack-Burnett! Thanks for the PR. I'll take a look in the next few days 😺

singer_sdk/helpers/_schema.py
```diff
@@ -0,0 +1,136 @@
...
+        self.anyOf = anyOf
+        self.format = format
+        self.additionalProperties = additionalProperties
+        self.patternProperties = patternProperties
+        self.required = required
+        self.enum = enum
+        self.title = title
+
+    def __str__(self):
+        return json.dumps(self.to_dict())
```
edgarrmondragon, line 87: This is implemented in the parent class, so you safely can omit this

```suggestion
```

singer_sdk/helpers/_schema.py
```diff
@@ -0,0 +1,136 @@
...
+    "enum",
+    # These are NOT simple keys (they can contain schemas themselves). We could
+    # consider adding extra handling to them.
+    "additionalProperties",
+    "anyOf",
+    "patternProperties",
+]
+
+
+class SchemaPlus(Schema):  # pylint: disable=too-many-instance-attributes
```
edgarrmondragon, line 32: @Jack-Burnett wdyt of making this a `dataclass`, similar to `singer_sdk.helpers._singer.CatalogEntry`?

That'd would save some redundant code in the constructor and give us a nice `__repr__` and `__eq__` for free.

edgarrmondragon (changes requested):

singer_sdk/helpers/_schema.py
Jack-Burnett: Yeah that sounds like a good idea, done.
It also meant I had to give some type hints which are nice too, though I left a couple at Any for the more flexible json-schema fields

singer_sdk/helpers/_schema.py
```diff
@@ -0,0 +1,104 @@
...
+    # These are NOT simple keys (they can contain schemas themselves). We could
+    # consider adding extra handling to them.
+    "additionalProperties",
+    "anyOf",
+    "patternProperties",
+]
+
+
+@dataclass
+class SchemaPlus(Schema):  # pylint: disable=too-many-instance-attributes
```
edgarrmondragon, line 34:
```suggestion
class SchemaPlus(Schema):
```

singer_sdk/helpers/_schema.py
```diff
@@ -0,0 +1,104 @@
...
+        result = {}
+
+        if self.properties is not None:
+            result["properties"] = {
+                k: v.to_dict()
+                for k, v in self.properties.items()  # pylint: disable=no-member
+            }
+
+        if self.items is not None:
+            result["items"] = self.items.to_dict()  # pylint: disable=no-member
```
edgarrmondragon, line 76:
```suggestion
            result["items"] = self.items.to_dict()
```

Jack-Burnett:
> @Jack-Burnett this is looking great!
> 
> I initially missed those pylint comments which we don't need. Once that's addressed this is g2g 🙂

Ohh yeah just lazy copy-pasting on my part 😅 All gone now

PR #924, opened 2022-08-24
ci: Use `meltano/commitizen-version-bump`

WillDaSilva:
To customize how Commitizen performs our version bump, Meltano and the SDK will now use https://github.com/meltano/commitizen-version-bump. It will be installed and used by commitizen-action in the version_bump GitHub workflow.

All of the customizations used by the SDK should continue to work unchanged: https://github.com/meltano/commitizen-version-bump/commit/d049f8eb96d9c2c0718b669c04d491d6bbea0124

Relates to https://github.com/meltano/gitops/issues/4

### DIFF
--- a/.github/workflows/version_bump.yml
+++ b/.github/workflows/version_bump.yml
@@ -55,6 +55,7 @@ jobs:
         push: "false"
         changelog: "true"
         github_token: ${{ secrets.GITHUB_TOKEN }}
+        extra_requirements: 'git+https://github.com/meltano/commitizen-version-bump@main'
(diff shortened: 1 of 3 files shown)


aaronsteers (commented):
Looks good for my part. As per my comment for the related change in `meltano/meltano`, the real bulk of the review will need to happen in the linked repo. But I would support a greenlight to iterate on that repo as needed, for this release, with the proof of success being the successfully run release process in itself.

aaronsteers (approved):

edgarrmondragon (approved):

PR #931, opened 2022-08-29
refactor: Move individual record logic from `Stream._sync_records` into new `Stream._process_record`

edgarrmondragon:


singer_sdk/streams/core.py
```diff
@@ -970,9 +969,45 @@ def finalize_state_progress_markers(self, state: Optional[dict] = None) -> None:
...
     ) -> None:
+        """Process a record.
+
+        Args:
+            record: The record to process.
+            selected: Whether the stream is selected.
+            record_context: The record context.
+            child_context: The child context.
+            partition_context: The partition context.
+            count: The current record count.
```
aaronsteers, line 989: Would be good in the docstring to clarify - is `count` per batch, per stream, per the above contexts, or total?

aaronsteers: I think 'per stream and (if applicable) per parent context or partition' - but I could be wrong on that.

singer_sdk/streams/core.py
```diff
@@ -1153,9 +1175,7 @@ def get_child_context(self, record: dict, context: Optional[dict]) -> dict:
     # Abstract Methods
 
     @abc.abstractmethod
-    def get_records(
-        self, context: Optional[dict]
-    ) -> Iterable[Union[dict, Tuple[dict, dict]]]:
+    def get_records(self, context: Optional[dict]) -> Iterable[dict]:
```
aaronsteers, line 1178: Can you say more about the change in the return signature?

Admittedly this is an awkward design, but for some parent child use cases it may still be in use, per the commend in the below docstring:

> Parent streams can optionally return a tuple, in which case the second item in the tuple being a `child_context` dictionary for the stream's `context`.

Primarily this is used when the child context is some python object that should not be part of the record - such as a python Tableau workbook object in the `tap-tableau` case.

edgarrmondragon: @aaronsteers No you're right. I've restored the signature with the optional `tuple` return value. It was an initial oversight on my part cause I thought the `child_context` returned from `get_records` was never used but it is.

We do seem to be using `copy.copy` a few times very defensively in case developers want (try?) to mutate the context dict.

singer_sdk/streams/core.py
edgarrmondragon: Updated!

edgarrmondragon:
@aaronsteers I'm gonna close this since I'm still iterating on this refactor for BATCH

singer_sdk/streams/core.py
aaronsteers: @edgarrmondragon - 

> We do seem to be using copy.copy a few times very defensively in case developers want (try?) to mutate the context dict.

I wish there was a convenient class for read-only `dict`. We often want to send contexts (and config) around between SDK code and dev-written code, and yes, the key word is 'defensive'. If a dev mutates or alters these, there could be adverse affects.

That said, I'm not going to hold tightly to those `copy` operators if you think we can remove them, I'll be okay with that.

singer_sdk/streams/core.py
aaronsteers: 👍 🙏

aaronsteers:
> @aaronsteers I'm gonna close this since I'm still iterating on this refactor for BATCH

Works for me. I think you can safely just include this as a refactor within the `BATCH` PR. 👍

PR #936, opened 2022-08-31
refactor: Use inheritance to construct plugin CLI

edgarrmondragon:


kgpayne:
@edgarrmondragon have you seen [this](https://github.com/sathoune/typer-oo-example/tree/master/click_example)?

I was wondering if this would allow us to add a new attribute `cli_class` to the `Tap`/`Target` base classes, which users could override with their own subclasses of the default Cli class, and return `self.cli_class().run()` from the existing `cli` attribute, rather than a Click command 🤔 I really like how the approach allows Click commands to be defined as class attributes 😍

edgarrmondragon:
> @edgarrmondragon have you seen [this](https://github.com/sathoune/typer-oo-example/tree/master/click_example)?

Thanks for sharing @kgpayne! After reviewing it, I don't think a pattern like that is something we can adopt without significant changes to the plugin class. A lot of the CLI options we currently support can and should be handled by simple callbacks (as this PR implements) and following the pattern in the link prohibits setting callbacks from class methods for things like version, discovery, etc. It also forces us to implement a God-method that handles all options (the status quo), which gets ugly as exemplified by tap-airbyte's implementation requiring too many copy and pasting.

> I was wondering if a similar approach would allow us to add a new attribute `cli_class` to the `Tap`/`Target` base classes, which users could override with their own subclasses of the default Cli class, and return `self.cli_class().run()` from the existing `cli` attribute, rather than a Click command 🤔 I really like how the approach allows Click commands to be defined as class attributes 😍

I've pushed another commit to the PR that changes the `PluginCLI` descriptor to be a decorator. It enables uses like:

```python
class ShortcutTap(Tap):
    @plugin_cli
    def update_schema(cls) -> click.Command:
        """Update the OpenAPI schema for this tap."""
        @click.command()
        def update():
            response = requests.get(
                "https://developer.shortcut.com/api/rest/v3/shortcut.swagger.json",
                timeout=5,
            )
            with Path("tap_shortcut/openapi.json").open("w") as f:
                f.write(response.text)

        return update
```

Also also, there's nothing stopping users from declaring custom CLIs and registering in `[tool.poetry.scripts]`.

Wdyt?

edgarrmondragon:
> @edgarrmondragon this looks great 🙌 Will do a full review tomorrow, but a docs addition would be helpful to support adoption/experimentation e.g. [MeltanoLabs/target-snowflake#21](https://github.com/MeltanoLabs/target-snowflake/issues/21) 🙂

Thanks for calling out the lack of docs! I've added a short guide in https://meltano-sdk--936.org.readthedocs.build/en/936/guides/custom-clis.html

kgpayne (approved):

kgpayne:
@edgarrmondragon my only thought is that it'd be great to have a more full-featured example for the docs in time. I learned from the Click codebase that `click.Group` is a subclasses of `click.Command` (via `click.MultiCommand`), so it would be possible to ship a multi-command CLI (along the lines of [target-snowflake#21](https://github.com/MeltanoLabs/target-snowflake/issues/21)) in addition to the Singer-compatible CLI 🚀 So there may be future value in standardising on a `<plugin-name>-utils` or `<plugin-name>-tools` entrypoint (rather than an entrypoint/command per util), and maybe advertising the extra functionality as `capabilities` in `--about`. No action needed, just registering the thought to check I've understood the potential here. Happy to try out the pattern in `target-snowflake` before any future docs changes 🙂

PR #968, opened 2022-09-14
feat: Added cookiecutter support and docs for VSCode debugging

kgpayne:
Added module entry point to tap and target cookiecutters, and guidance on how to use them for debugging in VSCode.

----
:books: Documentation preview :books:: https://meltano-sdk--968.org.readthedocs.build/en/968/


cookiecutter/tap-template/{{cookiecutter.tap_id}}/{{cookiecutter.library_name}}/tap.py
```diff
@@ -64,3 +64,6 @@ def discover_streams(self) -> List[Stream]:
         """Return a list of discovered streams."""
         return [stream_class(tap=self) for stream_class in STREAM_TYPES]
 {%- endif %}
+
+if __name__ == "__main__":
+    Tap{{ cookiecutter.source_name }}.cli()
```
edgarrmondragon, line 69:
```suggestion


if __name__ == "__main__":
    Tap{{ cookiecutter.source_name }}.cli()
```

edgarrmondragon (commented):
@kgpayne nice! Left a few suggestions

edgarrmondragon (approved):

edgarrmondragon:
@kgpayne wdyt about adding the `.vscode/launch.json` file to the tap cookiecutter, perhaps conditioned on a new question?

kgpayne:
> @kgpayne wdyt about adding the `.vscode/launch.json` file to the tap cookiecutter, perhaps conditioned on a new question?

I don't have a strong sense of how popular vscode is relative to others like pycharm/vim/sublime/atom 🤔 @tayloramurphy do we have any IDE data from our user surveys?

Either way, I think adding docs/instructions for 1 or 2 other IDE's might be a good next step, then adding a new cookie cutter question when there are a couple of options to choose from 🙂 @pandemicsyn uses pycharm so may be able to contribute a snippet for that, and I believe @cjohnhanson uses vim.

PR #978, opened 2022-09-17
feat: Allow configuring a dedicated metrics logger

edgarrmondragon:
# Summary

## New metrics helpers

The most important difference: the previous implementation only emitted total records at the end of sync, while the new one emits the accumulated record count between regular intervals (60s by default). This allows, among other things, to plot record counts against time for a child and parent stream:

metric

Takes inspiration from [`singer-python`](https://github.com/singer-io/singer-python/blob/master/singer/metrics.py). The new classes actually output time-series data for record count, sync duration and HTTP request duration.

### The `Point` class

This dataclass has a custom `__str__()` method that renders the object as JSON, that means disabling metrics via log config (e.g. level `ERROR`) results in not needing to perform the costly object->string conversion and makes testing metrics easier by capturing the raw object from the `LogRecord` instance.

## Logging configuration

Since metrics are just special logs, I think the best way to configure them is via generic log configuration. So, the SDK now supports an `SINGER_SDK_LOG_CONFIG` env var that points to a logging config file in YAML format (same as meltano).

----
:books: Documentation preview :books:: https://meltano-sdk--978.org.readthedocs.build/en/978/


edgarrmondragon:
pre-commit.ci autofix

aaronsteers:
Exciting!!

cjohnhanson (approved):
Looks good! :rocket:

aaronsteers (approved):

PR #999, opened 2022-09-26
fix: Absolute file paths created by taps running in BATCH mode can't be processed by the Sink

edgarrmondragon:


aaronsteers (approved):
🚀

Issue #333, opened 2022-02-24
Consider making logging more granular


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/335

Originally created by @edgarrmondragon on 2022-02-24 03:15:15

---

## Summary
[//]: # (Concisely summarize the feature you are proposing.)

Leverage Python's [dot-separated logger hierarchy](https://docs.python.org/3/library/logging.html#logging.getLogger) to give developers more control over logging behavior and make it more granular:

- Behavior: handler, format, level
- Source: tap, stream, SDK.

## Proposed benefits
[//]: # (Concisely summarize the benefits this feature would bring to yourself and other users.)

Developers can make output more or less verbose for certain streams, etc.

## Proposal details
[//]: # (In as much detail as you are able, describe the feature you'd like to build or would like to see built.)

![Screen_Shot_2022-02-23_at_21.06.12](http://gitlab.com/meltano/sdk/uploads/fb0c52c2bf98c005a1ddec7d408c804d/Screen_Shot_2022-02-23_at_21.06.12.png)

```python
>>> import logging
>>> formatter = logging.Formatter("{asctime} {levelname} {message}", style="{")
>>> handler = logging.StreamHandler()
>>> handler.setFormatter(formatter)
>>> tap_logger = logging.getLogger("tap-example")
>>> tap_logger.addHandler(handler)
>>> tap_logger.setLevel(logging.INFO)  # Display INFO and up
>>> stream_logger = logging.getLogger("tap-example.my_stream")
>>> stream_logger.setLevel(logging.WARNING)  # Display WARNING and up
>>> tap_logger.debug("Info from tap")
>>> tap_logger.info("Info from tap")
2022-02-23 21:10:34,165 INFO Info from tap
>>> stream_logger.info("Info from stream")  # No output at this level!
>>> stream_logger.warning("Warning from stream")
2022-02-23 21:10:34,175 WARNING Warning from stream
```

## Best reasons not to build
[//]: # (Will this negatively affect any existing functionality? Do you anticipate any breaking changes versus what may already be working today? Make the counter-argument to your proposal here.)

We _may_ not want developers to make their packages to quiet so end-users have useful debugging information by default.

MeltyBot:
[View 2 previous comments from the original issue on GitLab](https://gitlab.com/meltano/sdk/-/issues/335)

PR #1010, opened 2022-09-28
feat: Full url instead of just the path

### LINKED ISSUE #996: [Feature]: When _LOG_REQUEST_METRIC_URLS = True , the url base should be included
### Feature scope

### Description

Right now the metric url looks like this
```
INFO:tap-indeedsponsoredjobs:INFO METRIC: {"type": "timer", "metric": "http_request_duration", "value": 0.14543, "tags": {"endpoint": "/appinfo", "http_status_code": 403, "status": "failed", "url": "/v2/api/appinfo?perPage=1000000000
  2 Traceback (most recent call last):
 ```
 
 In my case (and most of my cases) I'm not worried about leaking sensitive crednetials via the URL, and having the baseURL is helpful. I tend to override this in my taps during debugging and I forget about it until I write a new tap. The line I add is here https://github.com/meltano/sdk/blob/main/singer_sdk/streams/rest.py#L247 I add base_url to the extra tags. 
 
This use case comes up with lots of Rest APIS because you have a different base URL for authentication vs the "data" api you'll hit later. Another large subset is when there's a number of different "data" apis you need to hit.

### PR DESCRIPTION
**Pull request by visch**
Closes #996 

----
:books: Documentation preview :books:: https://meltano-sdk--1010.org.readthedocs.build/en/1010/

### DIFF
--- a/singer_sdk/streams/rest.py
+++ b/singer_sdk/streams/rest.py
@@ -244,7 +244,7 @@ class RESTStream(Stream, Generic[_TToken], metaclass=abc.ABCMeta):
         if self._LOG_REQUEST_METRICS:
             extra_tags = {}
             if self._LOG_REQUEST_METRIC_URLS:
-                extra_tags["url"] = prepared_request.path_url
+                extra_tags["url"] = prepared_request.url
             self._write_request_duration_log(
                 endpoint=self.path,
                 response=response,

### CODE REVIEW

File: singer_sdk/streams/rest.py [hunk 1]
```diff
@@ -244,7 +244,7 @@ def _request(
         if self._LOG_REQUEST_METRICS:
             extra_tags = {}
             if self._LOG_REQUEST_METRIC_URLS:
-                extra_tags["url"] = prepared_request.path_url
+                extra_tags["url"] = prepared_request.url
```
edgarrmondragon, line 247: This includes any URL query params, so it could lead to secrets leaking into metrics logs.

Wdyt @aaronsteers?

edgarrmondragon:
From the linked issue:

> This use case comes up with lots of Rest APIS because you have a different base URL for authentication vs the "data" api you'll hit later. Another large subset is when there's a number of different "data" apis you need to hit.

I'm curious @visch, authenticator classes don't log metrics so there's no need for disambiguation using the base URL.

In the second case, where different streams may have different base URLs with the same endpoint (e.g. `users-api.myservice.com/records` and `stats-api.myservice.com/records`), would it be useful to add the stream class name to the tags dict and/or custom metric tags that differentiate the log records instead of including the base URL?

singer_sdk/streams/rest.py
aaronsteers: I wonder if we should just add `RESTStream.url_path_as_sensitive` and `url_params_as_sensitive` and maybe even `url_root_as_sensitive`.

@visch - in our conversation earlier I misunderstood and was thinking that you only wanted to include the base url.

If we want to include arga as well (and I definitely think it's valuable to do so!), then maybe we just ask devs to tell us what can/can't be allowed in logs.

aaronsteers: Also - If we put the args here as well in url, then there might be value in a separate entry for `path` or `endpoint` that excludes them...

edgarrmondragon: @aaronsteers I don't see the URL params as generally useful to include _as a metric tag_, since the main purpose of metrics is not to debug the tap operation, but rather monitor it and analyze its performance.

Maybe in a more generic "The request failed, and these were the request details"-type of debug log message, it'd be useful to include the URL params, but they'd hardly bring any valuable info to annotate the metrics timeseries as is being done here.

visch:
> From the linked issue:
> 
> > This use case comes up with lots of Rest APIS because you have a different base URL for authentication vs the "data" api you'll hit later. Another large subset is when there's a number of different "data" apis you need to hit.
> 
> I'm curious @visch, authenticator classes don't log metrics so there's no need for disambiguation using the base URL.
> 
> In the second case, where different streams may have different base URLs with the same endpoint (e.g. `users-api.myservice.com/records` and `stats-api.myservice.com/records`), would it be useful to add the stream class name to the tags dict and/or custom metric tags that differentiate the log records instead of including the base URL?

The end goal is to be able to debug things when they are failing.  It helps a ton in production to have more information than less information as well so when I get a ticket from people who have an issue they send me the logs and I have a nice way to debug everything. 

Having the stream isn't very helpful for me here as I'm trying to see what's going on, at the end of the day we're making HTTP requests to endpoints I want to see exactly what the HTTP request is. When I care about everything that's happening I switch on https://requests.readthedocs.io/en/latest/api/?highlight=logging#:~:text=import%20requests%0Aimport%20logging by overriding `__init__` 

There's lots of reasons for things to fail, seeing what is actually happening with your HTTP requests lets you debug it without having to memorize how the SDK is displaying that information to you. Metrics is a great way to debug this for me when I'm running things quick and saves me headaches.

I'd be down to a switch for turning on `https://requests.readthedocs.io/en/latest/api/?highlight=logging#:~:text=import%20requests%0Aimport%20logging` via a flag or something, but I think making it clear what requests are happening is the best approach in the logs.

singer_sdk/streams/rest.py
visch: Need a way to debug that's easier than today, I don't really care if it's in a metric or not 🤷

aaronsteers: @edgarrmondragon and @visch - Yeah, I think we're going to be blocked generically from giving robust debug info unless we add something like the sensitivity flags I suggested above.

> I wonder if we should just add
> 
> * `RESTStream.url_path_as_sensitive` and
> * `RESTStream.url_params_as_sensitive` and maybe even
> * `RESTStream.url_root_as_sensitive`.

I'd opt to keep the metrics fairly clean, perhaps adding the base url so it isn't just a partial path.

And I'd probably also opt to taking a pass through the error log and, upon failure, print _as much detail as is not marked sensitive_.

visch: I thought `_LOG_REQUEST_METRIC_URLS` was a sensitivity flag  https://github.com/meltano/sdk/blob/3e819ddccf17eb2431050363caef29966112fdfe/singer_sdk/streams/rest.py#L57

edgarrmondragon:
@visch thanks for the detailed explanation of your use case

> The end goal is to be able to debug things when they are failing. It helps a ton in production to have more information than less information as well so when I get a ticket from people who have an issue they send me the logs and I have a nice way to debug everything.

I agree that such information is useful. Here's a proposal:

- Since the interest is in the requests _made_ and in the responses _received, add a new method `RESTStream.log_request` that's located after https://github.com/meltano/sdk/blob/a06c1f347855bbde28f0552bb167e549bd4d0432/singer_sdk/streams/rest.py#L351-L354

  ```diff
                prepared_request = self.prepare_request(
                    context,
                    next_page_token=paginator.current_value,
                )
  +             self.log_request(prepared_request)
  ```

  A default implementation would be

  ```python
  def log_request(self, prepared_request):
      parsed_url = urlparse(parsed_url)._replace(query="")
      self.logger(
          "Making %s request to %s",
          prepared_request.method,
          parsed_url.geturl(),
      )
  ```

Then developers can override `log_request` if they need more details logged out, but by default the query params are excluded.

---

> When I care about everything that's happening I switch on [requests.readthedocs.io/en/latest/api/?highlight=logging#:~:text=import%20requests%0Aimport%20logging](https://requests.readthedocs.io/en/latest/api/?highlight=logging#:~:text=import%20requests%0Aimport%20logging) by overriding __init__

Yeah, the SDK should really discourage overriding `__init__`, so maybe https://github.com/meltano/sdk/issues/333 is a good way to address logging configuration. Taps could read a log config file path from the environment similar to `singer-python`'s `LOGGING_CONF_FILE`. In that file, all loggers could be set to `DEBUG`.

singer_sdk/streams/rest.py
edgarrmondragon: @visch and @aaronsteers So, looking into this, it seems that `PreparedRequest.path_url` already contains URL params and may leak secrets 🤦‍♂️:

```python
>>> import requests
>>> s = requests.Session()
>>> req = requests.Request(method="GET", url="https://httbin.org/test", params={"secret": "abc"})
>>> prep = s.prepare_request(req)
>>> prep.path_url
'/test?secret=abc'
```

We need to get rid of this from the metric log entirely, or at least parse the url and pass only the path here:

```
from urllib.parse import urlparse

extra_tags["url"] = urlparse(prepared_request.url).path
```

That said, I still think that there should be a dedicated for request details elsewhere.

edgarrmondragon:
See https://github.com/meltano/sdk/issues/996#issuecomment-1360474430

PR #1032, opened 2022-10-04
feat: Support stream property selection push-down in SQL streams

### LINKED ISSUE #974: feat: support stream property selection in SQL streams
### Feature scope

### Description

Currently, for SQL-type streams we do the equivalent of `select * from table` to retrieve records, and rely on the `pop_deselected_record_properties ` helper from the SDK to remove/pop any deselected properties before returning the final `RECORD`. This is OK in the general case, but does mean that we are retrieving more data than necessary, only to filter it locally. However, in the `BATCH` case, records are not processed locally at all, and therefore never have selection criteria applied. Therefore, to solve both requirements, we could add a `get_selected_properties` helper to return the column names of a given stream to be used when constructing the select statements for both `RECORD` and `BATCH` implementations of SQL-type streams.

kgpayne:
Closes #974 

- #974 

----
:books: Documentation preview :books:: https://meltano-sdk--1032.org.readthedocs.build/en/1032/

### DIFF
--- a/singer_sdk/streams/sql.py
+++ b/singer_sdk/streams/sql.py
@@ -510,7 +510,8 @@ class SQLConnector:
                 nullable=col_meta.get("nullable", False),
             )
             for col_meta in columns
-            if not column_names or col_meta["name"] in column_names
+            if not column_names
+            or col_meta["name"].casefold() in {col.casefold() for col in column_names}
         }
 
     def get_table(
--- a/singer_sdk/streams/sql.py
+++ b/singer_sdk/streams/sql.py
@@ -1023,7 +1023,6 @@ class SQLStream(Stream, metaclass=abc.ABCMeta):
         """Return a copy of the Stream JSON schema, dropping any fields not selected.
 
         Returns:
-
             A dictionary containing a copy of the Stream JSON schema, filtered
             to any selection criteria.
         """
--- a/singer_sdk/streams/sql.py
+++ b/singer_sdk/streams/sql.py
@@ -6,12 +6,13 @@ import abc
 import logging
 from datetime import datetime
 from functools import lru_cache
-from typing import Any, Iterable, cast
+from typing import Any, Iterable, List, cast
 
 import sqlalchemy
 from sqlalchemy.engine import Engine
 from sqlalchemy.engine.reflection import Inspector
 
+import singer_sdk.helpers._catalog as catalog
 from singer_sdk import typing as th
 from singer_sdk._singerlib import CatalogEntry, MetadataMapping, Schema
 from singer_sdk.exceptions import ConfigValidationError
@@ -325,11 +326,7 @@ class SQLConnector:
             # Some DB providers do not understand 'views'
             self._warn_no_view_detection()
             view_names = []
-        object_names = [(t, False) for t in table_names] + [
-            (v, True) for v in view_names
-        ]
-
-        return object_names
+        return [(t, False) for t in table_names] + [(v, True) for v in view_names]
 
     # TODO maybe should be splitted into smaller parts?
     def discover_catalog_entry(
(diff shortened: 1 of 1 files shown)


kgpayne:
I think the `pre-commit` action is broken 🤔 The line it refers to has no `typing.List` import and `flake8` doesn't find any lint errors locally 🤷‍♂️

singer_sdk/streams/sql.py
```diff
@@ -6,12 +6,13 @@
 import logging
 from datetime import datetime
 from functools import lru_cache
-from typing import Any, Iterable, cast
+from typing import Any, Iterable, List, cast
```
edgarrmondragon, line 9:
```suggestion
from typing import Any, Iterable, cast
```

singer_sdk/streams/sql.py
```diff
@@ -501,26 +503,31 @@ def get_table_columns(self, full_table_name: str) -> dict[str, sqlalchemy.Column
...
-        for col_meta in columns:
-            result[col_meta["name"]] = sqlalchemy.Column(
+        return {
+            col_meta["name"]: sqlalchemy.Column(
                 col_meta["name"],
                 col_meta["type"],
                 nullable=col_meta.get("nullable", False),
             )
+            for col_meta in columns
+            if not column_names or col_meta["name"] in column_names
```
aaronsteers, line 513: I think it would be safer to have a case-insensitive comparison here.

- Snowflake is 'case unaware (at least by default), case sensitive'
- Redshift is 'case unaware, case insensitive'
- SQL Server is 'case aware, case insensitive'

We can certainly standardize on our side what casing we want to use while creating - but we can't control what happens on the server side or with a DBA.

For instance, even if the target prefers `lower_case` for SQL Server, a DBA could still manually create a target table with a column name `'User_Id'`. The operation to create a new `'user_id'` column in that case would be incorrect and would not be allowed on server side.

So, I think we want both of these: 

- targets can and should have a preference for how to case tables and columns.
- targets should not assume the preference is honored in the database itself.

What do you think?

kgpayne: @aaronsteers this is still on the Tap side, called by the `SQLStream` class when composing the select statement during `get_records` [here](https://github.com/meltano/sdk/pull/1032/files#diff-61b20de8a12c7c91dfc4de62be5a230d2b4b1b2dd688628705554dc63a9490b6R1060-R1064). AFAIK we don’t do any conforming when reading from databases in Taps; the column names will always exactly match the stream property names because the singer schema is directly derived from the database schema.

I will start the work on column name conforming/transformations in the Target in the morning, where I will definitely run into comparison issues as you describe 😅

edgarrmondragon: https://docs.sqlalchemy.org/en/14/dialects/oracle.html#identifier-casing

> ## Identifier Casing
> In Oracle, the data dictionary represents all case insensitive identifier names using UPPERCASE text. SQLAlchemy on the other hand considers an all-lower case identifier name to be case insensitive. The Oracle dialect converts all case insensitive identifiers to and from those two formats during schema level communication, such as reflection of tables and indexes. Using an UPPERCASE name on the SQLAlchemy side indicates a case sensitive identifier, and SQLAlchemy will quote the name - this will cause mismatches against data dictionary data received from Oracle, so unless identifier names have been truly created as case sensitive (i.e. using quoted names), all lowercase names should be used on the SQLAlchemy side.

I think it's safe to compare the lower case identifiers?

aaronsteers: @edgarrmondragon - This is interesting... I wonder if we should just force everything to lower-case, since that represents probably the safest most reliable behavior across the board.

@kgpayne - Thanks for calling out the distinction between tap considerations (here) and target considerations (future). I still think forcing all comparisons and names to lower case may be safest default behavior across the board.

singer_sdk/streams/sql.py
```diff
@@ -501,26 +503,32 @@ def get_table_columns(self, full_table_name: str) -> dict[str, sqlalchemy.Column
...
                 col_meta["type"],
                 nullable=col_meta.get("nullable", False),
             )
-
-        return result
-
-    def get_table(self, full_table_name: str) -> sqlalchemy.Table:
+            for col_meta in columns
+            if not column_names
+            or col_meta["name"].casefold() in {col.casefold() for col in column_names}
```
edgarrmondragon, line 514: TIL `casefold`

> Casefolding is similar to lowercasing but more aggressive because it is intended to remove all case distinctions in a string. For example, the German lowercase letter 'ß' is equivalent to "ss". Since it is already lowercase, [lower()](https://docs.python.org/3/library/stdtypes.html#str.lower) would do nothing to 'ß'; [casefold()](https://docs.python.org/3/library/stdtypes.html#str.casefold) converts it to "ss".

edgarrmondragon (approved):
Looks good! Nice refactoring too 😃

_Note to myself_: we need more intuitive end-to-end tests that check different inputs to SQL taps like catalog, state, etc.

PR #1036, opened 2022-10-04
fix: create schema and table on `add_sink`

### LINKED ISSUE #1027: bug: Tables not created if no records arrive with `SQLSink`
### Singer SDK Version

0.11.1

### Python Version

3.8

### Bug scope

### Operating System

MacOS

### Description

With non-SDK Targets it is common for schema/table objects to be created on arrival of a `SCHEMA` singer message before and regardless of if any `RECORD` messages ever arrive. This can be important for downstream testing in tools such as `dbt` where `dbt` will error on missing entities rather than proceed to run tests. This may be less of a bug and more of an unexpected behaviour coming from other SQL-based targets.

### Code

_No response_

kgpayne:
Closes #1027 

- #1027 

----
:books: Documentation preview :books:: https://meltano-sdk--1036.org.readthedocs.build/en/1036/


kgpayne:
I see codecov is failing. Wasn't exactly sure where best to add tests, but will look again in the morning 👍

singer_sdk/streams/sql.py
```diff
@@ -3,900 +3,16 @@
...
 
 import sqlalchemy
-from sqlalchemy.engine import Engine
-from sqlalchemy.engine.reflection import Inspector
 
-import singer_sdk.helpers._catalog as catalog
-from singer_sdk import typing as th
-from singer_sdk._singerlib import CatalogEntry, MetadataMapping, Schema
-from singer_sdk.exceptions import ConfigValidationError
+from singer_sdk._singerlib import CatalogEntry, MetadataMapping
```
edgarrmondragon, line 10:
```suggestion
import singer_sdk.helpers._catalog as catalog
from singer_sdk._singerlib import CatalogEntry, MetadataMapping
```

singer_sdk/sql/connector.py
```diff
@@ -0,0 +1,914 @@
+"""Base class for SQLAlchemy-based connectors."""
```
aaronsteers, line 1: @kgpayne - Any changes to call out in the moving of the `SQLConnector` class to the new module?

From the diff, we can't really tell if nothing is changed or if a lot has changed.

aaronsteers:
@kgpayne - Looks like we've all had a first pass at this, and I see you are making a couple iterations, so I'll mark as draft. Feel free to move back to ready status once you are ready for final review. 👍

singer_sdk/sql/connector.py
kgpayne: There were more changes than was simple to call out, so I undid the new module. I still think moving the `SQLConnector` out to a file or module would make it easier to work on, but that can be done in a future PR with no other changes 🙂

kgpayne:
The dependency review failure is unrelated to my changes 🤔

tests/core/test_sqlite.py
```diff
@@ -265,6 +266,41 @@ def test_sync_sqlite_to_sqlite(
...
+            },
+        ]
+    )
+    # sqlite doesn't support schema creation
+    with pytest.raises(sqlalchemy.exc.OperationalError) as excinfo:
+        target_sync_test(
+            sqlite_sample_target, input=StringIO(tap_output), finalize=True
+        )
+    # check the target at least tried to create the schema
+    assert excinfo.value.statement == f"CREATE SCHEMA {schema_name}"
```
edgarrmondragon, line 301: @kgpayne this is great, thanks for adding the test!

singer_sdk/sinks/sql.py
```diff
@@ -65,36 +61,76 @@ def connection(self) -> sqlalchemy.engine.Connection:
...
+        )
+
+    def setup(self) -> None:
+        """Set up Sink.
+
+        This method is called on Sink creation, and creates the required Schema and
+        Table entities in the target database.
+        """
+        if self.schema_name:
+            self.connector.prepare_schema(self.schema_name)
```
edgarrmondragon, line 126: Should `full_schema_name` be used here?

kgpayne: @edgarrmondragon This isn't super clear from the documentation but SQLAlchemy doesn't appear to have native support for fully-qualified names 🤔 e.g. when specifying the schema for a table, you are supposed to do it either on the `MetaData()` metadata object (as I am doing [here](https://github.com/meltano/sdk/pull/1036/files#diff-61b20de8a12c7c91dfc4de62be5a230d2b4b1b2dd688628705554dc63a9490b6R596)) which will apply to all tables used with that metadata, or on the `Table()` object itself (as per the docs example [here](https://docs.sqlalchemy.org/en/14/core/metadata.html#specifying-the-schema-name)):

```python
metadata_obj = MetaData()

financial_info = Table(
    "financial_info",
    metadata_obj,
    Column("id", Integer, primary_key=True),
    Column("value", String(100), nullable=False),
    schema="remote_banks",
)
```

From my testing, I found that passing a fully qualified name (for table or schema) would result in a quoted identifier (as SQLAlchemy correctly detects and escapes the illegal `.` char in the identifier). I.e. in stead of creating an example fully-qualified table of `database.schema.table`, SQLAlchemy would create a table called `"database.schema.table"` in the default/public schema of the current database (from the connection object) 🤦‍♂️

This is different to executing raw SQL directly, where fully-qualified names (when supported by the underlying database) are an option (inserted without quotes). We may prefer to just compose the SQL ourselves and make the composition overridable, but I implemented via `CreateSchema` on the basis that SQLAlchemy will then do the work to compose dialect-specific SQL for us 👍

BuzzCutNorman:
@kgpayne I was wondering how should a developer set a target's default schema once this code is put in place?  Say for argument's sake I created a target-postgres and [set the default schema on connect ](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#setting-alternate-search-paths-on-connect) with some code like this.
```
        @event.listens_for(engine, "connect", insert=True)
        def set_search_path(dbapi_connection, connection_record):
            existing_autocommit = dbapi_connection.autocommit
            dbapi_connection.autocommit = True
            cursor = dbapi_connection.cursor()
            cursor.execute(f"SET SESSION search_path='{self.config.get('schema') or 'public'}'")
            cursor.close()
            dbapi_connection.autocommit = existing_autocommit
```
I believe with this PR in place any blank table would be created in the tap's schema not in the target's default schema from the `meltano.yml`

aaronsteers:
@BuzzCutNorman - Thanks for raising.

@kgpayne - There are probably several discussion points to unwind here, but what do you think of using a setting name of `load_schema` (aligned with internal Meltano naming) or `default_target_schema` (Pipelinewise convention)?

New discussion here where we can dive deep:

- https://github.com/meltano/sdk/discussions/1084#discussion-4484818

If we can remove the reference to `config.get('schema')` / `config["schema"]` and just used a calculated schema when available, and a default (?) otherwise, that safely would push the new scope to a subsequent PR.

What do you think?

kgpayne:
@aaronsteers @BuzzCutNorman 

> If we can remove the reference to config.get('schema') and/or config["schema"] from this PR...

We don't currently have any references to `config.get('schema')` in this PR or on `main` - that call is in BuzzCutNormans' example only. This PR already implements the behaviour you describe:

> apply a calculated schema when available (and just don't set a load schema otherwise?)

 I.e. the `schema_name` property will try to derive a schema name (by splitting the stream name on `-`) and return `None` is the name is not splittable 👍
 
I posted feedback on the [discussion](https://github.com/meltano/sdk/discussions/1084) too, but TLDR; is that we don't currently allow the specification of a schema/search path during the construction of the SQLAlchemy engine, but could easily do so with a `connect_args` setting (here or in another PR). This would then provide a default schema in cases where the `schema_name` property returns `None`, without the need for a custom snippet (like the one BuzzCutNorman suggested) or the overriding of `create_sqlalchemy_engine` on the `SQLConnecter` class 🙂 In future we could additionally support a `load_schema` to 'force' the schema to a user value irrespective of connection or derived schema names.

WDYT?

kgpayne:
> > We don't currently have any references to `config.get('schema')` in this PR or on `main` - that call is in BuzzCutNormans' example only.
> 
> This was my misreading then. Thanks very much for clarifying.

@aaronsteers I had to check, as I thought we used it somewhere too 🙂 Whilst its common in SQL target implementations (as you point out in the discussion) it isn't (yet) in the SDK 👍

singer_sdk/sinks/sql.py
```diff
@@ -65,36 +61,76 @@ def connection(self) -> sqlalchemy.engine.Connection:
...
         Returns:
             The target schema name.
         """
-        return None  # Assumes single-schema target context.
+        parts = self.stream_name.split("-")
+        if len(parts) == 2:
+            return parts[0]
+        if len(parts) == 3:
+            return parts[1]
+        return None
```
aaronsteers, line 84: Our business logic here is admittedly a bit obscure so a rewrite with comments might be easier to maintain/read.

What do you think of something like this?

```suggestion
        parts = self.stream_name.split("-")
        if len(parts) in {2, 3}:
            # Stream name is a two-part or three-part identifier.
            # Use the second-to-last part as the schema name.
            return parts[-2]

	# Schema name not detected.
        return None
```

singer_sdk/sinks/sql.py
edgarrmondragon: > From my testing, I found that passing a fully qualified name (for table or schema) would result in a quoted identifier (as SQLAlchemy correctly detects and escapes the illegal . char in the identifier).

That makes sense 👍. Let's stick with the `Metadata` defining the namespace and SQLAlchemy generating the right schema identifier.

aaronsteers (approved):
@kgpayne - Looks good from my side.

@edgarrmondragon - Any other changes you'd want to see in this iteration?

edgarrmondragon (approved):
@kgpayne Looks good to me!

aaronsteers:
@BuzzCutNorman re:

> @kgpayne I was wondering how should a developer set a target's default schema once this code is put in place?

I think we should follow this PR up with this one:

- #1086 

That would give a standard way for users to control the target load schemas. And presumably those behaviors would be built into the `SQLStream.schema_name` property, such that the `schema_name` returned there would honor the user's default schema preference and/or a preference for schema (re)mappings.

PR #1039, opened 2022-10-05
feat: Support conforming singer property names to target identifier constraints in SQL sinks

### LINKED ISSUE #1021: feat: Support column name conforming
### Feature scope

### Description

As per #1019 many targets have identifier name constraints that are more stringent than what JSON permits. This causes issues when using stream property names as identifiers in base implementations of the utilities provided by the `SQLSink` class for creating/updating schemas in databases.

## Proposed Solution

To create a new method on the `SQLSink` class called `conform_column_name()` (or equivalent) that users can override with their own logic and constraints. This would just return the stream property name by default, keeping the existing functionality effectively unchanged. The existing `database_name`, `schema_name` and `table_name` properties can already be used to conform database/schema/table identifiers.

Note, in cases where a Singer schema is being compared to a Table schema, stream property names should be conformed before comparison.

kgpayne:
Closes #1021

- #1021

----
:books: Documentation preview :books:: https://meltano-sdk--1039.org.readthedocs.build/en/1039/


singer_sdk/sinks/sql.py
```diff
@@ -116,6 +122,42 @@ def full_schema_name(self) -> str:
             schema_name=self.schema_name, db_name=self.database_name
         )
 
+    def conform_name(self, name: str, object_type: Optional[str] = None):
```
edgarrmondragon, line 125:
```suggestion
    def conform_name(self, name: str, object_type: Optional[str] = None) -> str:
```

singer_sdk/sinks/sql.py
```diff
@@ -116,6 +122,42 @@ def full_schema_name(self) -> str:
...
+        """Conform a stream property name to one suitable for the target system.
+
+        Transforms names to snake case by default, applicable to most common DBMSs'.
+        Developers may override this method to apply custom transformations
+        to database/schema/table/column names.
```
edgarrmondragon, line 130:
```suggestion
        to database/schema/table/column names.
        
        Args:
            name: Property name.
            object_type: One of ``database``, ``schema``, ``table`` or ``column``.

```

singer_sdk/sinks/sql.py
```diff
@@ -116,6 +122,42 @@ def full_schema_name(self) -> str:
...
+        name = re.sub(r"[^a-zA-Z0-9_\-\.\s]", "", name)
+        # convert to snakecase
+        name = snakecase(name)
+        # remove leading digits
+        return name.lstrip(digits)
+
+    def conform_schema(self, schema: dict) -> dict:
+        """Return schema dictionary with property names conformed.
+
+        Returns:
```
edgarrmondragon, line 145:
```suggestion
        """Return schema dictionary with property names conformed.

        Args:
			schema: JSON schema dictionary.

        Returns:
```

singer_sdk/sinks/sql.py
```diff
@@ -116,6 +122,42 @@ def full_schema_name(self) -> str:
...
+        """
+        conformed_schema = copy(schema)
+        conformed_schema["properties"] = {
+            self.conform_name(key): value
+            for key, value in conformed_schema["properties"].items()
+        }
+        return conformed_schema
+
+    def conform_record(self, record: dict) -> dict:
+        """Return record dictionary with property names conformed."""
```
edgarrmondragon, line 156:
```suggestion
        """Return record dictionary with property names conformed."""

		Args:
			record: Dictionary representing a single record.
			
		Returns:
			New record dictionary with conformed column names.
```

edgarrmondragon (changes requested):

singer_sdk/sinks/sql.py
```diff
@@ -116,6 +122,42 @@ def full_schema_name(self) -> str:
...
+
+        Returns:
+            The name transformed to snake case.
+        """
+        # strip non-alphanumeric characters, keeping -, _ and spaces
+        name = re.sub(r"[^a-zA-Z0-9_\-\.\s]", "", name)
+        # convert to snakecase
+        name = snakecase(name)
+        # remove leading digits
+        return name.lstrip(digits)
```
edgarrmondragon, line 140: @kgpayne Perhaps adding an alpha prefix instead of a replacing leading digits is preferable?

Suppose there's properties `1_users`, `2_users`, then they'd both be conformed to `_users`.

Wdyt?

samples/sample_tap_hostile/hostile_streams.py
```diff
@@ -0,0 +1,42 @@
...
+    in common DBMS systems.
+    """
+
+    name = "hostile_property_names_stream"
+    schema = th.PropertiesList(
+        th.Property("name with spaces", th.StringType),
+        th.Property("NameIsCamelCase", th.StringType),
+        th.Property("name-with-dashes", th.StringType),
+        th.Property("Name-with-Dashes-and-Mixed-cases", th.StringType),
+        th.Property("5name_starts_with_number", th.StringType),
```
edgarrmondragon, line 23: See my comment above regarding adding a prefix instead of removing leading digits

```suggestion
        th.Property("5name_starts_with_number", th.StringType),
        th.Property("6name_starts_with_number", th.StringType),
        th.Property("7name_starts_with_number", th.StringType),
```

edgarrmondragon (changes requested):

singer_sdk/sinks/sql.py
kgpayne: @edgarrmondragon I thought about this a bit over the weekend, and think a good general approach would be to implement a post-conform check and custom error on "conformed name clash". This will be safe for us as we add to the corpus of cases handled in the default implementation (happy to add an alpha prefix to the default transformations too) but also safe for users as they override the conform recipes for their own targets 🙂

edgarrmondragon: @kgpayne 

> implement a post-conform check and custom error on "conformed name clash".

That sounds like a good approach 👍. Do you think name conforming errors are always bugs, or is there potential for "user errors" that would cause them? I ask because if it's the former, then we should think of a way to direct users to report the bug directly in the connector repo.

(That maybe applies too for schema errors in REST taps, but it just occurred to me)

> happy to add an alpha prefix to the default transformations too

Yeah, if you think it's worth it go for it 😄

singer_sdk/sinks/sql.py
```diff
@@ -144,6 +145,37 @@ def conform_name(self, name: str, object_type: Optional[str] = None) -> str:
...
+            duplicated_property_string = ", ".join(
+                [
+                    f"{dup} -> {conformed_property_names[dup]}"
+                    for dup in duplicated_properties
+                ]
+            )
+            raise ConformedNameClashException(
+                "Duplicate stream properties produced when conforming property names: "
+                + duplicated_property_string
+            )
```
edgarrmondragon, line 176: There's quite a few loops here. Over

1. dict.values()
2. list(dict.values())
3. list.count()
4. conformed_property_names.items()
5. duplicated_properties

It might be worth simplifying to 2 (or 3 if pretty formatting the dups is a requirement):

```suggestion
        """Check if conformed names produce duplicate keys.

        Raises:
            ConformedNameClashException if duplicates found.
        """
        # group: {'_a': ['1_a'], 'abc': ['aBc', 'abC']}
        grouped = defaultdict(list)
        for k, v in mapping.items():
            grouped[v].append(k)

        # filter
        duplicates = list(filter(lambda p: len(p[1]) > 1, grouped.items()))

        if duplicates:
            raise ConformedNameClashException(
                f"Duplicate stream properties produced when conforming property names: {duplicates}"
            )
```

edgarrmondragon (changes requested):

singer_sdk/sinks/sql.py
aaronsteers: > Do you think name conforming errors are always bugs, or is there potential for "user errors" that would cause them?

Just to chime in here, I think name collisions are probably an area we'll need to come back to in future iterations. I don't remember the exact column names, but GitHub, for example has a '+1' and '-1' column for the emotes. And an early implementation of tap-github were basically rendering them with the same name, since both '-' and '+' were rendered as special characters.

I'd be okay with the 'hard failure' behavior when collisions are detected or 'warn and suppress the latter field' behavior. We probably don't want to _only_ warn, since if we potentially send both properties to the same target node, that could lead to data corruption.

If a hard failure occurs, especially if we give a clear error message, the user will go through:

> Oh, '+1' and '-1' both evaluate to 'c_1'? I don't really need those columns so I'll deselect them upstream in the pipeline.

Or

> Oh, '+1' and '-1' both evaluate to 'c_1'? Let me alias them explicitly in the tap...

So - I slightly prefer a hard failure because it gives the user more control over resolution. (Warnings are often missed in the massive volumes of logs.)

singer_sdk/sinks/sql.py
kgpayne: Way better, thanks 🙏

singer_sdk/sinks/sql.py
kgpayne: Agree that conforming names will inevitably cause clashes. As of now, the default implementation tries to mitigate some obvious clashes, and raises a `ConformedNameClashException` if any clashes are detected. Users are free to override `conform_name()` for their own targets, so I think hard-fail is the safest approach 🙂

edgarrmondragon (approved):

aaronsteers (approved):

PR #1072, opened 2022-10-14
chore: Update logo

tayloramurphy:
----
:books: Documentation preview :books:: https://meltano-sdk--1072.org.readthedocs.build/en/1072/


WillDaSilva (approved):
Looks good to me. This only affects the site on large screens. If you open the site on a small screen (e.g. a phone) then it uses plain text "Meltano SDK", which looks fine, but maybe we want to use the logo there too. That can be done in a future PR.

PR #1087, opened 2022-10-19
feat: S3 batch storage

### LINKED ISSUE #1070: [Feature]: Add s3:// as a supported storage root for batch files.
### Feature scope

### Description

Extend the [Batch](https://github.com/meltano/sdk/pull/904) feature to support AWS s3 as a supported storage root.

Other changes that would be good to ship with this feature:
* Add `cleanup_files=true|false` option for Batch config on the target side. `False` is currently the default behavior for Batch messages. Making it configurable would allow users flexibility on whether they want to keep the batch files after processing them.

jamielxcarter:
Closes #1070

----
:books: Documentation preview :books:: https://meltano-sdk--1087.org.readthedocs.build/en/1087/


pyproject.toml
```diff
@@ -75,6 +78,7 @@ docs = [
     "myst-parser",
     "sphinx-autobuild",
 ]
+s3fs = ["fs-s3fs"]
```
aaronsteers, line 81: As an 'extras' name, I might slightly prefer the shorter and more intuitive `s3`. What do you think?

jamielxcarter: I went ahead and updated the 'extras' name to `s3`. I think it does make more sense that way!

aaronsteers:
@jamielxcarter and @edgarrmondragon - This is awesome - I'm happy to see the implementation was so straightforward.

What is unclear to me as of yet is what instructions to provide to (1) users and (2) developers.

For developers, if they want to opt in to including the `s3` extra, they probably would do so in `poetry.toml` - should we have a commented-out section in the cookiecutter for guidance?

And for users, especially if the developer has not already bundled the `s3` extra, what's the recommendation to them for bundling the extra library during their install? Assuming a tool like `pipx` or `meltano` is auto-creating a virtual environment, it's not clear to me how to add the `s3` library into the virtualenv appropriately. Perhaps?:

1. For meltano users, should they append `singer-sdk[s3]` or `fs-s3fs` to the `pip_url` in `meltano.yml`.
2. For pipx users, should they use `pipx inject`, as in `pipx inject tap-mytap fs-s3fs` or similar?

noxfile.py
```diff
@@ -69,6 +69,7 @@ def tests(session: Session) -> None:
         "flake8-annotations",
         "flake8-docstrings",
         "mypy",
+        "fs-s3fs",
```
edgarrmondragon, line 72: You should be able to install the extras like this:

```python
    session.install(".[s3]")
    session.install(
        "coverage[toml]",
        "pytest",
        "freezegun",
        "pandas",
        "pyarrow",
        "requests-mock",
        # Cookiecutter tests
        "black",
        "cookiecutter",
        "PyYAML",
        "darglint",
        "flake8",
        "flake8-annotations",
        "flake8-docstrings",
        "mypy",
    )
```

edgarrmondragon (changes requested):

edgarrmondragon:
@aaronsteers @jamielxcarter I think there's a couple of options to trickle down support for S3, and the distinction between

- _Developers_: tap creators and maintainers
- _Users_: users of the tap (via Meltano or otherwise)

is useful.

## Option 1: All Users get the extra dependency

```
[tool.poetry.dependencies]
python = "<3.11,>=3.7.1"
singer-sdk = { version="0.11.1", extras = ["s3"] }
```

## Option 2: Extras all the way down

Make the extra dependency a passthrough, so users would have to explicitly request the extra:

```
pipx install 'tap-github[s3]'
```

Doesn't work at the moment, unfortunately:

- https://github.com/orgs/python-poetry/discussions/4313#discussioncomment-1757450
- https://github.com/python-poetry/poetry/issues/834

aaronsteers:
> @aaronsteers @jamielxcarter I think there's a couple of options to trickle down support for S3, and the distinction between
> 
> * _Developers_: tap creators and maintainers
> * _Users_: users of the tap (via Meltano or otherwise)
> 
> is useful.
> 
> ## Option 1: All Users get the extra dependency
> ```
> [tool.poetry.dependencies]
> python = "<3.11,>=3.7.1"
> singer-sdk = { version="0.11.1", extras = ["s3"] }
> ```
> 
> ## Option 2: Extras all the way down
> Make the extra dependency a passthrough, so users would have to explicitly request the extra:
> 
> ```
> pipx install 'tap-github[s3]'
> ```
> 
> Doesn't work at the moment, unfortunately:
> 
> * https://github.com/orgs/python-poetry/discussions/4313#discussioncomment-1757450
> * [How can I specify differing sets of dependency extras for each of my project's extras? python-poetry/poetry#834](https://github.com/python-poetry/poetry/issues/834)

Well said.

I think for taps and targets using AWS-inclined system like redshift and snowflake, the developer can choose option 1 and just ship it all the time for all users.

For the user opt-in scenario (meaning the tap doesn't bias towards aws/s3, but still the user might want/need to use S3 anyway), is it just the automatic pass-through that doesn't work? Or, said another way, could we explicitly re-declare the `s3` in the boilerplate cookiecutter templates? This would presumably give all users the option, but of course the `extra` declaration could still be removed by a developer - or not included in the first place.

Would that work? And if developers remove it (or don't add it), the user would need to find another means to inject?

Jamiel-Carter:
> > @aaronsteers @jamielxcarter I think there's a couple of options to trickle down support for S3, and the distinction between
> > 
> > * _Developers_: tap creators and maintainers
> > * _Users_: users of the tap (via Meltano or otherwise)
> > 
> > is useful.
> > ## Option 1: All Users get the extra dependency
> > ```
> > [tool.poetry.dependencies]
> > python = "<3.11,>=3.7.1"
> > singer-sdk = { version="0.11.1", extras = ["s3"] }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ## Option 2: Extras all the way down
> > Make the extra dependency a passthrough, so users would have to explicitly request the extra:
> > ```
> > pipx install 'tap-github[s3]'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Doesn't work at the moment, unfortunately:
> > 
> > * https://github.com/orgs/python-poetry/discussions/4313#discussioncomment-1757450
> > * [How can I specify differing sets of dependency extras for each of my project's extras? python-poetry/poetry#834](https://github.com/python-poetry/poetry/issues/834)
> 
> Well said.
> 
> I think for taps and targets using AWS-inclined system like redshift and snowflake, the developer can choose option 1 and just ship it all the time for all users.
> 
> For the user opt-in scenario (meaning the tap doesn't bias towards aws/s3, but still the user might want/need to use S3 anyway), is it just the automatic pass-through that doesn't work? Or, said another way, could we explicitly re-declare the `s3` in the boilerplate cookiecutter templates? This would presumably give all users the option, but of course the `extra` declaration could still be removed by a developer - or not included in the first place.
> 
> Would that work? And if developers remove it (or don't add it), the user would need to find another means to inject?

@aaronsteers @edgarrmondragon I think I've implemented what was discussed. Let me know if any further changes are needed!

jamielxcarter:
> > @aaronsteers @jamielxcarter I think there's a couple of options to trickle down support for S3, and the distinction between
> > 
> > * _Developers_: tap creators and maintainers
> > * _Users_: users of the tap (via Meltano or otherwise)
> > 
> > is useful.
> > ## Option 1: All Users get the extra dependency
> > ```
> > [tool.poetry.dependencies]
> > python = "<3.11,>=3.7.1"
> > singer-sdk = { version="0.11.1", extras = ["s3"] }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ## Option 2: Extras all the way down
> > Make the extra dependency a passthrough, so users would have to explicitly request the extra:
> > ```
> > pipx install 'tap-github[s3]'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Doesn't work at the moment, unfortunately:
> > 
> > * https://github.com/orgs/python-poetry/discussions/4313#discussioncomment-1757450
> > * [How can I specify differing sets of dependency extras for each of my project's extras? python-poetry/poetry#834](https://github.com/python-poetry/poetry/issues/834)
> 
> Well said.
> 
> I think for taps and targets using AWS-inclined system like redshift and snowflake, the developer can choose option 1 and just ship it all the time for all users.
> 
> For the user opt-in scenario (meaning the tap doesn't bias towards aws/s3, but still the user might want/need to use S3 anyway), is it just the automatic pass-through that doesn't work? Or, said another way, could we explicitly re-declare the `s3` in the boilerplate cookiecutter templates? This would presumably give all users the option, but of course the `extra` declaration could still be removed by a developer - or not included in the first place.
> 
> Would that work? And if developers remove it (or don't add it), the user would need to find another means to inject?

@aaronsteers @edgarrmondragon I think I've implemented the changes discussed by you both. Let me know if anything is missing!

cookiecutter/tap-template/{{cookiecutter.tap_id}}/pyproject.toml
```diff
@@ -12,7 +12,7 @@ license = "Apache 2.0"
 [tool.poetry.dependencies]
 python = "<3.11,>=3.7.1"
 requests = "^2.25.1"
-singer-sdk = "^0.13.0"
+singer-sdk = { version="^0.13.0", extras = ["s3"] }
```
aaronsteers, line 15: Wdyt of updating this also on the target SDK template to allow batches files in S3 to be ready by targets?

aaronsteers (commented):
Thanks for applying the recent changes.

One question/suggestion below.

cc @edgarrmondragon to rereview.

cookiecutter/tap-template/{{cookiecutter.tap_id}}/pyproject.toml
Jamiel-Carter: Ah, thanks for catching that!

jamielxcarter: Done

cookiecutter/tap-template/{{cookiecutter.tap_id}}/pyproject.toml
```diff
@@ -12,7 +12,7 @@ license = "Apache 2.0"
 [tool.poetry.dependencies]
 python = "<3.11,>=3.7.1"
 requests = "^2.25.1"
-singer-sdk = "^0.14.0"
+singer-sdk = { version="^0.14.0", extras = ["s3"] }
```
edgarrmondragon, line 15: This will install the extras in all derived packages. Passthrough extras are unfortunately not supported (see https://github.com/orgs/python-poetry/discussions/4313).

Can we make an explicit extra for the S3 dependencies?

For example:

```toml
[tool.poetry.dependencies]
python = "<3.11,>=3.7.1"
requests = "^2.25.1"
singer-sdk = { version="^0.14.0"}
fs-s3fs = { version = "^1.1.1", optional = true}

[tool.poetry.extras]
s3 = ["fs-s3fs"]
```

edgarrmondragon (changes requested):

cookiecutter/tap-template/{{cookiecutter.tap_id}}/pyproject.toml
jamielxcarter: Thanks Edgar. I've updated the pyproject.toml tap/target templates to include an explicit extra for s3 dependencies.

edgarrmondragon (approved):

Issue #317, opened 2022-01-29
Create HTTPConnector interface for all API interactions


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/319

Originally created by @edgarrmondragon on 2022-01-29 01:38:55

---

## Summary

[//]: # (Concisely summarize the feature you are proposing.)

Currently, for HTTP taps (REST, GraphQL, etc.), all aspects of making a request go through the stream class. This works well for most access patterns where one endpoint corresponds to one stream. However, there are other use cases that require making requests not with the direct purpose of yielding records.

## Proposed benefits

[//]: # (Concisely summarize the benefits this feature would bring to yourself and other users.)

- Another request is required to complement or enrich the data, like when trying to get a complete record after a response that contains only IDs (https://gitlab.com/meltano/sdk/-/issues/93+).
  
- The API is dynamic and offers metadata endpoints that may inform the tap about available streams (https://gitlab.com/meltano/sdk/-/issues/313+) and/or their schemas (https://gitlab.com/meltano/sdk/-/issues/310+).
  
- The underlying request is not accessible, making it harder or unclear how to tweak parameters like timeout limits, caching (https://gitlab.com/meltano/sdk/-/issues/237+), etc.
  
- Async requests are not currently supported. Separation of concerns in an object dedicated to HTTP requests might make it easier for the community to implement and contribute an async connector (https://gitlab.com/meltano/sdk/-/issues/291+ and https://gitlab.com/meltano/sdk/-/issues/184+).
  
- A single interface for all HTTP sources would make features and bugfixes automatically apply to existing stream types and new ones too (https://gitlab.com/meltano/sdk/-/issues/221+).
  

## Proposal details

[//]: # (In as much detail as you are able, describe the feature you'd like to build or would like to see built.)

- Build a module dedicated to HTTP sources, i.e. `singer_sdk.http`.
  
  - `singer_sdk.http.connector` contains the `HTTPConnector` class. In the MVC, `HTTPConnector` would manage aspects that the exiting stream classes do not, like the `Session` object.
    
  - `singer_sdk.http.streams.base`. In the MVC, methods useful to both GraphQL and REST streams can be moved to `HTTPStream` in this module without introducing breaking changes.
    
  - `singer_sdk.http.streams.rest` moved from `singer_sdk.streams.rest`
    
  - `singer_sdk.http.streams.graphql` moved from `singer_sdk.streams.grapql`
    
[...]

PR #1091, opened 2022-10-20
refactor: Move SQLConnector into a separate module, for use by both SQLStream and SQLSink

### LINKED ISSUE #1090: refactor: Move `SQLConnector` into a separate module, for use by both `SQLStream` and `SQLSink`
Currently the `SQLConnector` class is in `stream` module in the same file as `SQLStream`, however it is also needed by `SQLSink` from the `sink` module. Moving the connector into a `sql` module would make it easier to work on sinks, and also provide a place for other shared, SQL-related code.

kgpayne:
Closes #1090 

- #1090 

----
:books: Documentation preview :books:: https://meltano-sdk--1091.org.readthedocs.build/en/1091/


kgpayne:
This commit doesn't change any logic. I simply moved the `SQLConnector` into a new `sql` module 🙂

singer_sdk/streams/__init__.py
```diff
@@ -3,12 +3,6 @@
...
+from singer_sdk.streams.sql import SQLStream
 
-__all__ = [
-    "Stream",
-    "GraphQLStream",
-    "RESTStream",
-    "SQLStream",
-    "SQLConnector",
-]
+__all__ = ["Stream", "GraphQLStream", "RESTStream", "SQLStream"]
```
aaronsteers, line 8: @kgpayne

I think we could go either way on this but are we sure we want to remove the reference of `singer_sdk.streams.SQLConnector`?

I'm totally okay with the decision to do so, but probably we should also remove `singer_sdk.sinks.SQLConnector` here if so: https://github.com/meltano/sdk/blob/kgpayne/issue1090/singer_sdk/sinks/__init__.py

I'd also be okay leaving it as referenceable from both `singer_sdk.streams` and `singer_sdk.sinks`. Small benefit of not breaking existing code with the slight downside of possible confusion of where to best import from.

cc @edgarrmondragon

singer_sdk/sql/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Module for SQL-related operations."""
+
+from .connector import SQLConnector
```
aaronsteers, line 3: @kgpayne - It feels just a little odd to me that `SQLStream` and `SQLSink` are not in the `sql` module. 

What do you think of reversing the module ref name, from `singer_sdk/sql/connector.py` to `singer_sdk/connectors/sql.py` - which then would make `SQLConnector` importable from `singer_sdk.connectors` and also `singer_sdk.connectors.sql`. Not a ton of other uses for this namespace as of now, but this leaves open the possibility us us creating something like `singer_sdk.connectors.OpenAPIConnector` and similar.

Alternatively, we could put this under `singer_sdk/helpers/sql.py`.

What do you think?

cc @edgarrmondragon, @cjohnhanson

singer_sdk/streams/__init__.py
edgarrmondragon: @kgpayne @aaronsteers

TLDR I'm okay with removing `SQLConnector` from the "public API" of the module.

---

`__all__` only affects _star_ imports.

For example, consider the module

```python
# mymod.py
__all__ = ["X", "Y"]

X = 1
Y = 2
Z = 3
```

Then the module's "public API" is only reflected when importing with `*`:

```python
>>> from mymod import *
>>> X, Y, Z
  File "<stdin>", line 1, in <module>
NameError: name 'Z' is not defined
```

But the `Z` member can still be imported explicitly:

```python
>>> from mymod import X, Y, Z
>>> X, Y, Z
(1, 2, 3)
```

singer_sdk/sql/__init__.py
edgarrmondragon: I really like `singer_sdk/connectors/sql.py`. It's similar to what I imagined for https://github.com/meltano/sdk/issues/317 😄

singer_sdk/streams/__init__.py
aaronsteers: > `__all__` only affects _star_ imports.

Agreed, and understood. I believe it may also affect IDE auto-complete for those editors which will use it to hint at members within those modules, yes?

This SO article seems to confirm: 

https://stackoverflow.com/a/35710527/4298208

> Documentation tools
>
> Documentation and code autocompletion tools may (in fact, should) also inspect the `__all__` to determine what names to show as available from a module.

I do think it's worth being careful about out `__all__` declarations, for the purpose of those IDE hints.

Also - to clarify, I believe the 'breaking' part would be the prior references at the `__init__.py` level, specifically `singer_sdk.[streams|sinks].SQLConnector` - but this would still work `singer_sdk.[streams|sinks].sql.SQLConnector` because both `.../sql.py` files import `SQLConnector` at the top of the file.

kgpayne: Done ✅

edgarrmondragon (approved):

Issue #1092, opened 2022-10-20
tap-template has no environments configured


WillDaSilva:
https://github.com/meltano/sdk/blob/6d11a415c551b3caadd2ee59225d481e6a64d9b9/cookiecutter/tap-template/%7B%7Bcookiecutter.tap_id%7D%7D/meltano.yml

This template is used by some tutorials / blog posts which go on to call `meltano run`, which requires an environment. We should update the template to set `default_environment` and configure the default env/staging/prod environments.

PR #1093, opened 2022-10-20
fix(templates): add environment support to the cookie cutter for `meltano.yml`

pnadolny13:
Closes https://github.com/meltano/sdk/issues/1092

Related to slack thread https://meltano.slack.com/archives/C01TCRBBJD7/p1666232552939149

----
:books: Documentation preview :books:: https://meltano-sdk--1093.org.readthedocs.build/en/1093/

### DIFF
--- a/cookiecutter/tap-template/{{cookiecutter.tap_id}}/meltano.yml
+++ b/cookiecutter/tap-template/{{cookiecutter.tap_id}}/meltano.yml
@@ -1,10 +1,9 @@
 version: 1
 send_anonymous_usage_stats: true
 project_id: "{{cookiecutter.tap_id}}"
+default_environment: test
 environments:
-- name: dev
-- name: staging
-- name: prod
+- name: test
 plugins:
   extractors:
   - name: "{{cookiecutter.tap_id}}"
--- a/cookiecutter/tap-template/{{cookiecutter.tap_id}}/meltano.yml
+++ b/cookiecutter/tap-template/{{cookiecutter.tap_id}}/meltano.yml
@@ -1,6 +1,10 @@
 version: 1
 send_anonymous_usage_stats: true
 project_id: "{{cookiecutter.tap_id}}"
+environments:
+- name: dev
+- name: staging
+- name: prod
 plugins:
   extractors:
   - name: "{{cookiecutter.tap_id}}"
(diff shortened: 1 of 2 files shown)


cookiecutter/tap-template/{{cookiecutter.tap_id}}/meltano.yml
```diff
@@ -1,6 +1,10 @@
 version: 1
 send_anonymous_usage_stats: true
 project_id: "{{cookiecutter.tap_id}}"
+environments:
```
WillDaSilva, line 4:
```suggestion
default_environment: dev
environments:
```

WillDaSilva (changes requested):

cookiecutter/tap-template/{{cookiecutter.tap_id}}/meltano.yml
aaronsteers: @pnadolny13, @WillDaSilva - Was thinking on this more... since this is a `meltano.yml` file going into a tap or target's repo, should we just name a single environment called `test` or `devtest` (or maybe just `dev`).

I don't know if `staging` and `prod` would apply here. We could add something like `cicd` but that's probably more confusing/distracting than helpful.

```yml
default_environment: test
environments:
- name: test
```

aaronsteers: And whatever we decide, should we add the same for the target cookiecutter?

pnadolny13:
@WillDaSilva @aaronsteers I updated based on your feedback and added to the target cookiecutter too!

WillDaSilva (approved):

cookiecutter/target-template/{{cookiecutter.target_id}}/meltano.yml
```diff
@@ -1,16 +1,19 @@
...
   - name: "{{cookiecutter.target_id}}"
     namespace: "{{cookiecutter.library_name}}"
     pip_url: -e .
     capabilities:
-    - state
-    - catalog
-    - discover
+    - about
+    - stream-maps
+    - record-flattening
```
edgarrmondragon, line 16: @pnadolny13 It makes sense to add these to the tap template too

edgarrmondragon (changes requested):

edgarrmondragon (approved):

aaronsteers:
> @pnadolny13 This looks good. Just a last ping to @aaronsteers to check the PR title.

Looks good! Thanks!

PR #1094, opened 2022-10-20
feat: enable Taps and Targets to generate their own Meltano yaml files

### LINKED ISSUE #77: Handling secrets in config settings
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/77

Originally created by @aaronsteers on 2021-03-31 20:28:20

---

The following discussion from [!39](https://gitlab.com/meltano/sdk/-/merge_requests/39) should be addressed:

- [ ] @DouweM started a [discussion](https://gitlab.com/meltano/singer-sdk/-/merge_requests/39#note_541523129): 

    > If we're not doing anything with this yet, can we leave it out for the moment and move it to a WIP merge request?
    > 
    > Once we officially introduce the concept of secrets, we'll also want to let config settings explicitly be marked as such.

aaronsteers:
Update:

I've restructured this to build on #1096. That one would therefor need to merge before this one.

Timing:

- I'd like to get this delivered before Tap-Toberfest if possible.

Todo (this PR):

- [x] Typing helpers for flagging secrets with `writeOnly=True` and `secret=True`.
  - Resolves #77 
- [x] Unit tests.
- [x] Manual testing.

Todo (future iterations):

- [ ] Also detect and print setting nodes for sub-properties. (Currently, only top-level settings are documented.)
- [ ] (Maybe) Allow developers to give hints to maintenance/maintainer-related metadata like 'repo', 'variant' and 'maintenance_status'.
  - As of now, I think it's fair to say these are too self-referential. Many are just generally problematic to add to the repo directly _in code_ because all forks would inherit the same values by default. And a status like 'active' would also just not make sense as a permanent artifact of the repo.


aaronsteers:
@edgarrmondragon and @pnadolny13 - This is an early draft but feel free to add comments/suggestions. Thanks!

singer_sdk/helpers/_meltano.py
```diff
@@ -0,0 +1,110 @@
...
+    if is_datetime_type(jsonschema_type):
+        return "date_iso8601"
+
+    if is_integer_type(jsonschema_type):
+        return "integer"
+
+    return None
+
+
+def meltano_yaml_str(
```
edgarrmondragon, line 52: @aaronsteers So, I guess this is don't as pure string manipulation so comments can be added.

Wouldn't it make more sense for Meltano to ingest the output of `--about --format=json` to generate the config it expects? I say it because the SDK shouldn't need to know about what fields Meltano expects, but Meltano could have a utility CLI Meltano (bundled in the same Python package perhaps) to parse the SDK plugin metadata:

```
tap-github --about --format=json | meltano-parse-sdk-info
```

Another benefit of that approach is that developers (or us) don't need to upgrade their packages to be able to output the right Meltano config, rather users of Meltano bump to the latest `meltano` package to get this utility.

Developers could still bump the SDK version in their packages if we update the metadata exposed through the `--about --format=json` output.

Wydt?

aaronsteers: > @aaronsteers So, I guess this is done as pure string manipulation so comments can be added.
> 
> Wouldn't it make more sense for Meltano to ingest the output of `--about --format=json` to generate the config it expects? I say it because the SDK shouldn't need to know about what fields Meltano expects, but Meltano could have a utility CLI (bundled in the same Python `meltano` package perhaps) to parse the SDK plugin metadata:
> 
> ```
> tap-github --about --format=json | meltano-parse-sdk-info
> ```
> 
> Another benefit of that approach is that developers (or us) don't need to upgrade their packages to be able to output the right Meltano config, rather users of Meltano bump to the latest `meltano` package to get this utility.
> 
> Developers could still bump the SDK version in their packages if we update the metadata exposed through the `--about --format=json` output.
> 
> Wydt?

My intuition was similar at first, and certainly these are good points. Unfortunately, there just isn't a natural place for this to live in Meltano as of yet, and I think it will harder if we try to iterate both tools in lock step. I think we have an opportunity in the SDK to introduce a first round of interop, followed by quick patches if/when we find any issues. When stable and proven, we can migrate this functionality into Meltano - but I don't know that we yet have a good enough feature to make it worth adding to Meltano and then phasing out when we inevitably have a more robust paradigm. There just isn't anywhere I can imagine us putting this into Meltano as of today that wouldn't need to be removed/replaced in a future Meltano revision. But meanwhile, if we iterate in the SDK, we should get to a place where the issues are ironed out, and then we'll be closer to something like a `meltano publish extractor` or `meltano create extractor` type of a workflow.

All that said, I really don't think there is any issue with baking MeltanoHub interop directly into the SDK, since it directly helps developers promote their taps, and likewise streamlines adoption/onboarding for users even if the tap is not on the Hub. It's not intrusive and doesn't overly promote or favor Meltano over other orchestrators.

It's possible we'll phase out `--format=meltano` before we get to 1.0, but I think it's a valuable and (relatively) stable increment from where we are today.

aaronsteers: @edgarrmondragon - Since our conversation, I've refactored this to an internal/dev-side CLI.

```console
% sample-tap-countries --about --format=json > tap-countries.about.json
% singer-tools analyze --from-file=tap-countries.about.json --out-dir=.output
Meltano plugin definition: .output/sample-tap-countries.meltano.yml
```

This doesn't change the experience at all for tap users, but it makes a new command available to those who want to leverage it. As noted in `poetry.toml`, this can be executed from with the SDK repo during development, or by installing `singer-sdk` as a standalone program with `pipx`.

I don't think we should document or promote at this time, but developers can also execute this with `python -m singer_sdk.dev.cli --help` or `python -m singer_sdk.dev.cli analyze ...` anywhere that the `singer_sdk` library is installed.

aaronsteers:
@edgarrmondragon - Since our conversation, I've refactored this to an internal/dev-side CLI.

% sample-tap-countries --about --format=json > tap-countries.about.json
% singer-tools analyze --from-file=tap-countries.about.json --out-dir=.output
Meltano plugin definition: .output/sample-tap-countries.meltano.yml
This doesn't change the experience at all for tap users, but it makes a new command available to those who want to leverage it. As noted in poetry.toml, this can be executed from with the SDK repo during development, or by installing singer-sdk as a standalone program with pipx.

I don't think we should necessarily promote at this time, but developers can also execute this with python -m singer_sdk.dev.cli --help or python -m singer_sdk.dev.cli analyze ... anywhere that the singer_sdk library is installed.

@WillDaSilva - In the long run, I wonder if we can+should put the hub analyze work here in the SDK singer-tools dev CLI. One of the reasons the analyze command stores its output to a directory of files is that I am imagining we could move that complex /plugin-test logic here, and then let the Hub call this command - analyzing the tap or target and storing a bunch of file outputs based on the discovery and analysis.

singer_sdk/helpers/_meltano.py
pnadolny13: @aaronsteers my initial reactions were the same as @edgarrmondragon's. We're tightly coupling the hub and the SDK/Meltano CLI by baking it in.

What if we use the EDK to make a utility that converts the `--format=json` output to a hub yaml file. It's not necessarily where I was going with it originally but I created a [utility](https://github.com/pnadolny13/hub-utils) that helps create taps/targets hub definitions and puts them in the right location in the hub repo. If we had a meltano extension that contained this logic then we're not tied to the SDK version or meltano version either, the user could run `meltano add utility hub-utils` then `meltano invoke hub-utils add` and it kicks off prompts to get the `--about` scrapped, prompt for any other info like a logo png, and write the yaml file to the appropriate directory in the hub repo. If we make any changes to the format or whatever, the user just needs to re-install the updated utility to get support for those changes.

PR #1096, opened 2022-10-20
feat: add secrets support for tap and target config, via `Property(..., secret=True)`

aaronsteers:
Changes:

- [x] Implement `writeOnly=True` support for passwords and other secrets, as per the JSON Schema spec.
- [x] Additionally implement a `secret=True` annotation to help with readability for humans.
- [x] The `Property` type helper class now now implements both of the above via a `secret=True` argument in the constructor.
- [x] The same type helpers may be used within `schema` declarations for streams.

Todo:

- [x] Docs for tap config
- ~~[ ] Docs for stream schema creation~~

Note:

- This also includes general unit testing and type helper improvements, as it was initially developed as part of #1094.


singer_sdk/typing.py
```diff
@@ -352,21 +352,30 @@ def __init__(
...
         description: str = None,
+        secret: bool = False,
     ) -> None:
         """Initialize Property object.
 
+        Note: Properties containing secrets should be specified with `secret=True`.
+        Doing so will add the annotation `writeOnly=True`, in accordance with JSON
+        Schema Draft 7 and later, and `secret=True` as an additional hint to readers.
+
+        More info: https://json-schema.org/draft-07/json-schema-release-notes.html
```
edgarrmondragon, line 363: @aaronsteers Should we update the validator in `singer_sdk/plugin_base.py` to use the `Draft7Validator` object?

singer_sdk/typing.py
```diff
@@ -402,6 +411,13 @@ def to_dict(self) -> dict:
             type_dict.update({"default": self.default})
         if self.description:
             type_dict.update({"description": self.description})
+        if self.secret:
+            type_dict.update(
+                {
+                    "secret": True,
```
edgarrmondragon, line 417: `"sensitive"` is used elsewhere. If that was the intention, maybe it's a good idea to use a `JSON_SCHEMA_SENSITIVE_KEY = "sensitive"` constant that's imported where it's needed?

singer_sdk/plugin_base.py
```diff
@@ -28,6 +28,7 @@
 from singer_sdk.exceptions import ConfigValidationError
 from singer_sdk.helpers._classproperty import classproperty
 from singer_sdk.helpers._compat import metadata
+from singer_sdk.helpers._meltano import meltano_yaml_str
```
edgarrmondragon, line 31: Is this a leftover from refactoring the other PR?

edgarrmondragon (changes requested):

singer_sdk/plugin_base.py
aaronsteers: Good catch - yes! Will remove...

aaronsteers: This whole file, actually.

singer_sdk/typing.py
aaronsteers: Good catch. I switched during development from 'sensitive' to 'secret' on the basis that _lots_ of data is "sensitive" that is not also a secret in the technical sense of secret-management. (Revenue, PII, and lots of other fields, for instance can be sensitive but not 'writeOnly' or 'secret' in the annotation type.)

singer_sdk/typing.py
aaronsteers: @edgarrmondragon - I don't know, honestly. We don't rely on the validator taking any action from these hints, so I thin it is fine to keep things as they are. Optionally, we can bump to Draft7 and confirm that nothing breaks. I don't have a strong feeling either way.

aaronsteers: We'll see! 

👀 https://github.com/meltano/sdk/pull/1096/commits/d9647bd412db8db7b2c91aeb49b7955576ed34b5

singer_sdk/typing.py
aaronsteers: Done.

edgarrmondragon (approved):

singer_sdk/typing.py
aaronsteers: Looks like no issues. 👍

edgarrmondragon:
pre-commit.ci autofix

PR #1111, opened 2022-10-26
fix: push _MAX_RECORDS_LIMIT down into SQL

### LINKED ISSUE #1110: Bug: SQLStreams do not push `_MAX_RECORDS_LIMIT` down to SQL query
### Singer SDK Version

0.13.0

### Python Version

3.9

### Bug scope

### Operating System

MacOS

### Description

When using the standard tap tests in pytest with `tap-snowflake` on the `tpcds_sf100tcl.catalog_returns` Snowflake sample dataset, I noticed that no `LIMIT` was applied in the query submitted to snowflake. This means the entire 557.7M row dataset is processed to be returned for the purposes of testing, when the stream will otherwise terminate after the first record arrives. Using the `_MAX_RECORDS_LIMIT` down to the SQL query would dramatically reduce the time and cost associated with Snowflake testing, and similarly reduce the load on other database engines due to testing, without compromising the validity of each test.

### Code

_No response_

kgpayne:
Closes #1110 

- #1110 

----
:books: Documentation preview :books:: https://meltano-sdk--1111.org.readthedocs.build/en/1111/

### DIFF
--- a/singer_sdk/streams/sql.py
+++ b/singer_sdk/streams/sql.py
@@ -181,6 +181,7 @@ class SQLStream(Stream, metaclass=abc.ABCMeta):
             column_names=selected_column_names,
         )
         query = table.select()
+
         if self.replication_key:
             replication_key_col = table.columns[self.replication_key]
             query = query.order_by(replication_key_col)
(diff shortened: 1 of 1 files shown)


kgpayne:
@aaronsteers @edgarrmondragon this solves the performance issue I was seeing [here](https://github.com/MeltanoLabs/tap-snowflake/issues/5#issuecomment-1290867696) (making automated tests for `tap-snowflake` [pass](https://github.com/MeltanoLabs/tap-snowflake/actions/runs/3328839431) 🚀) however, in the same vein as min and max constraints (discussed [here](https://github.com/meltano/sdk/discussions/1011)) I think the limit should also be passed in the `get_records` and `get_batches` signatures. Have added a note to the min/max discussion 🙂

aaronsteers (approved):
LGTM!

PR #1114, opened 2022-10-26
refactor: make DDL overridable for column `ADD`, `ALTER`, and `RENAME` operations

### LINKED ISSUE #1033: feat: allow `ALTER TABLE` and `ALTER COLUMN` syntax to be overridden
### Feature scope

### Description

Snowflake uses a different syntax for altering columns than that implemented by the `_adapt_column_type` method on `SQLConnector`. To account for these differences, we could i) use a SQLAlchemy abstraction for altering columns (if one exists) or ii) provide an overridable public method for target developers to bring their own syntax.

kgpayne:
Closes #1033 

- #1033 

----
:books: Documentation preview :books:: https://meltano-sdk--1114.org.readthedocs.build/en/1114/


kgpayne:
Codecov is down by 0.52% ⬇️ As this is a refactor, and as I have a handful of other things for tap/target snowflake to get through to get to done, I'd prefer not to add testing on the refactored methods unless anyone feels very strongly about them 🙂

singer_sdk/connectors/sql.py
```diff
@@ -871,6 +860,87 @@ def _get_column_type(
 
         return cast(sqlalchemy.types.TypeEngine, column.type)
 
+    @staticmethod
+    def get_column_create_ddl(
```
aaronsteers, line 864: Should we rename this to `get_column_add_add()`? The verb `CREATE` is normally applied to tables, but I think we generally use `ADD` for columns. Wdyt?

I don't have a strong opinion here, so feel free to chose what feels best. 🤷

aaronsteers:
> Codecov is down by 0.52% ⬇️ As this is a refactor, and as I have a handful of other things for tap/target snowflake to get through to get to done, I'd prefer not to add testing on the refactored methods unless anyone feels very strongly about them 🙂

Do you mind adding a simple test for the DDL return values? Specifically, I'm thinking of a test for the static methods which are easily tested from a class reference and don't need the a full sink object to be created. If they return the DDL or string result expected when passed specific input values, I think that's probably good.

aaronsteers (approved):
For my part, I think I'm okay with code coverage as is here - especially since we are also working on:

- #1123 

Thanks for adding the additional tests for the SQL generation.

edgarrmondragon (approved):
lgtm, the SQL API is coming together! @kgpayne the dependency review error is a false positive afaict, so feel free to merge 😄

PR #1115, opened 2022-10-26
chore: Add support for python 3.11

laurentS:
This just adds a few lines to run CI tests on the recently released python 3.11.

Substantial speedups are expected, and as the taps and targets are run inside their own virtual envs, it should be pretty easy to upgrade them even if the main app is running on an earlier version of python.

----
:books: Documentation preview :books:: https://meltano-sdk--1115.org.readthedocs.build/en/1115/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,6 +76,11 @@ docs = [
     "myst-parser",
     "sphinx-autobuild",
 ]
+# Mark these as extras so that installs pass on python 3.11
+devextras = [
+    "pyarrow",
+    "viztracer",
+]
 
 [tool.poetry.dev-dependencies]
 # snowflake-connector-python = "2.0.4" # Removed: Too many version conflicts!
@@ -85,9 +90,9 @@ xdoctest = "^1.1.0"
 mypy = "^0.982"
 cookiecutter = "^2.1.1"
 PyYAML = "^6.0"
-pyarrow = "^9.0.0"
+pyarrow = {version = "^9.0.0", optional = true, python = "<3.11"}
 freezegun = "^1.2.2"
-viztracer = "^0.15.4"
+viztracer = {version = "^0.15.4", optional = true, python = "<3.11"}
 requests-mock = "^1.10.0"
 sqlalchemy2-stubs = {version = "^0.0.2a29", allow-prereleases = true}
 types-python-dateutil = "^2.8.19"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ classifiers = [
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: Implementation :: CPython",
     "Topic :: Software Development :: Libraries :: Application Frameworks",
     "Typing :: Typed",
@@ -37,7 +38,7 @@ license = "Apache 2.0"
 "Youtube" = "https://www.youtube.com/meltano"
 
 [tool.poetry.dependencies]
-python = "<3.11,>=3.7.1"
+python = "<3.12,>=3.7.1"
 backoff = ">=2.0.0,<3.0"
 pendulum = "^2.1.0"
 click = "~=8.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ fs = "^2.4.16"
 PyJWT = "~=2.4"
 requests = "^2.25.1"
 cryptography = ">=3.4.6,<39.0.0"
-importlib-metadata = {version = "^4.13.0", markers = "python_version < \"3.8\""}
+importlib-metadata = {version = "<5.0.0", markers = "python_version < \"3.8\""}
 importlib-resources = {version = "5.9.0", markers = "python_version < \"3.9\""}
 memoization = ">=0.3.2,<0.5.0"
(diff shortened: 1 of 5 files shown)


edgarrmondragon:
Hey @laurentS!

You'll need to add `3.11` to the known version in `noxfile.py`:

```diff
- python_versions = ["3.10", "3.9", "3.8", "3.7"]
+ python_versions = ["3.11", "3.10", "3.9", "3.8", "3.7"]
```

laurentS:
Ok, it looks liks `numpy` is brought in by `pyarrow` which itself is working on supporting 3.11 (see [this issue](https://github.com/apache/arrow/pull/14499)). Looks like they're still supporting python 3.7, so chances are they'll fix the problem for us :)
I'll keep an eye on it, and come back here when there's a new release over there. I have a feeling it won't take long.

aaronsteers:
Agreed we can probably wait this out if pyarrow/numpy updates are coming soon.

Since `pyarrow` is only a [dev dependency](https://github.com/oviohub/meltano-sdk/blob/add-support-for-python-3-11/pyproject.toml#L88), we can optionally remove it at the cost of disabling/removing the `target-parquet` sample implementation. This is the only place where the `pyarrow` dependency is used as of now:

https://github.com/meltano/sdk/blob/6adeead98c25e13092bc657aaa3593234e1eed84/samples/sample_target_parquet/parquet_target_sink.py#L7-L8

laurentS:
I'm wondering if these dependencies that are only used for samples would be better managed as `extras`, like is done for `docs`?

edgarrmondragon:
> I'm wondering if these dependencies that are only used for samples would be better managed as `extras`, like is done for `docs`?

@laurentS Poetry would still resolve those dependencies along with the rest even if they're extras. The samples would need their own separate dependency trees/`pyproject.toml`.

laurentS:
I'm seeing 2 packages that cause problems with python 3.11:
- `pyarrow`, used only for the target-parquet sample code. I feel like there's value in having tested code samples that are known to run, but also wonder if it's worth blocking the entire tap+target ecosystem in a situation like this. In order to progress on this PR, I've marked the test as skipped for python 3.11, so as to avoid dropping checks that were in place. Once `pyarrow` releases a new version, we can re-enable the test and imports for all versions of python.
- `viztracer` Can I suggest removing it entirely from the `pyproject.toml` file? This is a debug tool, I think it's enough to have the docs mention how to install it, and nothing else. Right now, it's being installed multiple times on each CI run, without being used. It costs CI time :clock1: and carbon footprint :cloud: 

At this point, all checks are green, with a bit of a caveat for 3.11. I had to adjust dependency specifications for older pythons, as newer releases came out recently that broke tests.

edgarrmondragon:
> * `viztracer` Can I suggest removing it entirely from the `pyproject.toml` file? This is a debug tool, I think it's enough to have the docs mention how to install it, and nothing else. Right now, it's being installed multiple times on each CI run, without being used. It costs CI time 🕐 and carbon footprint ☁️

Agreed 100%. See https://github.com/meltano/sdk/issues/1124.

edgarrmondragon:
> * `pyarrow`, used only for the target-parquet sample code. I feel like there's value in having tested code samples that are known to run, but also wonder if it's worth blocking the entire tap+target ecosystem in a situation like this. In order to progress on this PR, I've marked the test as skipped for python 3.11, so as to avoid dropping checks that were in place. Once `pyarrow` releases a new version, we can re-enable the test and imports for all versions of python.

@laurentS Sounds good. Skipping the test on 3.11 is a good approach 👍

pyproject.toml
```diff
@@ -37,16 +38,16 @@ license = "Apache 2.0"
...
 pendulum = "^2.1.0"
 click = "~=8.0"
 fs = "^2.4.16"
 PyJWT = "~=2.4"
 requests = "^2.25.1"
 cryptography = ">=3.4.6,<39.0.0"
-importlib-metadata = {version = "*", markers = "python_version < \"3.8\""}
-importlib-resources = {version = "^5.9.0", markers = "python_version < \"3.9\""}
+importlib-metadata = {version = "<5.0.0", markers = "python_version < \"3.8\""}
+importlib-resources = {version = "5.9.0", markers = "python_version < \"3.9\""}
```
edgarrmondragon, line 50: These updates have certainly been problematic. If anyone's curious:

- https://github.com/meltano/sdk/pull/1023#issuecomment-1267796121
- https://github.com/meltano/sdk/pull/1057#issuecomment-1273610294

There's a whole story of maintainer disagreements between `flake8` and `importib-metadata` 😬

pyproject.toml
```diff
@@ -75,6 +76,11 @@ docs = [
     "myst-parser",
     "sphinx-autobuild",
 ]
+# Mark these as extras so that installs pass on python 3.11
+devextras = [
+    "pyarrow",
+    "viztracer",
+]
```
edgarrmondragon, line 83: @laurentS wdyt of splitting these between two extras:

```
samples = ["pyarrow"]
debugging = ["viztracer"]
```

We may want to [drop viztracer](https://github.com/meltano/sdk/issues/1124) as you suggest but it's probably not worth increasing the scope of this PR.

laurentS: done in 740f547

edgarrmondragon (approved):

aaronsteers (approved):
With the latest updates, this is looking good to me. 👍 

Thanks very much.

edgarrmondragon:
Alright @laurentS, this LGTM! Thanks so much for making all the necessary updates in tests and dependencies.

PR #1120, opened 2022-10-27
feat: add `--discover` capability for all SQL targets

aaronsteers:
This was an experiment which actually turned out to be simpler than expected.

Todo:

- [ ] Tests
- [ ] Docs
- [ ] Discussion

Note:

The primary value of this operation is to be able to analyze and version the schema of targets. This is important because when trying to understand destination schema in the target, we can't necessarily assume that the target schema will be identical with the schema of upstream tap streams. Stream maps and selection rules can change this, of course, but also factors like `schema_mappings` and `_sdc_*` metadata columns.

Potentially this could be extended further to actually querying data from the target directly - but that is not in scope any time in the near future.

----
:books: Documentation preview :books:: https://meltano-sdk--1120.org.readthedocs.build/en/1120/

### DIFF
--- a/singer_sdk/cli/common_options.py
+++ b/singer_sdk/cli/common_options.py
@@ -35,3 +35,9 @@ PLUGIN_FILE_INPUT = click.option(
     help="A path to read messages from instead of from standard in.",
     type=click.File("r"),
 )
+
+PLUGIN_DISCOVERY = click.option(
+    "--discover",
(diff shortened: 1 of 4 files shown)


aaronsteers:
pre-commit.ci autofix

edgarrmondragon:
Closing for now but we might revisit in the future

PR #1126, opened 2022-10-28
fix: resolve failure in `_increment_stream_state()` for cases when `replication_method` is `LOG_BASED`

aaronsteers:
I did not add any extra tests.

Would be good to pick those up and add official handling for LOG_BASED incremental at the same time.

Reviewing with whitespace changes hidden will help to highlight changed lines. 

While creating this PR, I remembered that `get_context_state()` is _not_ a no-op, so I added a comment to help future maintainers.

----
:books: Documentation preview :books:: https://meltano-sdk--1126.org.readthedocs.build/en/1126/


aaronsteers:
Codecov CI failure is basically same as before; no increase or decrease per se.

Issue #1086, opened 2022-10-19
For SQL-based taps, add built-in handling for "default load schema" and "schema mappings"


aaronsteers:
As discussed in:

- #1084

BuzzCutNorman:
@aaronsteers what do you think about adding this one to the milestone SQL Target Out-of-Box Stability

aaronsteers:
Good idea! Otherwise (at least without `default_load_schema`), certain expected behaviors do not work.

aaronsteers:
Regarding the naming of these, I originally was inclined to remove the `default_` prefix from `default_[load|target]_schema`. Then I realized that the `default_` prefix is in part to communicate that the 'default' will be overriden when `schema_mappings` are _also_ provided.

So, I'd be inclined to use either:

```
schema_mappings
default_load_schema
```

Or just keep with pipelinewise precedent:

```
schema_mappings
default_target_schema
```

With all other factors the same, I don't have strong enough preference of `default_load_schema` being _vastly_ better than `default_target_schema` - and for perhaps a very minor improvement in phrasing, I don't know that it is worth it to break from precedent and now have both floating around in otherwise very similar configuration specs.

@tayloramurphy, @kgpayne, @edgarrmondragon (and all):

**Q:** What do you think of just keeping `schema_mappings` and `default_target_schema` as our SDK-default setting names when we implement this?

BuzzCutNorman:
@aaronsteers @kgpayne I just ran an issue when I bummped my `target-mssql` to SDK 0.13.0 where `tap-postgres` is sending a stream with `public-` as a prefix so the target tries to create the schema `public`.  That is when I get the following error.  

```
sqlalchemy.exc.ProgrammingError: (pyodbc.ProgrammingError) ('42S01', "[42S01] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]There is already an object named 'public' in the database. (2714) (SQLExecDirectW); [42S01] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]CREATE SCHEMA failed due to previous errors. (2759)")
```
I was able to override the ` @property def schema_name(self)` and set it to return a sting of `dbo` and it worked.  I would now need to add the `default_target_schema` to the config and set the property to return it value if it is present.

WDTY?  Good idea, ok idea not the best, or there be monsters!!!

aaronsteers:
[...]

PR #1157, opened 2022-11-08
feat: Built-in handling of `default-target-schema` for SQL Targets

BuzzCutNorman:
This is an attempt to add support for a `default_target_schema` , as mentioned in issue #1086, into the SDK for SQL Targets.

----
:books: Documentation preview :books:: https://meltano-sdk--1157.org.readthedocs.build/en/1157/


singer_sdk/helpers/capabilities.py
```diff
@@ -47,6 +48,13 @@
         description="The max depth to flatten schemas.",
     ),
 ).to_dict()
+TARGET_SCHEMA_CONFIG = PropertiesList(
+    Property(
+        "default_target_schema",
+        StringType(),
+        description="The Default schema to place all streams",
```
aaronsteers, line 55: Small wording tweak suggested here, mostly because the word 'schema' is overloaded and can sometimes mean 'JSON schema'.

```suggestion
        description="The default schema name to use for all streams",
```

BuzzCutNorman: What do you think about  `description="The default target database schema name to use for all streams",`

singer_sdk/sinks/sql.py
```diff
@@ -82,11 +82,31 @@ def schema_name(self) -> Optional[str]:
...
             # Stream name is a two-part or three-part identifier.
             # Use the second-to-last part as the schema name.
-            return self.conform_name(parts[-2], "schema")
+            stream_schema = self.conform_name(parts[-2], "schema")
+
+            # MS SQL Server has a public database role so the name is reserved
+            # and it can not be created as a schema.  To avoid this common error
+            # we convert "public" to "dbo" if the target dialet is mssql
+            if target_sqla_dialect == "mssql" and stream_schema == "public":
+                return "dbo"
```
edgarrmondragon, line 107: I'm not sure if the SDK should be aware of this mssql restriction other than providing a way to override/extend functionality to handle it. Without this check, an example implementation in a target-mssql could be:

```python
class MSSQLSink(SQLSink):
    def schema_name(self):
        stream_schema = super().schema_name()
        if self.connector._dialect == "mssql" and stream_schema == "public":
            return "dbo"
        else:
     	    return stream_schema
```

edgarrmondragon (changes requested):

singer_sdk/helpers/capabilities.py
aaronsteers: Yeah, sounds great!

singer_sdk/sinks/sql.py
```diff
@@ -82,11 +82,31 @@ def schema_name(self) -> Optional[str]:
         Returns:
             The target schema name.
         """
+        # Get the current SQL Dialect being used
+        target_sqla_dialect = self.connection.engine.dialect.name
```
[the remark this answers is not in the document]
BuzzCutNorman: Thanks, that is very helpful to know.  Since I am moving it from the SDK `SQLSink` to `MSSQLSink` I was going to drop the dialect check since I know it has been set to `mssql` or would you recommend leaving it in place.

singer_sdk/sinks/sql.py
BuzzCutNorman: Good call.  I will get this moved into the `MSSQLSink` and out of `SQLSink`.  Thanks for the code 🙏

singer_sdk/sinks/sql.py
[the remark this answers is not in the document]
edgarrmondragon: @BuzzCutNorman yeah it makes sense to just remove this line in the SDK 👍

singer_sdk/sinks/sql.py
```diff
@@ -82,11 +82,31 @@ def schema_name(self) -> Optional[str]:
...
+        # Look for a default_target_scheme in the configuraion fle
+        default_target_schema = self.config.get("default_target_schema", None)
```
edgarrmondragon, line 88:
```suggestion
        default_target_schema: str = self.config.get("default_target_schema", None)
```

edgarrmondragon (changes requested):

singer_sdk/target_base.py
```diff
@@ -571,4 +576,46 @@ def cli(
 class SQLTarget(Target):
     """Target implementation for SQL destinations."""
 
+    @classproperty
+    def capabilities(self) -> List[CapabilitiesEnum]:
```
edgarrmondragon, line 580:
```suggestion
    def capabilities(self) -> list[CapabilitiesEnum]:
```

singer_sdk/target_base.py
```diff
@@ -571,4 +576,46 @@ def cli(
...
+        """Get target capabilities.
+
+        Returns:
+            A list of capabilities supported by this target.
+        """
+        sql_target_capabilities: List[CapabilitiesEnum] = super().capabilities
```
edgarrmondragon, line 586:
```suggestion
        sql_target_capabilities: list[CapabilitiesEnum] = super().capabilities
```

edgarrmondragon (changes requested):
This module supports [PEP 585](https://peps.python.org/pep-0585/)

singer_sdk/sinks/sql.py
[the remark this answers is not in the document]
BuzzCutNorman: @edgarrmondragon the line `target_sqla_dialect = self.connection.engine.dialect.name` has been removed

singer_sdk/sinks/sql.py
BuzzCutNorman: @BuzzCutNorman I removed the checking for `stream_schema == "public"` for mssql from the SDK and got it added into my target-mssql.

singer_sdk/target_base.py
```diff
@@ -571,4 +576,46 @@ def cli(
...
+        Returns:
+            A list of capabilities supported by this target.
+        """
+        sql_target_capabilities: list[CapabilitiesEnum] = super().capabilities
+        sql_target_capabilities.extend([TargetCapabilities.TARGET_SCHEMA])
+
+        return sql_target_capabilities
+
+    @classmethod
+    def append_builtin_config(self, config_jsonschema: dict) -> None:  # noqa: ANN102
```
BuzzCutNorman, line 592: @edgarrmondragon I don't know how to fix this and was wondering if you could weigh in.  If I remove the `@classmethod` then `meltano invoke target-mssql --about` fails but when I leave it in `pre-commit` checks fail asking me to specify a class type for cls.  That is why I have `# noqa: ANN102`

edgarrmondragon: @BuzzCutNorman the `cls` parameter needs a class annotation, i.e.

```python
@classmethod
def append_builtin_config(cls: type[SQLTarget], config_jsonschema: dict) -> None:
```

I've pushed this change because I wasn't sure myself if that was the sole issue.

In the future, we should probably just ignore `ANN102` in the whole repo.

BuzzCutNorman: @edgarrmondragon ahh got it.  Thanks 🙏 for that.  `target-mssql --about` tested good post commit 🎉

edgarrmondragon (approved):

edgarrmondragon:
Forgot to merge this one 😅.

Thanks @BuzzCutNorman!

PR #1168, opened 2022-11-10
fix: `SQLConnector.table_exists()` to use separate `table_name` and `schema_name` instead of fully qualified name

### LINKED ISSUE #1167: [Bug]: Function 'table_exists()' returning false when the table does exist in database
### Singer SDK Version

0.13.1

### Python Version

3.9

### Bug scope

### Operating System

Windows

### Description

When an SDK target checks to see if a table exists with `table_exists()` it always come back as false if the `full_table_name` is in `schema.table` format or `database.schema.table` format.  That is happening because the function `sqlalchemy.engine.reflection.Inspector.has_table` is not able to deal with two part of three part table names.

image
image
image
image

### Code

```python
def table_exists(self, full_table_name: str) -> bool:
        """Determine if the target table already exists.

        Args:
            full_table_name: the target table name.

        Returns:
            True if table exists, False if not, None if unsure or undetectable.
        """
        return cast(
            bool,
            sqlalchemy.inspect(self._engine).has_table(full_table_name),
        )
```

BuzzCutNorman:
Attempt to fix `table_exists()` to use the parsed `full_table_name` parts table and schema instead of a combined multipart name.

Fixes #1167  

----
:books: Documentation preview :books:: https://meltano-sdk--1168.org.readthedocs.build/en/1168/

### DIFF
--- a/singer_sdk/connectors/sql.py
+++ b/singer_sdk/connectors/sql.py
@@ -491,7 +491,7 @@ class SQLConnector:
         else:
             return cast(
                 bool,
-                sqlalchemy.inspect(self._engine).has_table(table_name, schema_name),
+                sqlalchemy.inspect(self._engine).has_table(table_name),
             )
 
     def schema_exists(self, schema_name: str) -> bool:
--- a/singer_sdk/connectors/sql.py
+++ b/singer_sdk/connectors/sql.py
@@ -483,16 +483,11 @@ class SQLConnector:
         )
         table_name: str = pftn[2]
         schema_name: str = pftn[1]
-        if schema_name:
-            return cast(
-                bool,
-                sqlalchemy.inspect(self._engine).has_table(table_name, schema_name),
-            )
-        else:
-            return cast(
-                bool,
-                sqlalchemy.inspect(self._engine).has_table(table_name),
-            )
+
+        return cast(
+            bool,
+            sqlalchemy.inspect(self._engine).has_table(table_name, schema_name),
+        )
 
     def schema_exists(self, schema_name: str) -> bool:
         """Determine if the target database schema already exists.
--- a/singer_sdk/connectors/sql.py
+++ b/singer_sdk/connectors/sql.py
@@ -669,9 +669,7 @@ class SQLConnector:
             partition_keys: list of partition keys.
             as_temp_table: True to create a temp table.
         """
-        self.logger.info(f"Preparing table {full_table_name}")
         if not self.table_exists(full_table_name=full_table_name):
-            self.logger.info(f"Table {full_table_name} does not exist")
             self.create_empty_table(
                 full_table_name=full_table_name,
                 schema=schema,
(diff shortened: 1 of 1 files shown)


BuzzCutNorman:
@edgarrmondragon when you get a chance, please take a look.  I left the logging in to show the points I was looking at.  I am planning on removing them before I click `ready for review`.

singer_sdk/connectors/sql.py
```diff
@@ -477,9 +477,16 @@ def table_exists(self, full_table_name: str) -> bool:
         Returns:
             True if table exists, False if not, None if unsure or undetectable.
         """
+        # pftn -> Parsed Full Table Name [0] = db, [1] = schema, [2] = table
+        pftn: tuple = SQLConnector.parse_full_table_name(
+            self, full_table_name=full_table_name
+        )
+        table_name: str = pftn[2]
+        schema_name: str = pftn[1]
```
aaronsteers, line 485: Since `parse_full_table_name()` always return 3 parts, this should also work:

```py
        db_name, schema_name, table_name = SQLConnector.parse_full_table_name(
            self, full_table_name=full_table_name
        )
```

Or if we know we won't use the db_name part:

```py
        _, schema_name, table_name = SQLConnector.parse_full_table_name(
            self, full_table_name=full_table_name
        )
```

Wdyt?

singer_sdk/connectors/sql.py
```diff
@@ -477,9 +477,16 @@ def table_exists(self, full_table_name: str) -> bool:
...
+
         return cast(
             bool,
-            sqlalchemy.inspect(self._engine).has_table(full_table_name),
+            sqlalchemy.inspect(self._engine).has_table(table_name, schema_name),
```
aaronsteers, line 489: I think for the most part, we usually expect a single DB as source.

However, if the table in question does reference something other than the default db (e.g. if `full_table_name` is in `db.schema.table` format), we may want to do something here to handle db_name as well. That said, I don't know what the generic way to do this would be, and probably we can leave omitted if the parsing does not return an explicit db name.

What do you think?

BuzzCutNorman: The only inputs has_table() takes is table name and scheme as far as I can tell.

https://docs.sqlalchemy.org/en/14/core/reflection.html#sqlalchemy.engine.reflection.Inspector.has_table

edgarrmondragon: @BuzzCutNorman yeah, the db is set at the connection/engine level 👍

BuzzCutNorman: Though in a target instance I think I would like it to only look at the database I specified in the connection.  which like you said should be a single database.  I know you can have fall back connections to different servers, but I haven't seen different databases given in SQLA examples.

singer_sdk/connectors/sql.py
BuzzCutNorman: Those are much nicer than what I did.  Just tested the second option and worked great.  also changed to `self` from `SQLConnector`.

singer_sdk/connectors/sql.py
aaronsteers: @BuzzCutNorman + @edgarrmondragon - Yeah, makes sense. 👍

edgarrmondragon (approved):

edgarrmondragon:
LGTM!

Thanks, @BuzzCutNorman 😄

PR #1175, opened 2022-11-11
feat(mappers): Add `datetime` functions to simpleeval env in stream maps

### LINKED ISSUE #986: [Feature]: Datetime handling in streammaps
### Feature scope

### Description

Being able to reference datetime (or equivalent) Python objects in streammaps statements would be very helpful.

The simple way to do this is [mentioned here](https://github.com/meltano/sdk/issues/922#issuecomment-1224184583): grafting a few custom datetime uses into streammaps, similarly to the way it was done with `md5`. Anything more complex than that (I think) would dramatically increase the lift, because you'd likely have to change the way statements are evaluated (currently simpleeval).

A few cases where this would be useful:
- Filtering out rows that were created <24 hours ago (would need something like "datetime.now()")
- Filtering out rows that were created on some specific date or date range (e.g. date of a known bug that created bad data)

qbatten:
Closes #986

This adds the datetime module to the simpleeval environment, similarly to how an md5 function was added. This enables datetime handling/processing in streammaps.

----
:books: Documentation preview :books:: https://meltano-sdk--1175.org.readthedocs.build/en/1175/

### DIFF
--- a/docs/stream_maps.md
+++ b/docs/stream_maps.md
@@ -160,6 +160,8 @@ can be referenced directly by mapping expressions.
     of the hash's hex digest.
   - This is defined by the SDK internally with native python:
     `hashlib.md5(<input>.encode("utf-8")).hexdigest()`.
+- `datetime` - This is the datetime module object from the Python standard library. You can access
+    datetime.datetime, datetime.timedelta, etc.
 
 #### Built-in Variable Names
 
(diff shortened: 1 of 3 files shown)


edgarrmondragon (approved):
Thanks, @qbatten!

This is a great first contribution 👍

Issue #1176, opened 2022-11-11
[Bug]: SQL connector forces column names to lower case but does not 


hsyyid:
### Singer SDK Version

0.13.0

### Python Version

3.10

### Bug scope

Targets (data type handling, batching, SQL object generation, etc.)

### Operating System

Linux

### Description

Pushing the following payload to `target-mssql` (https://github.com/radbrt/target-mssql/blob/main/target_mssql/connector.py):

```
{"type": "SCHEMA", "stream": "GeneralLedgerReport", "schema": {"type": ["object", "null"], "properties": {"Date": {"type": ["string", "null"]}, "CompanyName": {"type": ["string", "null"]}, "TransactionType": {"type": ["string", "null"]}, "Num": {"type": ["string", "null"]}, "Name": {"type": ["string", "null"]}, "Memo": {"type": ["string", "null"]}, "Split": {"type": ["string", "null"]}, "Class": {"type": ["string", "null"]}, "Department": {"type": ["string", "null"]}, "Amount": {"type": ["number", "null"]}, "Categories": {"type": ["string", "null"]}, "id": {"type": ["string", "null"]}, "entity_name": {"type": ["string", "null"]}, "entity_type": {"type": ["string", "null"]}}}, "key_properties": ["id"]}
{"type": "RECORD", "stream": "GeneralLedgerReport", "record": {"Date": "2022-01-01 00:00:00", "CompanyName": "Honeycomb Holdings Inc.", "TransactionType": "Invoice", "Num": "1100", "Name": "A1 NO_LASTNAME", "Split": "1100 Accounts Receivable", "Amount": 20.0, "Categories": "AcctRec : 1100 Accounts Receivable", "id": "1948988_0", "entity_name": "A1 NO_LASTNAME", "entity_type": "Customer"}}
```

As you can see the column names are capitalized. e.g. `Date` instead of `date`

The Meltano SDK forces all column names to be lowercase here: https://github.com/meltano/sdk/blob/main/singer_sdk/connectors/sql.py#L523

However, when checking if columns exist, it does not do the lookup with lowercase:
https://github.com/meltano/sdk/blob/main/singer_sdk/connectors/sql.py#L547

So in my case, it creates the table with column `date` and then checks if the column `Date` exists. It doesn't (due to the difference in case), so it tries to create again and fails because it tries to create a column `date` again. This fails the entire target.

Solution would be to force the `column_exists` function to also force column names to lowercase so targets don't deal with this weird behavior

[...]

PR #1183, opened 2022-11-14
wip: Trigger rename of unconformed columns

BuzzCutNorman:
Here is a proof-of-concept fix/feature that finds unconformed column names and triggers a rename of the column to the conformed version.  This came about while looking at #1176 which I think is an instance of a pre SDK 0.13.0 target which created tables and when the default `conform_name` started working the table column names don't match and is causing an issue.

----
:books: Documentation preview :books:: https://meltano-sdk--1183.org.readthedocs.build/en/1183/


BuzzCutNorman:
@edgarrmondragon @aaronsteers This is an idea for a way to help get sql targets to meet a conformed naming convention for column names.  Please take a look and let me know your thoughts.

edgarrmondragon:
@BuzzCutNorman I think this approach is sensible. This is still a draft so I'm not sure if you're planning on still making changes, but can you can add one or more tests that validate the fix?

BuzzCutNorman:
@edgarrmondragon Thanks for taking a look.  I had aspirations of adding the same logic to tables and or schemes since `conform_name` could change them as well.  I am starting to think that isn't a great idea.  More than happy to attempt adding a few tests.

BuzzCutNorman:
@edgarrmondragon the requested covering tests were added.  When adding these I also added `from unittest.mock import patch` to `test_connector_sql.py` to use in my test.  I hope that is ok I just needed a way to `@patch` in result for a function and ran across this way of doing it.

BuzzCutNorman:
> @BuzzCutNorman and @edgarrmondragon - Renaming existing columns to conform naming convention is a novel approach and I want to spend a bit more time thinking about the repercussions and use cases.
> 
> A few things give me pause:
> 
> 1. Generally the _primary_ purpose of name conformance is to create column names that are legal in the destination system. So, renaming columns that already exist there seems a bit off from the main case.

You are correct I didn't write the PR to assist with the original purpose.  It was as an attempt to reduce the number of duplicate columns or column addition errors that arise when running a target that is conforming column names in existing tables that have an allowed name.

> 2. Renaming existing columns is definitely a breaking change for downstream consumers like dbt.
>    
>    * And there doesn't seem to be a good way for dbt to fail gracefully.

I agree, but even if you drop all the tables and do a full load aren't you in the exact same position?

> 3. If we do add this, we'll need to ensure that our names do not conform into the same text. The classic example of this is a pair of properties (adapted from a real case) called `reactions_+1` and `reactions_-1`.
>   

If this PR is not in place the conformed names for these columns would be identical and you would be putting the new results in the conformed name column and the other original two columns would be untouched.  But you would have a new column in the table with a merge of both columns data going forward.

If this PR was in place the first column given would have been renamed.  The rename of the second column would have stopped the process with a duplicate column name error which could never be resolved without turning off the `conform_name` function in the target's code.

>    * This is not unique here, but it becomes a larger factor.
> 4. I'm not entirely sure that this solves for 'slightly conformed' column names being able to be renamed into 'better conformed' column names. E.g. `MyEC2Column` might get conformed to `my_e_c2_column` in "v1" but "v2" is smarter and would suggest: `my_ec2_column`. However, the existing column `my_e_c2_column` doesn't violate any rules on its own and can't be identified as a prior-conformed version, unless we keep histories of all versions of our conform logic - which itself is problematic because `my_e_c2_column` _might_ not be the same thing as we think it is, and the more logic we add the more possible false-positive merges we may end up with.
> 

You are correct this PR would not solve this issue since `conform_name` is generating changes that may be best handled by a schema registry.

> I guess in summary, getting this right seems really hard, and I'm not sure it would ever be fully 'safe'.

These are all great points and I agree this is hard to do and this PR is an imperfect way to handle some of the changes.

> 
> What do yo think?

I like that the SDK is assisting end users by making safe names for SQL targets.  After working on this PR I think my preference would be to make `conform_name` a developer option like `allow_temp_tables`, so developers can have time to setup `conform_name` rules that work with their version of SQL.  I believe there should be a default option for `conform_name` in the target configuration.  This would allow the pipeline admin the ability to turn `conform_name` on or off if allowed by the target.  The target configuration option will allow pipeline admins to work on the changes in their dev environment before making changes in production.  A configuration option also allows them to keep an original target configuratin with `conform_name` off and add an inherited version with `conform_name` on.

cc @edgarrmondragon @kgpayne

edgarrmondragon:
> I like that the SDK is assisting end users by making safe names for SQL targets. After working on this PR I think my preference would be to make `conform_name` a developer option like `allow_temp_tables`, so developers can have time to setup `conform_name` rules that work with their version of SQL.

@BuzzCutNorman I like that option better than the current situation. Would be willing to file an issue for that?

BuzzCutNorman:
I just wanted to check and see if anyone had objections to me closing this PR?  I am assuming there won't be since the consensus was this wasn't the best way to correct the situation I described.  Thought it would be best to ask instead of assuming. 😁

edgarrmondragon:
@BuzzCutNorman thanks for your patience and understanding! Closing now...

PR #1192, opened 2022-11-16
fix(taps): Change max record age for emitting state messages to 5 instead of 30 mins

### LINKED ISSUE #1064: Bug: If one stream has an error, then all streams fail and no STATE messages are ever sent
### Singer SDK Version

0.11.1

### Python Version

3.9

### Bug scope

### Operating System

Linux

### Scenario

* Your target is using incremental replication
* Your tap is sending multiple streams
* Each stream gets sent to its own `Sink`
* After a while, one of the Sinks throws an exception. For example if the tap sent a bad value in a record.
* This exception stops everything. The `Target` and all `Sink`s stop.

### What is the actual result

Not a single `STATE` message is emitted by the target during this process even though it might have spent hours processing data already. Which means you need to start over from scratch. For ALL streams.

### What is the expected result

The `TARGET` or `SINK`s should emit state messages periodically so state is not lost.

### Code

_No response_

spacecowboy:
Fixes #1064

----
:books: Documentation preview :books:: https://meltano-sdk--1192.org.readthedocs.build/en/1192/

### DIFF
--- a/singer_sdk/target_base.py
+++ b/singer_sdk/target_base.py
@@ -38,7 +38,7 @@ class Target(PluginBase, SingerReader, metaclass=abc.ABCMeta):
     object for that record.
     """
 
-    _MAX_RECORD_AGE_IN_MINUTES: float = 30.0
+    _MAX_RECORD_AGE_IN_MINUTES: float = 5.0
 
     # Default class to use for creating new sink objects.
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):
Thanks!

PR #1196, opened 2022-11-16
fix: Add a to dict method to RecordMessage to improve performance

Jack-Burnett:
In my testing, taps spend about half of their time converting record messages to dicts. This basically reduces that to nothing, doubling the throughput of the average tap.


edgarrmondragon (approved):
thanks @Jack-Burnett!

PR #1219, opened 2022-11-24
fix(mappers): Schema passthrough for whitelisted fields

edgarrmondragon:
Related:

- https://github.com/MeltanoLabs/meltano-map-transform/issues/30
- https://github.com/MeltanoLabs/meltano-map-transform/issues/66

We'll have to release a new version of the SDK, and bump it in MeltanoLabs/meltano-map-transform.

---

I changed the `test_mapped_stream` to use [snapshots](https://pypi.org/project/pytest-snapshot/) so we can validate the entire tap output.

----
:books: Documentation preview :books:: https://meltano-sdk--1219.org.readthedocs.build/en/1219/


aaronsteers:
@edgarrmondragon - This is great. 

I like this approach a lot, but the library is new to me and I'm not sure if I would know where to look to confirm the changes. So, if we fix the double-state messages at the end of each sync, I expect the job would fail. And the proper course for developer would be to run a snapshot update, then confirm the git diff of those snapshots that got updated, and _if_ the new schema files are better or at least as good, then those could be merged back to main.

Could we add something in `CONTRIBUTING.md` about the snapshots feature and how they should be maintained/updated?

Thanks!

edgarrmondragon:
> Could we add something in `CONTRIBUTING.md` about the snapshots feature and how they should be maintained/updated?

@aaronsteers I've updated the contributing page with more details: af1ef39853ff755f55294a5ee8da49849af1ee77, https://meltano-sdk--1219.org.readthedocs.build/en/1219/CONTRIBUTING.html#snapshot-testing

afolson (approved):

Issue #875, opened 2022-08-01
[Bug]: Recommended way of packaging stream schemas as JSON files should use `importlib.resources`


edgarrmondragon:
### Singer SDK Version

latest

### Python Version

NA

### Bug scope

Taps (catalog, state, stream maps, etc.)

### Operating System

Any

### Description

The currently recommended way (as used in the cookiecutter templates) to ship JSON files for the stream schemas relies on `__file__`, which doesn't play well with modern tools for packaging a Python application (such as a Singer tap/target) as a standalone binary, such as [PyOxidizer](https://pyoxidizer.readthedocs.io/en/stable/).

The solution is to use `importlib.resources` or the backport package `importlib_resources` on Python < 3.9.

This would also require changing the following:

1. drop support for raw string file path
2. discourage use of `Path` instances
3. change our type hints to use `importlib.resources.abc.Traversable`
4. add an `__init__.py` to the `schemas/` directory shipped with the cookiecutter

For a example, see https://github.com/edgarrmondragon/tap-bitso/pull/125. The PR is currently failing static type checks because type annotations are not compatible with the latest SDK.

### Code

```diff
+ from my_tap import schemas  # assumes a __init__.py exists in my_tap/schemas/
+ 
+ if sys.version_info >= (3, 9):
+     import importlib.resources as resources
+ else:
+     import importlib_resources as resources

- SCHEMAS_DIR = Path(__file__).parent / "./schemas"
+ SCHEMAS_DIR = resources.files(schemas)
```

edgarrmondragon:
Still relevant

edgarrmondragon:
Closed by #2130

PR #1254, opened 2022-12-07
feat: Read plugin version from distribution, with fallback to package

kgpayne:
In cases where the distribution name does not match the tap/target ID (also the package name by default) the `Version` reported in `--about` registers as `[could not be detected]`.

e.g. `tap-snowflake` with package name `tap_snowflake` distributed as `meltanolabs-tap-snowflake`, to avoid PyPI name clashes.

Related to [MeltanoLabs/tap-snowflake#8](https://github.com/MeltanoLabs/tap-snowflake/issues/8)

----
:books: Documentation preview :books:: https://meltano-sdk--1254.org.readthedocs.build/en/1254/

### DIFF
--- a/singer_sdk/plugin_base.py
+++ b/singer_sdk/plugin_base.py
@@ -4,6 +4,7 @@ import abc
 import json
 import logging
 import os
+import sys
 from collections import OrderedDict
 from pathlib import PurePath
 from types import MappingProxyType
@@ -162,7 +163,9 @@ class PluginBase(metaclass=abc.ABCMeta):
         Returns:
             The package version number.
         """
-        distribution = get_project_distribution(__file__)
+        # get the file location of the subclass
+        path = sys.modules[cls.__module__].__file__
+        distribution = get_project_distribution(path)
         if distribution:
             version = str(distribution.metadata["Version"])
         else:
--- a/singer_sdk/plugin_base.py
+++ b/singer_sdk/plugin_base.py
@@ -27,7 +27,7 @@ from singer_sdk import metrics
 from singer_sdk.configuration._dict_config import parse_environment_config
 from singer_sdk.exceptions import ConfigValidationError
 from singer_sdk.helpers._classproperty import classproperty
-from singer_sdk.helpers._compat import metadata
+from singer_sdk.helpers._compat import get_project_distribution, metadata
 from singer_sdk.helpers._secrets import SecretString, is_common_secret_key
 from singer_sdk.helpers._util import read_json_file
 from singer_sdk.helpers.capabilities import (
@@ -162,10 +162,14 @@ class PluginBase(metaclass=abc.ABCMeta):
         Returns:
             The package version number.
         """
-        try:
-            version = metadata.version(cls.name)
-        except metadata.PackageNotFoundError:
-            version = "[could not be detected]"
+        distribution = get_project_distribution(__file__)
+        if distribution:
+            version = distribution.metadata["Version"]
+        else:
+            try:
+                version = metadata.version(cls.name)
+            except metadata.PackageNotFoundError:
+                version = "[could not be detected]"
         return version
 
     @classproperty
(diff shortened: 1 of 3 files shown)


kgpayne:
Windows tests are failing, and I don't have an environment (or the familiarity of Windows+Python 😅) to debug. Given this doesn't cause a hard-fail (we simply get `[could not be detected]` in `--about` if no distribution _and_ no package is found), I feel comfortable leaving the platform-specific `xfail`. Maybe a Windows-familiar community member will help us out in a future iteration 👍

singer_sdk/plugin_base.py
```diff
@@ -162,10 +162,14 @@ def plugin_version(cls) -> str:
         Returns:
             The package version number.
         """
-        try:
-            version = metadata.version(cls.name)
-        except metadata.PackageNotFoundError:
-            version = "[could not be detected]"
+        distribution = get_project_distribution(__file__)
```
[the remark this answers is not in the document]
edgarrmondragon: tbf anything to do with `__file__` is a minefield (e.g. https://github.com/meltano/sdk/issues/875). It's problematic for standalone app distributions:

> ### `__file__` Ambiguity
> It isn’t clear whether `__file__` is actually required and what all is derived from existence of `__file__`. It also isn’t clear what `__file__` should be set to if it wouldn’t be a concrete filesystem path. Can `__file__` be virtual? Can it refer to a binary/archive containing the module?
> 
> Semantics of `__file__` need more clarification.

https://pyoxidizer.readthedocs.io/en/stable/pyoxidizer_technotes.html?#file-ambiguity

That's probably why it's failing on Windows.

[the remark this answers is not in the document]
kgpayne: This was easy in the end - `cls` of the property references the current class (i.e. the subclass), so it was just a case of getting the module and its file `path = sys.modules[cls.__module__].__file__`. For SDK taps/targets, this will return `tap.py` or `target.py` 🎉

tests/core/test_helpers.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def test_get_project_distribution():
+    """Test `get_project_distribution()`.
+
+    click is a representative example of a distributed package from our dependency tree.
+    Any similar singer_sdk dependency could be used in stead.
+    """
+    if platform.system() == "Windows":
+        pytest.xfail("Doesn't pass on windows.")
```
edgarrmondragon, line 17: Does this mean the package name can't be retrieved in Windows?

kgpayne: It means the lookup via distribution will not work, yes. The fallback to `metadata.version(cls.name)` was the primary mechanism before this change (it relies on the library name being the same as the class name), though I don't know how well it worked for Windows in the first place 🤔

Given the hurdles we have to jump through to try and infer the plugin version via distribution or library metadata, I still  think we are better off adding `__version__ = "<version>"` in `__init__.py` and importing that for `--about`. `poetry-dynamic-versioning` supports substitution from a `0.0.0` placeholder in any file (including `__init__.py` by default), and this approach is standard in other python packages 🤷‍♂️ This PR specifically addresses [this ask](https://github.com/meltano/sdk/pull/1241#pullrequestreview-1207289029) from @aaronsteers 👍

edgarrmondragon: @kgpayne Yeah, I'd say it's probably not worth spending time trying to solve a problem that's only present when migrating existing taps and targets that are published to PyPI. Wdyt?

cjohnhanson (approved):
approved, pending that one open thread, but looks good to me

kgpayne:
Closing in favour of `__version__ = "<version>"` in `__init__.py` solution:

> I still think we are better off adding `__version__ = "<version>"` in `__init__.py` and importing that for `--about`. poetry-dynamic-versioning supports substitution from a `0.0.0` placeholder in any file (including `__init__.py` by default), and this approach is standard in other python packages 🤷‍♂️

kgpayne:
Closing in favour of `__version__ = "<version>"` in `__init__.py` solution:

> I still think we are better off adding `__version__ = "<version>"` in `__init__.py` and importing that for `--about`. poetry-dynamic-versioning supports substitution from a `0.0.0` placeholder in any file (including `__init__.py` by default), and this approach is standard in other python packages 🤷‍♂️

kgpayne:
Related issue:

- #1403

PR #1260, opened 2022-12-08
chore: Annotate `Tap.discover_streams` as `Sequence` instead of `list`

edgarrmondragon:
`List` is invariant, which means `list[DerivedStreamClass]` is not
compatible with `list[Stream]`.

```
error: Incompatible return value type (got "List[CustomStream]", expected "List[Stream]")  [return-value]
 note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
 note: Consider using "Sequence" instead, which is covariant
 note: Perhaps you need a type annotation for "streams"? Suggestion: "List[Stream]"
```

----
:books: Documentation preview :books:: https://meltano-sdk--1260.org.readthedocs.build/en/1260/


aaronsteers (approved):
LGTM!

PR #1261, opened 2022-12-08
wip: 1125-sql-ignore-column-collations

### LINKED ISSUE #1125: bug: column collations should be ignored during SQL type comparisons (SQL Server)
### Singer SDK Version

0.11.1

### Python Version

3.10

### Bug scope

### Operating System

Mac

### Description

SQL Server target: Initial load of a table with a varchar column works nicely, but on second write, the loader fails with a ValueError, after attempting to change column types. In the example, a PG source table with a VARCHAR(64) column:

`ValueError: Unable to merge sql types: VARCHAR(64) COLLATE "SQL_Latin1_General_CP1_CI_AS", VARCHAR(64)`

My interpretation is that the target initially ran a create table statement with a VARCHAR(64) column, SQL Server added the collation as a detail, but when inserting again the target interprets varchar with the added collation as a different column type than VARCHAR(64) and tries to alter it, but fails.

An ugly fix is to explicitly declare the collation on table creation: https://github.com/radbrt/target-mssql/blob/37544ad6306095b66bef9f01d35211f486b9bf77/target_mssql/sinks.py#L262

### Code

_No response_

BuzzCutNorman:
This is an attempt to ignore collation.  I am also trying to make the evaluation and adaption of exiting tables clearer.  The addition of `accept_sql_type` checks if sql server will implicitly convert the types during insert.  The column also need to be large enough for the data will fit in. If the column type is acceptable but the column size isn't large enough `merge_sql_types` will be run and the column's size will be adapted.  

This is a work in progress.  I still need to add the size checks and clean up `merge_sql_types`.

Please take a look and let me know what you think.

fixes #1125 

----
:books: Documentation preview :books:: https://meltano-sdk--1261.org.readthedocs.build/en/1261/


BuzzCutNorman:
Just a heads up that I am planning on closing this PR tomorrow but will hopefully come back with an attempted fix for 1125 soon.

BuzzCutNorman:
hopefully be back with an attempted fix for 1125 soon

PR #1280, opened 2022-12-19
fix: set repo file perms after checkout in Version bump workflow

kgpayne:
----
:books: Documentation preview :books:: https://meltano-sdk--1280.org.readthedocs.build/en/1280/


WillDaSilva (approved):
Some small part of me hopes that this doesn't fix it, because this really shouldn't be necessary. That said, I do expect this will fix it.

PR #1283, opened 2022-12-19
feat: Automatic catalog selection of replication keys

### LINKED ISSUE #1217: [Feature]: Automatic catalog inclusion of replication keys
### Feature scope

### Description

The standard metadata generated for streams with known schemas marks primary keys with `automatic` selection:

https://github.com/meltano/sdk/blob/eab95fca545c55b06f48e3264271197d12b812af/singer_sdk/_singerlib/catalog.py#L194-L200

We should probably do the same about replication keys.

edgarrmondragon:
Closes #1217 

----
:books: Documentation preview :books:: https://meltano-sdk--1283.org.readthedocs.build/en/1283/


visch:
@edgarrmondragon  awesome stuff! Fixes this one for everybody

edgarrmondragon:
@kgpayne can you think of any undesired consequences of doing this? I don't think many folks are relying on the replication key not being selected by default. Wdyt?

kgpayne:
@edgarrmondragon I can't think of any reason not to select replication keys by default 🙂 Users can always override if necessary too, so seems like a good default behaviour to me 👍

kgpayne (approved):

PR #1286, opened 2022-12-20
Properties are required in a schema

### LINKED ISSUE #1145: [Feature]: SQLTarget Invalid Schema should fail nicely
### Feature scope

### Description

https://github.com/MeltanoLabs/target-postgres/blob/main/target_postgres/tests/test_standard_target.py#L70 is the `test_invalid_schema` test. 

We should fail nicely, this test was taken from https://github.com/MeltanoLabs/target-sqlite/blob/master/tests/test_target_sqlite.py#L34-L42 

visch:
Closes #1145 

Example implementation here https://github.com/MeltanoLabs/target-postgres/commit/9c475a140c21d4cd849ba4ede022bc19eb6b2c51 (With test)

----
:books: Documentation preview :books:: https://meltano-sdk--1286.org.readthedocs.build/en/1286/

### DIFF
--- a/singer_sdk/target_base.py
+++ b/singer_sdk/target_base.py
@@ -331,7 +331,7 @@ class Target(PluginBase, SingerReader, metaclass=abc.ABCMeta):
         Args:
             message_dict: The newly received schema message.
         """
-        self._assert_line_requires(message_dict, requires={"stream", "schema"})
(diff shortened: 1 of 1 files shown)


visch:
This PR is wrong

PR #1287, opened 2022-12-20
fix: Targets to fail gracefully when schema message is missing the `properties` key

visch:
Closes https://github.com/meltano/sdk/issues/1145

Example implementation https://github.com/MeltanoLabs/target-postgres/commit/41822b548ef99dcb5aa79ea140e872a6beb7892a

----
:books: Documentation preview :books:: https://meltano-sdk--1287.org.readthedocs.build/en/1287/

### DIFF
--- a/singer_sdk/target_base.py
+++ b/singer_sdk/target_base.py
@@ -332,6 +332,7 @@ class Target(PluginBase, SingerReader, metaclass=abc.ABCMeta):
             message_dict: The newly received schema message.
         """
         self._assert_line_requires(message_dict, requires={"stream", "schema"})
+        self._assert_line_requires(message_dict["schema"], requires={"properties"})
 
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):

singer_sdk/target_base.py
```diff
@@ -332,6 +332,7 @@ def _process_schema_message(self, message_dict: dict) -> None:
             message_dict: The newly received schema message.
         """
         self._assert_line_requires(message_dict, requires={"stream", "schema"})
+        self._assert_line_requires(message_dict["schema"], requires={"properties"})
```
edgarrmondragon, line 335: The exception raised here (`Exception`) is too generic. A more concrete exception class would be nice

PR #1306, opened 2023-01-03
docs: Update code samples docs

pnadolny13:
I was using [the code samples docs](https://sdk.meltano.com/en/latest/code_samples.html#dynamically-discovering-schema-for-a-stream) for remembering how to do dynamic schema discovery and noticed a few bugs that made a copy/paste not work:
- missing a colon in for loop
- the cookiecutter imports `from singer_sdk import typing as th` so all  the properties references needed to be prefixed with `th.` to work

----
:books: Documentation preview :books:: https://meltano-sdk--1306.org.readthedocs.build/en/1306/

### DIFF
--- a/docs/code_samples.md
+++ b/docs/code_samples.md
@@ -41,7 +41,7 @@ class TapCountries(Tap):
     config options and does not require authentication.
     """
     name = "tap-countries"
-    config_jsonschema = PropertiesList([]).to_dict()
+    config_jsonschema = th.PropertiesList([]).to_dict()
 
     def discover_streams(self) -> List[Stream]:
         """Return a list containing the two stream types."""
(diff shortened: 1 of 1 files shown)


WillDaSilva:
> the cookiecutter imports `from singer_sdk import typing as th` so all the properties references needed to be prefixed with `th.` to work

Why do we do that though? I've never seem `typing` imported as `th` - it's somewhat unintuitive.

pnadolny13:
@WillDaSilva good question, I dont know. I guess I assumed it had something to do with multiple modules. Either way though if we switch off using `th` in this example we should add the imports in so users know how to import Properties correctly.

tayloramurphy (approved):

edgarrmondragon (approved):

PR #1308, opened 2023-01-04
fix(taps): Replace hyphens with underscores when generating expected env var name `<PLUGIN_NAME>_LOGLEVEL`

adherr:
Currently, we're looking for `cls.name.upper()`, with hyphens intact. Since hyphens are not allowed in shell variable names, this makes it impossible to set the loglevel for a single tap. This change makes this variable match other plugin specific variables by replacing hyphens with underscores.

Also, fix a formatting glitch in the contributing guide.

----
:books: Documentation preview :books:: https://meltano-sdk--1308.org.readthedocs.build/en/1308/


aaronsteers (approved):
@adherr - Thanks very much for this contribution!

edgarrmondragon:
Thanks @adherr, merging now!

PR #1365, opened 2023-01-30
feat: Add `strptime_to_utc` and `strftime` functions to `_singerlib.utils`

menzenski:
Slack discussion here https://meltano.slack.com/archives/C01PKLU5D1R/p1675106007009959

Context: I am trying to add Python 3.11 support to [tap-tableau](https://github.com/GtheSheep/tap-tableau/pull/4). Currently that tap depends on the sdk version ^0.4.5 — python 3.11 support was added in 0.13.1 so I need to bump the sdk dependency at least that far, but would like to bring it to current 0.19.0 if possible. However, one of the sdk changes made since 0.4.5 was dropping the dependency on singer-python , and this tap depends on that package (transitively via the sdk dependency). It uses that singer package to format a datetime https://github.com/GtheSheep/tap-tableau/blob/main/tap_tableau/utils.py#L6

----
:books: Documentation preview :books:: https://meltano-sdk--1365.org.readthedocs.build/en/1365/


singer_sdk/_singerlib/utils.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    if dtime.utcoffset() != timedelta(0):
+        raise Exception("datetime must be pegged at UTC tzoneinfo")
+
+    try:
+        dt_str = dtime.strftime(format_str)
+        if dt_str.startswith("4Y"):
+            return dtime.strftime(DATETIME_FMT_SAFE)
+    except ValueError:
+        return dtime.strftime(DATETIME_FMT_SAFE)
```
edgarrmondragon, line 47: I think a branch is missing in the logic here:

```suggestion
        dt_str = dtime.strftime(format_str)
            if dt_str.startswith("4Y"):
                dt_str = dtime.strftime(DATETIME_FMT_SAFE)
    except ValueError:
        dt_str = dtime.strftime(DATETIME_FMT_SAFE)
    
    return dt_str
```

menzenski: Good catch, thanks - I've corrected that

edgarrmondragon (approved):
Thank you @menzenski!

Issue #1377, opened 2023-02-01
bug: Consider moving cookiecutter tests to root of generated project to facilitate test discovery


edgarrmondragon:
### Singer SDK Version

0.19.0

### Python Version

NA

### Bug scope

Other

### Operating System

Any

### Description

See https://github.com/meltano/sdk/issues/1370#issuecomment-1412430065

> Aha! Yes, that solved this. I also had to update pytest and, along with that, move the tests folder out into the rootdir of the repo, from `tap-postgres/tap-postgres/tests/` -> `tap-postgres/tests/`. Thank you!!

qbatten:
Specifically, pytest >=4.0 requires you to do this if you're defining pytest_plugins https://docs.pytest.org/en/7.1.x/deprecations.html#pytest-plugins-in-non-top-level-conftest-files. AIUI.

edgarrmondragon:
Thanks for the additional context @qbatten!

menzenski:
I am seeing this issue with SDK version 0.22.1 and `tests/` in the top level repository directory.

Tests directory: https://github.com/menzenski/tap-mongodb/blob/main/tests/core_test.py
Failing tests: https://github.com/menzenski/tap-mongodb/actions/runs/4635115290/jobs/8201896087#step:7:24

edgarrmondragon:
@menzenski were you able to fix this by adding a `tests/conftest.py`?

PR #1380, opened 2023-02-01
fix(templates): Move tests in cookiecutters to project root to support `pytest_plugins`

edgarrmondragon:
Closes https://github.com/meltano/sdk/issues/1377

----
:books: Documentation preview :books:: https://meltano-sdk--1380.org.readthedocs.build/en/1380/


edgarrmondragon:
@kgpayne I would appreciate your 👀 on this PR. This may be a blocker for new taps and targets created from the cookiecutter templates to start using the new testing framework immediately.

kgpayne (approved):

kgpayne:
LGTM @edgarrmondragon 🚢 I did this in `tap-snowflake` but didn't clock that it might be a wider issue 🤦‍♂️ Once this merges, we should update the testing guide. IIRC the pytest error is quite clear, but still useful to document 🙂

PR #1385, opened 2023-02-02
fix: SQL Targets ignore collation when evaluating column data types

### LINKED ISSUE #1125: bug: column collations should be ignored during SQL type comparisons (SQL Server)

BuzzCutNorman:
This is an attempt to fix #1125

----
:books: Documentation preview :books:: https://meltano-sdk--1385.org.readthedocs.build/en/1385/


singer_sdk/connectors/sql.py
```diff
@@ -959,10 +959,18 @@ def _adapt_column_type(
         Raises:
             NotImplementedError: if altering columns is not supported.
         """
+        collation_removed = False
         current_type: sqlalchemy.types.TypeEngine = self._get_column_type(
             full_table_name, column_name
         )
 
+        # remove collation if present and save it
+        if hasattr(current_type, "collation"):
```
edgarrmondragon, line 968: Missing coverage here may point to a useful refactor that can be unit-tested:

Wdyt about moving this block to a helper function that takes a type object and returns the collation, or `None` if the collation is falsy:

```python
def remove_collation(column_type: TypeEngine) -> "Collation" | None:  # don't know the actual type of a collation
    if hasattr(current_type, "collation") and current_type.collation:
        current_type_collation = current_type.collation
        setattr(current_type, "collation", None)
        return current_type_collation
    return None

def update_collation(column_type: TypeEngine, collation: "Collation" | None) -> None:
    """Replace column collation if it has one."""
    if hasattr(column_type, "collation") and collation:
        setattr(column_type, "collation", collation)
```

Later we can check if the collation is not `None` to set it in the new type

```python
        current_type_collation = remove_collation(current_type)

        ...

        update_collation(compatible_sql_type, current_type_collation)
```

edgarrmondragon (changes requested):

singer_sdk/connectors/sql.py
BuzzCutNorman: I think this looks great. I will work on getting these two functions in place and tests for them.

edgarrmondragon: Thanks @BuzzCutNorman!

BuzzCutNorman: @edgarrmondragon when you have a moment, please review the changes I made.

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
...
         )
         assert statement == rendered_statement
+
+    def test_remove_collation_text_type(self, connector):
+        remove_collation = getattr(connector, "remove_collation")
+        test_collation = "SQL_Latin1_General_CP1_CI_AS"
+        current_type = sqlalchemy.types.Text(collation=test_collation)
+        current_type, current_type_collation = remove_collation(current_type)
+        # Check collation was set to None by the function
+        assert current_type.collation == None
```
edgarrmondragon, line 101:
```suggestion
        assert current_type.collation is None
```

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
...
+        # Check that we get the same collation we put in back out
+        assert current_type_collation == test_collation
+
+    def test_remove_collation_non_text_type(self, connector):
+        remove_collation = getattr(connector, "remove_collation")
+        current_type = sqlalchemy.types.Integer()
+        current_type, current_type_collation = remove_collation(current_type)
+        # Check there is not a collation attribute
+        assert hasattr(current_type, "collation") == False
```
edgarrmondragon, line 110:
```suggestion
        assert not hasattr(current_type, "collation")
```

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
...
+        # Check that we get the same type we put in
+        assert str(current_type) == "INTEGER"
+        # Check that this variable is missing
+        assert not current_type_collation
```
edgarrmondragon, line 114:
```suggestion
        assert current_type_collation is None
```

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
...
+        # Check collation was set to the value we put in
+        assert compatible_type.collation == test_collation
+
+    def test_update_collation_non_text_type(self, connector):
+        update_collation = getattr(connector, "update_collation")
+        test_collation = "SQL_Latin1_General_CP1_CI_AS"
+        compatible_type = sqlalchemy.types.Integer()
+        compatible_type = update_collation(compatible_type, test_collation)
+        # Check there is not a collation attribute
+        assert hasattr(compatible_type, "collation") == False
```
edgarrmondragon, line 130:
```suggestion
        assert not hasattr(compatible_type, "collation")
```

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
             )
         )
         assert statement == rendered_statement
+
+    def test_remove_collation_text_type(self, connector):
+        remove_collation = getattr(connector, "remove_collation")
```
edgarrmondragon, line 96:
```suggestion
    def test_remove_collation_text_type(self):
        remove_collation = SQLConnector.remove_collation
```

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
...
+        test_collation = "SQL_Latin1_General_CP1_CI_AS"
+        current_type = sqlalchemy.types.Text(collation=test_collation)
+        current_type, current_type_collation = remove_collation(current_type)
+        # Check collation was set to None by the function
+        assert current_type.collation == None
+        # Check that we get the same collation we put in back out
+        assert current_type_collation == test_collation
+
+    def test_remove_collation_non_text_type(self, connector):
+        remove_collation = getattr(connector, "remove_collation")
```
edgarrmondragon, line 106:
```suggestion
    def test_remove_collation_non_text_type(self):
        remove_collation = SQLConnector.remove_collation
```

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
...
+        current_type, current_type_collation = remove_collation(current_type)
+        # Check there is not a collation attribute
+        assert hasattr(current_type, "collation") == False
+        # Check that we get the same type we put in
+        assert str(current_type) == "INTEGER"
+        # Check that this variable is missing
+        assert not current_type_collation
+
+    def test_update_collation_text_type(self, connector):
+        update_collation = getattr(connector, "update_collation")
```
edgarrmondragon, line 117:
```suggestion
    def test_update_collation_text_type(self):
        update_collation = SQLConnector.update_collation
```

tests/core/test_connector_sql.py
```diff
@@ -91,3 +91,42 @@ def test_get_column_ddl(
...
+        test_collation = "SQL_Latin1_General_CP1_CI_AS"
+        compatible_type = sqlalchemy.types.Text(collation=None)
+        compatible_type = update_collation(compatible_type, test_collation)
+        # Check collation was set to the value we put in
+        assert compatible_type.collation == test_collation
+
+    def test_update_collation_non_text_type(self, connector):
+        update_collation = getattr(connector, "update_collation")
```
edgarrmondragon, line 125:
```suggestion
    def test_update_collation_non_text_type(self):
        update_collation = SQLConnector.update_collation
```

edgarrmondragon (changes requested):
@BuzzCutNorman `remove_collation` and `update_collation` both mutate the type object _in-place_, i.e. you'll notice the returned value points to the same object as the argument:

```python
new_type = remove_collation(current_type)
assert new_type is current_type
```

So, do you think it makes to only return the collation from `remove_collation` and return nothing from `update_collation`. Similar to operations on a dictionary:

```python
>>> d = {1: 2, 2: 3, 3: 4}
>>> d.pop(1)  # Return the removed value
2
>>> d.update(a=0)  # Return nothing, mutate in-place
>>> d
{2: 3, 3: 4, 'a': 0}
```

BuzzCutNorman:
> So, do you think it makes to only return the collation from `remove_collation` and return nothing from `update_collation`. Similar to operations on a dictionary:

@edgarrmondragon, Yes it does make sense and it looks much cleaner.

edgarrmondragon (approved):
Thanks @BuzzCutNorman, this is g2g!

BuzzCutNorman:
Thank you @edgarrmondragon ! Couldn't have done it without your excellent guidance.  I learned a lot today.

PR #1390, opened 2023-02-06
docs: Add incremental replication example

### LINKED ISSUE #1372: [Docs]: Add usage example of `Stream.get_starting_timesamp()`
### Documentation type

How-to guides

### Description

https://meltano.slack.com/archives/C01PKLU5D1R/p1675116342431159?thread_ts=1674850558.848759&cid=C01PKLU5D1R

> I think a bit more documentation in the [Meltano Singer SDK](https://sdk.meltano.com/en/main/index.html) would be a great improvement to the user experience. I think either sometime in the [SDK Code Samples](https://sdk.meltano.com/en/main/code_samples.html#sdk-code-samples) with example of an implementation of [get_starting_timesamp()](https://sdk.meltano.com/en/latest/classes/singer_sdk.Stream.html#singer_sdk.Stream.get_starting_timestamp) or a dedicated section similar to [Parent-Child Streams](https://sdk.meltano.com/en/main/parent_streams.html#parent-child-streams) would be awesome.

flexponsive:
adds a new docs page with an example of incremental replication based on a time field; closes #1372

----
:books: Documentation preview :books:: https://meltano-sdk--1390.org.readthedocs.build/en/1390/


docs/incremental_replication.md
```diff
@@ -0,0 +1,56 @@
...
+        if next_page_token is not None:
+            params["page"] = next_page_token
+        self.logger.info(f"QUERY PARAMS: {params}")
+        return params
+
+    url_base = "https://example.com/wp-json/wp/v2/comments"
+    authenticator = None
+```
+
+First we inform the SDK of the `replication_key`, which automatically triggers incremental import mode. Second, optionally, set `is_sorted` to true; with this setting, Singer will throw an error if a supposedly incremental import sends results older than the starting timestamp.
```
davegot12, line 34: This is still unclear to me how this property works. Could there be a concrete example provided?

docs/incremental_replication.md
```diff
@@ -0,0 +1,56 @@
...
+
+Last, we have to adapt the query to the remote system, in this example by adding a query parameter with the ISO timestamp.
+
+Note:
+- unlike a `primary_key`, a `replication_key` does not have to be unique
+- in incremental replication, it is OK to resend rows where the replication key is equal to previous highest key.
```
davegot12, line 40: I would add here that it is ok to resend rows because the target will just update the row with the same values instead of inserting duplicates. At least that's the behavior of the target [snowflake](https://hub.meltano.com/loaders/target-snowflake/) I'm familiar with.

davegot12:
Thank you @flexponsive for this PR, I was the original poster on Slack for this documentation and you beat me to it :)

I left two comments with suggestion that could add a bit more depth to the docs, feel free to take them into account or not.

docs/incremental_replication.md
davegot12: I just realized my comment is confusing, what is unclear to me is how the `is_sorted` property is used and I think an example might help.

Like what if the property is set to True and some data points come in, what happens exactly in the messages that are sent to the target? And what if the same data points come in, but the property was set to False, what would be different? Something along those lines.

tayloramurphy:
@aaronsteers @edgarrmondragon can you give this a review when you're able to?

Thanks @flexponsive for the PR!

docs/incremental_replication.md
edgarrmondragon, line 40:
```suggestion

```{note}
- The SDK will throw an error if records come out of order when `is_sorted` is true.
- Unlike a `primary_key`, a `replication_key` does not have to be unique
- In incremental replication, it is OK and usually recommended to resend rows where the replication key is equal to previous highest key.
```

docs/incremental_replication.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Manually testing incremental import during development
+
+To test the tap stand-alone, manually create a state file and run the tap:
```
edgarrmondragon, line 44:
```suggestion
To test the tap in standalone mode, manually create a state file and run the tap:
```

edgarrmondragon (changes requested):

docs/incremental_replication.md
```diff
@@ -0,0 +1,56 @@
...
+        self.logger.info(f"QUERY PARAMS: {params}")
+        return params
+
+    url_base = "https://example.com/wp-json/wp/v2/comments"
+    authenticator = None
+```
+
+First we inform the SDK of the `replication_key`, which automatically triggers incremental import mode. Second, optionally, set `is_sorted` to true; with this setting, Singer will throw an error if a supposedly incremental import sends results older than the starting timestamp.
+
+Last, we have to adapt the query to the remote system, in this example by adding a query parameter with the ISO timestamp.
```
edgarrmondragon, line 36:
```suggestion
3. Last, we have to adapt the query to the remote system, in this example by adding a query parameter with the ISO timestamp.
```

edgarrmondragon (approved):
Thank you @flexponsive for the PR, and @davegot12 for your suggestions!

flexponsive:
Thank you for the feedback and improvements!

Issue #1393, opened 2023-02-06
refactor: SQLConnector connection lifecycle handling


qbatten:
### Feature scope

Other

### Description

This is a sub-issue broken out of #1352.

[This comment explains this sub-issue in detail](https://github.com/meltano/sdk/issues/1352#issuecomment-1405703394). 

The idea here is to make the SQLConnector correctly handle it's database connection, and take on sole ownership of that job. In other words:
- The connection itself is protected
- Streams & other classes/functions/etc are not allowed to open/close/access a connection (subclasses of SQLConnector are still allowed & expected to do so).
- The SQLConnector itself opens and closes connections correctly and doesn't leave connections hanging open

Some more detailed next steps here are:
1. Make a SQLConnector (protected) method, `_connect` that can be used as a context manager, and use that every single time a connection is opened. (An alternative is just to make sure that any connection opened gets closed as soon as it's done, but I think that's not as nice & is more prone to error.)
2. Only the SQLConnector can use the connection. No other objects should access or be aware of the SQLConnector's connection; and anything that other objects ask the SQLConnector to do that involve interaction w the database should be accessible via a SQLConnector method that correctly opens and closes the connection for that db interaction.

To fully realize this goal, we need to make breaking changes. In particular:
- Stop accepting a `_connection` arg on instantiation
- Stop storing and accessing `_connection` altogether
- Remove `connection` and `create_sqlalchemy_connection`. It's a public method that returns a connection. That encourages bad behavior.
- Remove `create_sqlalchemy_engine`. There shouldn't be a public method for creating or getting a sqlalchemy engine; also there should just be one engine per SQLConnector.

qbatten:
There is a mostly-not-breaking way to do this, and I'm going to push a draft PR of that soon. It does...sorta break things, but only for subclasses.
 
The breaking change is as follows: we currently allow someone to pass `_connection`. In order to move towards the goal of this issue, that must be removed, it's just incompatible with this change.

tayloramurphy:
Thanks for the issue / PR! I'll have the engineering team take a look at the PR>

BuzzCutNorman:
If you remove `create_sqlalchemy_engine` would there, be a new way for developers, be to set engine arguments?  I have overloaded it to allow some user configuration options in this [portion](https://github.com/BuzzCutNorman/tap-mssql/blob/46a1663095791914163dd5e58dbf10920de0ba0a/tap_mssql/client.py#L69-L85) of my tap-mssql.

[...]

PR #1394, opened 2023-02-06
feat: refactor SQLConnector connection handling

qbatten:
This is a draft PR that proposes a solution to issue #1393.

Broad overview:
- We are no longer storing an open connection on this object. Instead, we store an engine. Wherever we were previously making a new engine, we instead access the connector's existing engine. When a connection is needed, we open one at that time.
- When connections are opened, they're opened using `self._connect()` as a context manager. This rule is put into place where it's possible to do so without making a breaking change.
- Several methods are prepared for deprecation, but their function is preserved. 
    - (Breaking change) `_connection` is removed altogether. The tap no longer accepts that argument on init, and we don't store a single connection at all
        - This is a breaking change for any subclasses that were using this property
        - This is a breaking change for any subclasses that counted on a single connection being reused throughout the lifetime of the instance
        - This is a breaking change for any cases where an external method call counted on a single connection being returned each time that `create_sqlalchemy_connection` was called (or for anyone who didn't respect the protectedness of _connection).
    - (Slightly breaking) `connection` still exists, it just calls `create_sqlalchemy_connection`. So this is a slightly breaking change—- it returns a new connection instead of the _connection that prviously had been preserved on the instance.
    - `create_sqlalchemy_engine` now just returns self._engine, since we have an engine on this object anyway.
    - `create_sqlalchemy_connection` now opens a connection using self._engine and returns the open connection.
    
    
Detailed changes in functionality:
1. `self._connect()` is added. This is the correct way to open connections now—- you use it as a context manager, exactly as you would sqlalchemy.engine.connect()... in fact currently it just directly wraps that.
2. We now create and cache a single engine for this instance. It's accessed as a protected property under `_engine`.
2. `create_sqlalchemy_connection` calls `self._engine` now instead of `self.create_sqlalchemy_engine`, since the connector now has an engine. This and the next item help extricate and disconnect these soon-to-be-deprecated methods.
3. `create_sqlalchemy_engine` now calls `self._engine` instead of making the engine itself.
4. Everywhere else in the SQLConnector that connection.execute was called now uses the correct/new _connect strategy.
5. Internal calls to create_sqlalchemy_engine now just call _engine (this is just a change to discover_catalog_entries).

### DIFF
--- a/singer_sdk/connectors/sql.py
+++ b/singer_sdk/connectors/sql.py
@@ -2,10 +2,12 @@
 
 from __future__ import annotations
 
+import functools
 import logging
+from contextlib import contextmanager
 from datetime import datetime
 from functools import lru_cache
-from typing import Any, Iterable, cast
+from typing import Any, Iterable, Iterator, cast
 
 import sqlalchemy
 from sqlalchemy.engine import Engine
@@ -46,7 +48,6 @@ class SQLConnector:
         """
         self._config: dict[str, Any] = config or {}
         self._sqlalchemy_url: str | None = sqlalchemy_url or None
-        self._connection: sqlalchemy.engine.Connection | None = None
 
     @property
     def config(self) -> dict:
@@ -66,8 +67,17 @@ class SQLConnector:
         """
         return logging.getLogger("sqlconnector")
 
+    @contextmanager
+    def _connect(self) -> Iterator[None]:
+        with self._engine.connect().execution_options(stream_results=True) as conn:
+            yield conn
+
     def create_sqlalchemy_connection(self) -> sqlalchemy.engine.Connection:
-        """Return a new SQLAlchemy connection using the provided config.
+        """(DEPRECATED) Return a new SQLAlchemy connection using the provided config.
+
+        Do not use the SQLConnector's connection directly. Instead, if you need
+        to execute something that isn't available on the connector currently,
+        make a child class and add a method on that connector.
 
         By default this will create using the sqlalchemy `stream_results=True` option
         described here:
@@ -81,14 +91,10 @@ class SQLConnector:
         Returns:
             A newly created SQLAlchemy engine object.
         """
-        return (
-            self.create_sqlalchemy_engine()
-            .connect()
-            .execution_options(stream_results=True)
-        )
+        return self._engine.connect().execution_options(stream_results=True)
 
     def create_sqlalchemy_engine(self) -> sqlalchemy.engine.Engine:
-        """Return a new SQLAlchemy engine using the provided config.
+        """(DEPRECATED) Return a new SQLAlchemy engine using the provided config.
 
         Developers can generally override just one of the following:
         `sqlalchemy_engine`, sqlalchemy_url`.
@@ -96,19 +102,20 @@ class SQLConnector:
         Returns:
             A newly created SQLAlchemy engine object.
         """
-        return sqlalchemy.create_engine(self.sqlalchemy_url, echo=False)
+        return self._engine
 
     @property
     def connection(self) -> sqlalchemy.engine.Connection:
-        """Return or set the SQLAlchemy connection object.
+        """(DEPRECATED) Return or set the SQLAlchemy connection object.
+
+        Do not use the SQLConnector's connection directly. Instead, if you need
+        to execute something that isn't available on the connector currently,
+        make a child class and add a method on that connector.
 
         Returns:
             The active SQLAlchemy connection object.
         """
-        if not self._connection:
-            self._connection = self.create_sqlalchemy_connection()
-
-        return self._connection
+        return self.create_sqlalchemy_connection()
 
     @property
     def sqlalchemy_url(self) -> str:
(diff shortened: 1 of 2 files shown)


qbatten:
I used an lru_cache decorator at first, but there doesn't seem to be any real way to handle typing for a lru_cache-wrapped function that returns something— [see here](https://github.com/python/mypy/issues/5107). So instead I just used another property to cache it. Maybe not the most pythonic(?) but it works fine. Another option is to `type: ignore` that line. I haven't worked much with type checkers, so I'm not sure what the right move is here. Ofc open to suggestions if anyone has an opinion.

kgpayne:
@qbatten this is great, thanks for all your work on connection handling 🙏 As we are pre-1.0 for the SDK, and especially since there are far fewer users of the SQLConnector at this point, I think the breaking changes are OK. However, it would be good to proactively reach out to the Taps/Targets we know about that would potentially be impacted (using Hub data), to make sure they are pinning their SDK version and are aware of the changes 🙂 I can help with that 👍

singer_sdk/connectors/sql.py
```diff
@@ -3,9 +3,10 @@
 from __future__ import annotations
 
 import logging
+from contextlib import contextmanager
```
kgpayne, line 6:
```suggestion
import warnings
from contextlib import contextmanager
```

singer_sdk/connectors/sql.py
```diff
@@ -81,34 +91,31 @@ def create_sqlalchemy_connection(self) -> sqlalchemy.engine.Connection:
...
+        """(DEPRECATED) Return a new SQLAlchemy engine using the provided config.
 
         Developers can generally override just one of the following:
         `sqlalchemy_engine`, sqlalchemy_url`.
 
         Returns:
             A newly created SQLAlchemy engine object.
         """
-        return sqlalchemy.create_engine(self.sqlalchemy_url, echo=False)
+        return self._engine
```
kgpayne, line 105:
```suggestion
        warnings.warn(
            "`create_sqlalchemy_engine` is deprecated. Override `sqlalchemy_engine` or sqlalchemy_url` instead.",
            DeprecationWarning
        )
        return self._engine
```

singer_sdk/connectors/sql.py
```diff
@@ -81,34 +91,31 @@ def create_sqlalchemy_connection(self) -> sqlalchemy.engine.Connection:
...
+        make a child class and add a method on that connector.
 
         Returns:
             The active SQLAlchemy connection object.
         """
-        if not self._connection:
-            self._connection = self.create_sqlalchemy_connection()
-
-        return self._connection
+        return self.create_sqlalchemy_connection()
```
kgpayne, line 118:
```suggestion
        warnings.warn(
            "`connection` is deprecated. If you need to execute something "
            "that isn't available on the connector currently, make a child "
            "class and add your required method on that connector.",
            DeprecationWarning
        )
        return self.create_sqlalchemy_connection()
```

singer_sdk/connectors/sql.py
```diff
@@ -81,34 +91,31 @@ def create_sqlalchemy_connection(self) -> sqlalchemy.engine.Connection:
         Returns:
             A newly created SQLAlchemy engine object.
         """
-        return (
-            self.create_sqlalchemy_engine()
-            .connect()
-            .execution_options(stream_results=True)
-        )
+        return self._engine.connect().execution_options(stream_results=True)
```
kgpayne, line 94:
```suggestion
        warnings.warn(
            "`create_sqlalchemy_connection` is deprecated. "
            "If you need to execute something that isn't available "
            "on the connector currently, make a child class and "
            "add your required method on that connector.",
            DeprecationWarning
        )
        return self._engine.connect().execution_options(stream_results=True)
```

kgpayne (changes requested):
LGTM 🚢 Minor suggestions around deprecation warnings 👍

singer_sdk/connectors/sql.py
qbatten: Yeah that's much better than a warning in the docstring haha. Thanks!

edgarrmondragon (changes requested):

qbatten:
Test coverage! Woohoo.

tests/core/test_connector_sql.py
```diff
@@ -14,7 +14,7 @@ class TestConnectorSQL:
 
     @pytest.fixture()
     def connector(self):
-        return SQLConnector()
+        return SQLConnector(config={"sqlalchemy_url": "sqlite:///tmp.db"})
```
edgarrmondragon, line 20: Can an in-memory db be used here?

```suggestion
        return SQLConnector(config={"sqlalchemy_url": "sqlite://"})
```

tests/core/test_connector_sql.py
```diff
@@ -130,3 +130,14 @@ def test_update_collation_non_text_type(self):
...
+    def test_create_engine_returns_new_engine(self, connector):
+        engine1 = connector.create_engine()
+        engine2 = connector.create_engine()
+        assert engine1 is not engine2
+
+    def test_engine_creates_and_returns_cached_engine(self, connector):
+        assert not connector._cached_engine
+        engine1 = connector._engine
+        engine2 = connector._cached_engine
+        assert engine1 is engine2
```
edgarrmondragon, line 146: 👍

edgarrmondragon:
@qbatten this is looking great 😄

tests/core/test_connector_sql.py
qbatten: Oh oops yes! One other small change to make downstream of that, will push it in a sec

edgarrmondragon (approved):
Thanks @qbatten, this is g2g!

@kgpayne I'll leave this open in case you want to review, otherwise I'll merge first thing on Monday.

kgpayne (approved):

Issue #1349, opened 2023-01-25
bug: default implementations of RESTStream (and others) do not adhere to `_MAX_RECORDS_LIMIT`


kgpayne:
### Singer SDK Version

0.18.0

### Python Version

3.10

### Bug scope

Taps (catalog, state, stream maps, etc.)

### Operating System

MacOS

### Description

In testing [MeltanoLabs/tap-stackexchange](https://github.com/MeltanoLabs/tap-stackexchange) (PR [here]()) with the upcoming test improvements, we noticed that the `RESTStream` implementation included in the SDK does not adhere to the `_MAX_RECORDS_LIMIT` internal attribute when returning records from `get_records()`. Specifically, `get_records()` returns _more than_ the limit, causing the `_check_max_record_limit()` to fail and throw a `MaxRecordsLimitException`.

For connection testing (currently the only use of `_MAX_RECORDS_LIMIT`) this error is caught and passed. However, in the new testing framework, we need the stream to complete gracefully (emitting state etc.) so that we can test the records returned. Catching the raised error, whilst possible, is still unhelpful when records/messages are required for testing.

```python
    # Connection test:
    @final
    def run_connection_test(self) -> bool:
        """Run connection test.

        Returns:
            True if the test succeeded.
        """
        for stream in self.streams.values():
            # Initialize streams' record limits before beginning the sync test.
            stream._MAX_RECORDS_LIMIT = 1

        for stream in self.streams.values():
            if stream.parent_stream_type:
                self.logger.debug(
                    f"Child stream '{type(stream).__name__}' should be called by "
                    f"parent stream '{stream.parent_stream_type.__name__}'. "
                    "Skipping direct invocation."
                )
                continue
            try:
                stream.sync()
            except MaxRecordsLimitException:
                pass
        return True
```

In the recent `SQLStream` implementation we added a snippet to "push down" the limit to the remote sql engine, to avoid materialising a complete dataset on the server but exiting early after a small number of records:

```python
    # Get records from stream
    def get_records(self, context: dict | None) -> Iterable[dict[str, Any]]:
        """Return a generator of record-type dictionary objects.

        If the stream has a replication_key value defined, records will be sorted by the
        incremental key. If the stream also has an available starting bookmark, the
        records will be filtered for values greater than or equal to the bookmark value.

        Args:
            context: If partition context is provided, will read specifically from this
                data slice.

        Yields:
            One dict per record.

        Raises:
            NotImplementedError: If partition is passed in context and the stream does
[...]

Issue #1366, opened 2023-01-30
Consider exposing config like `dry_run_record_limit` 


aaronsteers:
We could theoretically expose a record limit in config for SDK-based taps. This would basically be a more advanced implementation of `--test` CLI flag.

However, it is important to properly disclaim the following caveats for any users that opt into this behavior:

1. This mode will very likely break referential integrity expectations.
   - The 'first 100' records from the `orders` table, for instance, might not match all - or even any - of the customer IDs from the first 100 records of the customers table.
1. This mode will break incremental sync expectations for many taps.
   - There's no guarantee that this will align with paging requirements of the upstream source, and for many taps, you would not be able to simply "run it again" to get the _next_ 100 records per stream.
1. The total number of records sent for child streams will be indeterminant, and could be unexpectedly more or less than the max.
   - If the first `n` parents don't have any children, then _no records for the child stream_ would be synced. Example: I'm syncing with max count of `10`. My parent stream is `Issues` and my child stream is `Issue comments`. If my first 10 issues do not have comments, then exactly zero `Issue comments` records will be synced, even though the max for that stream type was never met.
   - Conversely, if the 10 parent streams each have exactly 5 child records (for instance), then the total records streams for the child stream will be 50, since each pass through the child streams would reset the counter. The total records synced could be therefor be much more than the max.
   - Both of these behaviors are tolerable if communicated clearly to users. Most example cases would not mind 50 records instead of 10 for child streams. And conversely, if child items are sparse, the user could increase the limit to something large enough to ensure at least one child item is sent.

For these reasons, I suggest we pay special attention to naming. I propose above a `dry_run_` prefix but there are other ways this could be communicated as well.

A similar approach would be something like `dry_run: true` to enable a developer-specified default (or the SDK default of something like 1000), possibly also allowing an integer instead of the boolean value `dry_run: 10` - with docs clearly explaining that `true` uses the built-in defaults and any other `n` will override the limit at the users' request.

## Alternative using the CLI

An alternative using just the CLI would be to accept this integer as an optional input to `--test`. So, `--test` today tries to sync zero or one records per stream, where-as if `--test=N` were specified, we'd try to sync (at least) `N` records per stream.

cc @kgpayne, @edgarrmondragon 

Related:

- https://github.com/meltano/sdk/issues/1349

pnadolny13:
[...]

PR #1399, opened 2023-02-07
feat: Set `Stream._MAX_RECORDS_LIMIT ` during tap testing

kgpayne:
This currently only works with `SQLStream` instances, which apply the `max_records_limit` in `get_records()` by default. For other types of `Stream`, setting a `max_records_limit` in `SuiteConfig` will cause the tests to fail. See #1349 for more details.

----
:books: Documentation preview :books:: https://meltano-sdk--1399.org.readthedocs.build/en/1399/


kgpayne:
Testing in [tap-snowflake#15](https://github.com/MeltanoLabs/tap-snowflake/pull/15). Reduced total test run time from ~20m to ~2m 👏

kgpayne:
More testing in [MeltanoLabs/tap-stackexchange#196](https://github.com/MeltanoLabs/tap-stackexchange/pull/196). Reduced test run time from ~8m to >1m ⏰

This is accomplished by adding an early exit to `get_records()` on the `RESTStream` subclass directly in the Tap, similar to the one in the SDK implementation of `SQLStream`. Applying this change to `RESTStream` in the SDK would resolve [#1349](https://github.com/meltano/sdk/issues/1349).

```python
# overridden to test _MAX_RECORDS_LIMIT
def get_records(self, context: dict | None) -> Iterable[dict[str, Any]]:
    """Return a generator of record-type dictionary objects.

    Each record emitted should be a dictionary of property names to their values.

    Args:
        context: Stream partition or context dictionary.

    Yields:
        One item per (possibly processed) record in the API.
    """
    index = 1
    for record in self.request_records(context):
        transformed_record = self.post_process(record, context)
        if transformed_record is None:
            # Record filtered out during post_process()
            continue

        if self._MAX_RECORDS_LIMIT and (index + 1) == self._MAX_RECORDS_LIMIT:
            # if we have reached the max records limit, return early
            break

        yield transformed_record
        index += 1
```

pyproject.toml
```diff
@@ -175,6 +175,7 @@ omit = [
     "tests/*",
     "samples/*",
     "singer_sdk/helpers/_compat.py",
+    "singer_sdk/testing/*",
```
edgarrmondragon, line 178: Why ignore these for coverage?

edgarrmondragon (changes requested):
I may be missing something but I don't see `get_records` using the limit or handling the exception.

kgpayne:
@edgarrmondragon 

> I may be missing something but I don't see `get_records` using the limit or handling the exception.

That change isn't in this PR 😅 All I am doing here is setting `_MAX_RECORDS_LIMIT` on the streams after the Tap is instantiated and before testing begins. `SQLStream.get_records()` already reads `_MAX_RECORDS_LIMIT` and respects the limit, and there is an open PR in `tap-stackexchange` to test the same approach in `RESTStream.get_records()` (snippet [here](https://github.com/MeltanoLabs/tap-stackexchange/pull/196/files#diff-6ea50d44bd3c3d25fe6558f96d694fbdd50be1ac22d2a4ca6ac3c85a2e0e41a2R158-R183)).

I didn't want to conflate too many similar but separate changes in one PR, and `SQLStream`s already [stand to gain](https://github.com/MeltanoLabs/tap-snowflake/pull/15#issuecomment-1420962939) from this even if we never make the `RESTStream` change 🙂

pyproject.toml
kgpayne: I didn't think it made sense to cover them as I can't see a way to easily test the test framework (as it relies on the pytest plugin) 🤔 I suppose we can test the individual tests, so I can roll back the ignore and create an issue for "testing the test cases". WDYT?

edgarrmondragon: Yeah, I think that's what we want eventually to be safe

edgarrmondragon:
> @edgarrmondragon
> 
> > I may be missing something but I don't see `get_records` using the limit or handling the exception.
> 
> That change isn't in this PR 😅 All I am doing here is setting `_MAX_RECORDS_LIMIT` on the streams after the Tap is instantiated and before testing begins. `SQLStream.get_records()` already reads `_MAX_RECORDS_LIMIT` and respects the limit, and there is an open PR in `tap-stackexchange` to test the same approach in `RESTStream.get_records()` (snippet [here](https://github.com/MeltanoLabs/tap-stackexchange/pull/196/files#diff-6ea50d44bd3c3d25fe6558f96d694fbdd50be1ac22d2a4ca6ac3c85a2e0e41a2R158-R183)).
> 
> I didn't want to conflate too many similar but separate changes in one PR, and `SQLStream`s already [stand to gain](https://github.com/MeltanoLabs/tap-snowflake/pull/15#issuecomment-1420962939) from this even if we never make the `RESTStream` change 🙂

Oh, ok. I've changed the title of the PR to better reflect that. Is there an issue to make `SuiteConfig.max_records_limit` apply universally?

aaronsteers (changes requested):
Requesting hold to allow time to review.

kgpayne:
@aaronsteers nudge 🙏 Would be good to get this out before it goes stale 🚢

pyproject.toml
kgpayne: Done ✅

aaronsteers:
I spent more time reviewing this, and I don't think we should merge in this current state.

Specifically, this doesn't seem to fit requirements of a [stable increment](https://handbook.meltano.com/engineering/stable-increments).

I'd suggest 2 changes:

1. The test function should be calling `run_connection_test()` which handles the error `MaxRecordsLimitException`.
2. We should consider renaming `_MAX_RECORDS_LIMIT` to `_DRY_RUN_RECORD_LIMIT` - or any other variable name which clearly informs developers (SDK contributors as well as tap developers) that this is _not_ a viable lever for forcing sync via small batches.

As I explain in #1349 and #1366, aborting the stream without reaching the end has to generate an error, and that error needs to be caught. Otherwise, state messages for the stream are not finalized - and cannot be finalized - 

It's important to note that developers may run these codepaths without a detailed understanding of implications - and we need to be careful that all options presented by the SDK still promise valid taps that comply with the Singer Spec. Exposing a records limit variable will _break_ compliance with the Singer Spec whenever that max limit is reached. 

There's a path forward here if we want to add this to a "partial success" return code, but the partial success return code needs to distinguish interruptions between these two categories: (1) streams that validly incremented state but had more records remaining, and (2) streams which did _not_ successfully reach a bookmark and therefor represent the start of an infinite loop, where no progress will ever be made towards syncing the stream successfully.

aaronsteers:
I've addressed my own feedback above into this PR: #1436.

That PR combined with this one make the increment "stable", specifically in that we deal with the exceptions that are raised as a result of a max record count limit - including proper handling of streams according to the `replication_method` and whether or not the stream state is resumable.

Related, this proposes non-zero exit code options for taps "aborted" which were able to be successfully paused and left in a valid state despite leaving records in the source:

- #1409 

We don't need to tackle non-zero return codes as of now; this PR and #1436 both ensure that:

1. Test functions properly leverage max record abort thresholds.
2. Non-test functions will properly raise exceptions if record limit is misapplied by developers during "real" sync operations.

PR #1353, opened 2023-01-26
docs: Reorganize docs and add deprecation timeline

edgarrmondragon:

PR #1407, opened 2023-02-08
docs: Docs nav structure reorg follow-up

aaronsteers:
I did a pass to clean up titles and organization a little bit, following from https://github.com/meltano/sdk/pull/1353#issuecomment-1414272456.

Preview:

- https://meltano-sdk--1407.org.readthedocs.build/en/1407/

Notable changes:

Removed "SDK Implementation Details - " as a prefix from pages in this section:

image

Moved "Plugin Capabilities" to Reference:

image

Combined "Contributing" page and "Release Process" under common parent. Raised "Deprecation Timeline" up one level to be visible to non-contributors/non-maintainers.

image


docs/CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
-# Contributing to the SDK
+# Contributor's Guide
```
WillDaSilva, line 1:
```suggestion
# Contributing Guide
```

docs/development/release_process.md
```diff
@@ -2,7 +2,7 @@
 
 ## PyPI releases
 
-TODO: This section needs to be updated to reflect the release process.
+Releases are published to PyPi by a GitHub Actions workflow, which in turn is triggered by a GitHub [Release](https://github.com/meltano/sdk/releases) being published.
```
WillDaSilva, line 5:
```suggestion
Releases are published to PyPI by a GitHub Actions workflow, triggered when a GitHub [Release](https://github.com/meltano/sdk/releases) is published.
```

docs/index.rst
```diff
@@ -68,16 +69,15 @@ within the `#singer-tap-development`_ and `#singer-target-development`_ Slack ch
...
-   implementation/index
    testing
 
 .. toctree::
-   :caption: Development
+   :caption: SDK Development
    :maxdepth: 1
    :hidden:
 
-   CONTRIBUTING
```
WillDaSilva, line 79: Why remove this?

aaronsteers: Moved under "development" parent.

image

If we raise up, we should raise up "Release Process" also; otherwise the "Development" parent would only have a single parent.

aaronsteers: Optionally, we could just have "Contributor's Guide", "Release Process", and "Deprecation Timeline" as top-level items under the "SDK Development" section.

@edgarrmondragon - any objections to us moving them up to the top level?

aaronsteers: As noted in the description, I do think "Deprecation Timeline" should be visible to non-contributors, since that effects tap/target maintainers even if they are not themselves contributors.

WillDaSilva: Oh I see where it is now. I was just failing to see the new location.

I have a slight preference for moving it to the top-level under "SDK Development". There's only 2 items nested there after all, and the name they're nested under is identical to the category name above.

WillDaSilva (approved):
Looks good to me. Moving those 2 items out from "SDK Development" would be nice, but this PR is still good with or without that.

docs/index.rst
edgarrmondragon: > any objections to us moving these up to the top level?

@aaronsteers no objections, it makes sense :+1:

edgarrmondragon (approved):
This LGTM, regardless if those two pages are moved up.

aaronsteers:
I noticed we removed the dedicated `typing` page when consolidating into `classes`. Because there was helpful docstring at the top of that module, and because I like having a clean discoverable top-level nav for this, I've reverted the removal of the other page. I added `:noindex:` to the restored page so that the new pages are still the authority for class lookups.

Issue #1224, opened 2022-11-28
[Bug]: Cookiecutter doesn't pass mypy for post_process method signature


visch:
### Singer SDK Version

0.14.0

### Python Version

3.8

### Bug scope

Taps (catalog, state, stream maps, etc.)

### Operating System

Ubuntu

### Description

```
lint run-test: commands[5] | poetry run mypy tap_messagebird --exclude=tap_messagebird/tests        
tap_messagebird/client.py:77: error: Signature of "post_process" incompatible with supertype "Stream"    
```

This is generated from https://github.com/meltano/sdk/blob/main/cookiecutter/tap-template/%7B%7Bcookiecutter.tap_id%7D%7D/%7B%7Bcookiecutter.library_name%7D%7D/%7B%25if%20'REST'%20%3D%3D%20cookiecutter.stream_type%20%25%7Dclient.py%7B%25endif%25%7D#L141

To fix change the signature to 

```
def post_process(self, row: dict, context: dict | None = None) -> dict | None:
```
And import annotations from future (if using a python version without those types)
```
from __future__ import annotations
```

Bump to https://github.com/meltano/sdk/issues/993

edgarrmondragon:
Closed by https://github.com/meltano/sdk/pull/1648

Issue #993, opened 2022-09-26
Feat: Build (empty?) cookiecutters and run with Meltano to be sure they work


visch:
### Feature scope

Taps (catalog, state, stream maps, etc.)

### Description

Making sure cookiecutters continue to work is hard! It'd be great to build a cookiecutter project as a part of CI with the SDK to be sure the latest version of the cookiecutter with the latest version of meltano work!

edgarrmondragon:
Closed by https://github.com/meltano/sdk/pull/1410

Issue #360, opened 2022-04-13
target cookiecutter has no tox.ini


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/362

Originally created by @vischous on 2022-04-13 20:02:34

---

target cookiecutter has no tox.ini

MeltyBot:
[View 3 previous comments from the original issue on GitLab](https://gitlab.com/meltano/sdk/-/issues/362)

edgarrmondragon:
Resolved by https://github.com/meltano/sdk/pull/1438

PR #1410, opened 2023-02-10
feat: Build empty cookiecutters and run lint task during CI

flexponsive:
Related to #993

# Feature Scope

Adds a Github Action to run basic end-to-end tests for the cookiecutters.

It builds empty cookiecutters from [replay files](https://cookiecutter.readthedocs.io/en/stable/advanced/replay.html), install its dependencies with poetry and runs some auto fixes on the generated code (flake8 & co). Finally, it runs the lint task *as specified by the cookiecutter* against itself. Additionally, a script which creates and lints in one step for developer convenience while working on cookiecutter templates.

This PR also includes a small shell script to build, install and lint a cookiecutter with a single command. 

This PR also includes a few fixes (method signatures #1224, docstrings etc.) demanded by the linter for the case of a tap with REST stream type and API Key authentication method.

# Relationship to `test_cookiecutter.py`

The SDK test suite already includes tests for building cookiecutters in `test_cookiecutter.py`, which runs flake8 and mypy. However, it does not use the same configuration for these tools as the template contains and also does not install dependencies with poetry so some errors are not picked up (e.g. method signatures mismatch). End-to-end tests can be a useful addition here.

# Suggested GIthub Action config

Run the `cookiecutter-e2e` action only when cookiecutter files are changed; optionally on a regular schedule.

----
:books: Documentation preview :books:: https://meltano-sdk--1410.org.readthedocs.build/en/1410/


tayloramurphy:
Thanks for the PR @flexponsive - looks like we've got some engineers assigned as reviewers who will get back to you likely sometime next week!

edgarrmondragon:
Hey @flexponsive! Looks like flake8 is complaining. Let me know when this is ready to be reviewed 🙂

flexponsive:
@edgarrmondragon I have updated the PR and description. Github does not show the results from the E2E-Tests of cookiecutters here in the PR, they are available here:
- [tap, rest, API key](https://github.com/flexponsive/meltano-sdk/actions/runs/4156469784/jobs/7190280635)
- [target, per record](https://github.com/flexponsive/meltano-sdk/actions/runs/4156469784/jobs/7190280688) - does not actually lint due to #360

aaronsteers:
@flexponsive - First of all - _thank you_ for creating this PR! 

What do you think of adding a step to upload the generated projects using GitHub artifacts, if not already doing so?

https://github.com/actions/upload-artifact

```yml
- uses: actions/upload-artifact@v3
  with:
    name: my-artifact
    path: path/to/artifact/world.txt
```

.github/workflows/cookiecutter-e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+    - name: Check out the repository
+      uses: actions/checkout@v3.3.0
+
+    - name: Setup Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4.5.0
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: x64
+        cache: 'pip'
+        cache-dependency-path: 'poetry.lock'
```
edgarrmondragon, line 36: Probably doesn't make sense to use the package's `poetry.lock` as cache key for the templates, e.g. the templates cache could need to refreshed even if `poetry.lock` hasn`t changed.
```suggestion
```

.github/workflows/cookiecutter-e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Upgrade pip
+      env:
+        PIP_CONSTRAINT: .github/workflows/constraints.txt
+      run: |
+        pip install pip
+        pip --version
+
```
edgarrmondragon, line 44: Extra line break

```suggestion
```

e2e-tests/cookiecutters/test_cookiecutter.sh
```diff
@@ -0,0 +1,52 @@
...
+
+CC_TEST_OUTPUT=$CC_BUILD_PATH/$CC_OUTPUT_DIR
+if [[ -d "$CC_TEST_OUTPUT" ]]; then
+    rm -fr "$CC_TEST_OUTPUT"
+fi
+#echo cookiecutter --replay-file $REPLAY_FILE $TAP_TEMPLATE -o $CC_BUILD_PATH; exit
+cookiecutter --replay-file $REPLAY_FILE $TAP_TEMPLATE -o $CC_BUILD_PATH &&
+    cd $CC_TEST_OUTPUT &&
+    pwd &&
+    poetry install
```
edgarrmondragon, line 35: Is this maybe missing a `poetry lock` step?

edgarrmondragon (changes requested):
Hey @flexponsive. A few comments and suggestions. The bash script is good for an initial iteration of this but we should probably (not necessarily) in this PR port it to Nox since that's where we centralize dev tasks.

flexponsive:
Thank you for the fast feedback and improvements to the PR!

@edgarrmondragon I accepted the suggestions and believe the issues you raised are fixed. I agree that it would be a good idea in the future to migrate from the bash script to tox but probably outside the scope of this PR.

@aaronsteers build artifact upload has been added too

Sample run: https://github.com/flexponsive/meltano-sdk/actions/runs/4172051621

aaronsteers:
> build artifact upload has been added too

Nice! Thank you!

edgarrmondragon:
Reviewing now...

edgarrmondragon (approved):

edgarrmondragon (approved):

edgarrmondragon:
Thanks @flexponsive!

Issue #1242, opened 2022-12-05
[Feature]: Lint code with Ruff


edgarrmondragon:
### Feature scope

Other.

### Description

- Extremely fast (~400x time faster than flake8). See https://github.com/charliermarsh/ruff#benchmarks
- No dependency conflicts (that's been a minor pain with flake8)
- Very actively developed
- Can be configured in `pyproject.toml`

https://github.com/charliermarsh/ruff

PR #1417, opened 2023-02-14
chore: Use modern-style type annotations

edgarrmondragon:
Blocked by https://github.com/meltano/sdk/pull/1410 to add them to the cookiecutter templates.

----
:books: Documentation preview :books:: https://meltano-sdk--1417.org.readthedocs.build/en/1417/

### DIFF
--- a/samples/sample_mapper/mapper.py
+++ b/samples/sample_mapper/mapper.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from pathlib import PurePath
-from typing import Generator, List, Optional, Union
+from typing import Generator
 
 import singer_sdk._singerlib as singer
 import singer_sdk.typing as th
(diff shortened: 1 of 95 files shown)


samples/sample_mapper/mapper.py
```diff
@@ -1,5 +1,7 @@
 """A sample inline mapper app."""
 
+from __future__ import annotations
+
 from pathlib import PurePath
 from typing import Generator, List, Optional, Union
```
WillDaSilva, line 6: Looks like some unused imports are left over. I'm not sure why flake8 isn't failing because of this.

edgarrmondragon: Good catch!

We are not using Pyflakes, so we don't get the `F401` error code anywhere. Probably worth including when we do #1242, so we even get pre-commit autofixes.

I've removed the lingering imports in `samples/` for now.

PR #1420, opened 2023-02-14
fix: use separate runner for discovery and tests

kgpayne:


edgarrmondragon:
Still relevant, I think

PR #1434, opened 2023-02-20
fix(templates): Tap template: fix style and docstrings, and add test cases for SQL and "Other" sources

flexponsive:
Follow-up to #1410: Now all tap templates are covered by e2e tests

- Add replay file for tap from SQL source and Other source
- Fix mypy + docstring errors

----
:books: Documentation preview :books:: https://meltano-sdk--1434.org.readthedocs.build/en/1434/


tayloramurphy:
Thanks for the PR @flexponsive - looks like we've got some engineers on it as reviewers 👍

edgarrmondragon (commented):
@flexponsive should we add these new cases to the GitHub workflow file?

flexponsive:
@edgarrmondragon good catch! thanks for pointing it out. added the missing cases

edgarrmondragon (approved):

edgarrmondragon:
thanks @flexponsive!

PR #1449, opened 2023-02-22
fix: Cookiecutter pydocstring formatting updates

radbrt:
When creating a new tap that includes the github CI action, the pydocstyle tests fails on a number of issues - I'm guessing because of updates to the pydocstyle guide.

These changes are to update formatting of docstring so that `poetry run pydocstyle <tap-name>` does not return errors on a cleanly generated template.

Alternatively, disabling rule D406 and D407 would go a long way in resolving this.

----
:books: Documentation preview :books:: https://meltano-sdk--1449.org.readthedocs.build/en/1449/

### DIFF
--- a/cookiecutter/tap-template/{{cookiecutter.tap_id}}/{{cookiecutter.library_name}}/tap.py
+++ b/cookiecutter/tap-template/{{cookiecutter.tap_id}}/{{cookiecutter.library_name}}/tap.py
@@ -52,8 +52,10 @@ class Tap{{ cookiecutter.source_name }}({{ 'SQL' if cookiecutter.stream_type == 
     def discover_streams(self) -> list[streams.{{ cookiecutter.source_name }}Stream]:
         """Return a list of discovered streams.
 
-        Returns:
+        Returns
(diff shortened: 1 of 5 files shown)


radbrt:
Maybe check that the issue appears on your system as well, so that this isn't just an issue with my own environment.

edgarrmondragon:
We should remove pydocstyle here https://github.com/meltano/sdk/blob/d034b3e0e8336424767d54a3144a503dc7290875/cookiecutter/tap-template/%7B%7Bcookiecutter.tap_id%7D%7D/tox.ini#L16

and discourage its direct use since we're using the flake8 anyway, so we should rely on [`flake8-docstrings`](https://pypi.org/project/flake8-docstrings/) instead.

Wdyt @radbrt?

radbrt:
OK I have no idea why this fails on flake8 now... 😕

radbrt:
@edgarrmondragon These static code checkers are quite foreign to me anyways, I'm fine with closing this one and just removing pydocstyle altogether.

PR #1467, opened 2023-02-28
fix(deps): Remove `debugging` and `samples` extras (they were never really available to users)

### LINKED ISSUE #1465: bug: Package extras in `pyproject.toml` are not valid
### Singer SDK Version

0.21.0

### Is this a regression?

- [ ] Yes

### Python Version

NA

### Bug scope

Other

### Operating System

NA

### Description

```console
$ poetry --version
Poetry (version 1.4.0)

$ poetry check
Error: Cannot find dependency "pyarrow" for extra "samples" in main dependencies.
Error: Cannot find dependency "viztracer" for extra "debugging" in main dependencies.
```

### Code

_No response_

edgarrmondragon:
Closes #1465 

----
:books: Documentation preview :books:: https://meltano-sdk--1467.org.readthedocs.build/en/1467/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,6 +99,10 @@ mypy = "^1.0"
 cookiecutter = "^2.1.1"
 PyYAML = "^6.0"
 freezegun = "^1.2.2"
+numpy = [
+    { version = "<1.22", python = "<3.8" },
+    { version = ">=1.22", python = ">=3.8" },
+]
 requests-mock = "^1.10.0"
 sqlalchemy2-stubs = {version = "^0.0.2a32", allow-prereleases = true}
 types-jsonschema = "^4.17.0.6"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,7 +108,7 @@ types-requests = "^2.28.11"
 types-simplejson = "^3.18.0"
 types-PyYAML = "^6.0.12"
 coverage = {extras = ["toml"], version = "^7.2"}
-pyarrow = {version = "^11.0.0", python = "<3.11"}
+pyarrow = "^11.0.0"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -117,6 +108,7 @@ types-requests = "^2.28.11"
 types-simplejson = "^3.18.0"
 types-PyYAML = "^6.0.12"
 coverage = {extras = ["toml"], version = "^7.2"}
+pyarrow = {version = "^11.0.0", python = "<3.11"}
```
WillDaSilva, line 111: PyArrow is available for Python 3.11, so why the restriction?

What are the implications of missing the PyArrow dependency? Presumably nothing significant since it was previously an optional dependency, but I'd like to know.

WillDaSilva (approved):

pyproject.toml
edgarrmondragon: Yeah, I think there used to be some conflicts between numpy and pyarrow but those seem resolved now. I'll remove the restriction.

WillDaSilva (approved):

pyproject.toml
edgarrmondragon: Oops, it's still the case: https://github.com/meltano/sdk/actions/runs/4359995790/jobs/7622418687

edgarrmondragon: > Do you know why we're using an old version of Numpy here?

@WillDaSilva Edited my comment above as you were typing 😅. It's because only such old version supports Python 3.7

WillDaSilva: As a solution for the interim, we could add two dependencies for PyArrow: an old one for Python 3.7, and the one we have selected currently for all other versions of Python.

PR #1471, opened 2023-03-01
feat: Support custom JSON schema validation and string format checkers in targets

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1471.org.readthedocs.build/en/1471/


BuzzCutNorman:
@edgarrmondragon I like it so far.  No format checking at all to start and then you can add checks as you need them.  When working with this I added a default format checker that I knew would fail `date` and `time`.  Then created a simple `is_allowed` function to get past the validation fail to see how the values filed. 
```
    def get_format_checker(self) -> FormatChecker:
        format_checker: FormatChecker = Draft3Validator.FORMAT_CHECKER

        def is_allowed(value):
            return True

        format_checker.checks("time")(is_allowed)
        format_checker.checks("date")(is_allowed)
        return format_checker
```

edgarrmondragon:
Related: https://github.com/meltano/sdk/issues/1831

tests/core/sinks/test_format_checker.py
edgarrmondragon, line 1: @BuzzCutNorman Do you think this is still relevant? The original implementation was rather specific `python-jsonschema` so I wonder if we should keep this?

BuzzCutNorman: @edgarrmondragon If you would please clarify if the question of relevance is about `test_format_checker.py` or the PR in general?

edgarrmondragon: @BuzzCutNorman I commented on the wrong PR 😅. I meant https://github.com/meltano/sdk/pull/2136#discussion_r1442494880

PR #1480, opened 2023-03-04
feat: Added `RESTStream.backoff_jitter` to support custom backoff jitter generators

### LINKED ISSUE #1477: docs: Custom-Backoff header sample doesn't work due to backoff defaulting to "full jittter"
### Documentation type

How-to guides

### Description

Team I'm working with followed the steps from https://sdk.meltano.com/en/latest/code_samples.html#custom-backoff  with an API that gives a 429 response with a Header including the amount of time to wait before trying their next request.

Copied the implementation and they noticed the wait time wasn't even close to what they were passing in. We dove in and found out that by default https://github.com/litl/backoff/blob/master/backoff/_jitter.py#L18 `full_jitter` is now the default since backoff 1.2 also see https://github.com/litl/backoff/blob/master/backoff/_decorator.py#L32-L63

To make us wait the exact amount of time we had to override the request decorator to be

```
    def request_decorator(self, func: Callable) -> Callable:
        """Instantiate a decorator for handling request failures.
        Uses a wait generator defined in `backoff_wait_generator` to
        determine backoff behaviour. Try limit is defined in
        `backoff_max_tries`, and will trigger the event defined in
        `backoff_handler` before retrying. Developers may override one or
        all of these methods to provide custom backoff or retry handling.
        Args:
            func: Function to decorate.
        Returns:
            A decorated method.
        """
        decorator: Callable = backoff.on_exception(
            self.backoff_wait_generator,
            (
                ConnectionResetError,
                RetriableAPIError,
                requests.exceptions.ReadTimeout,
                requests.exceptions.ConnectionError,
            ),
            max_tries=self.backoff_max_tries,
            on_backoff=self.backoff_handler,
            jitter=None
        )(func)
        return decorator
```

Following the steps they overrode the wait generator doing something similar to what's below
```
def backoff_wait_generator() -> Callable[..., Generator[int, Any, None]]:
    def _backoff_from_headers(retriable_api_error):
        response_headers = retriable_api_error.response.headers
        return int(response_headers.get("Retry-After", 0))

    return self.backoff_runtime(value=_backoff_from_headers)
```

We should do one of these I think:

1. Tell folks to set `jitter` to None in the request decorator
1. Default to having jitter set to  [`random_jitter`](https://github.com/litl/backoff/blob/master/backoff/_jitter.py#L6-L15) , which is much better than the `full_jitter` option for what people would expect to have happen
1. Add a `jitter` property in the SDK that allows folks to set jitter to random, full, or None.

visch:
Closes #1477 

----
:books: Documentation preview :books:: https://meltano-sdk--1480.org.readthedocs.build/en/1480/

### DIFF
--- a/docs/code_samples.md
+++ b/docs/code_samples.md
@@ -339,6 +339,8 @@ def backoff_wait_generator() -> Callable[..., Generator[int, Any, None]]:
 
 To utilise a response header as a wait value you can use [`backoff_runtime`](singer_sdk.RESTStream.backoff_runtime), and pass a method that returns a wait value:
 
+**Note**: By default there's jitter added which makes this function wait a bit longer than the value provided. To disable jitter override `backoff_jitter`](singer_sdk.RESTStream.backoff_jitter). In sdk versions 0.21.0 the default jitter function will make the function below not work as you would expect without disabling jitter, ([here](https://github.com/meltano/sdk/issues/1477) for more information) to disable jitter override the `request_decorator` and pass `jitter=None` to the `backoff.on_exception` function. 
+
 ```
 def backoff_wait_generator() -> Callable[..., Generator[int, Any, None]]:
     def _backoff_from_headers(retriable_api_error):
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):
Thanks @visch!

PR #1499, opened 2023-03-15
chore: Enable `ISC` Ruff checks

edgarrmondragon:
Motivation for [`flake8-implicit-str-concat`](https://pypi.org/project/flake8-implicit-str-concat/):

> This is a plugin for the Python code-checking tool Flake8 to encourage correct string literal concatenation.
>
> It looks for style problems like implicitly concatenated string literals on the same line (which can be introduced by the code-formatting tool Black), or unnecessary plus operators for explicit string literal concatenation.

----
:books: Documentation preview :books:: https://meltano-sdk--1499.org.readthedocs.build/en/1499/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -230,6 +230,7 @@ select = [
     "ANN",  # flake8-annotations
     "COM",  # flake8-commas
     "T10",  # flake8-debugger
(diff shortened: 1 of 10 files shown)


WillDaSilva (approved):
This is great!

Issue #1502, opened 2023-03-16
Feat: add the `json` module to stream maps context


aaronsteers:
Per

- #1500 

Specifically, `json.loads()` and `json.dumps()` would be helpful.

edgarrmondragon:
@aaronsteers Do we need all members of `json` or can we do with just `json.loads` and `json.dumps`?

aaronsteers:
> @aaronsteers Do we need all members of `json` or can we do with just `json.loads` and `json.dumps`?

Good question. Want to add thoughts in the PR?

I like adding `json` because it's simplest and doesn't require retraining from what users would expect. But I don't mind if we want to implement in another way - especially if for security reasons we want to keep to "safe" implementations only.

edgarrmondragon:
Still relevant

edgarrmondragon:
Things to keep in mind for anyone that's interested in contributing this:

* How to communicate schema types? Perhaps aliases of `json.loads` should be used, which the SDK can identify as specific types, e.g. `parse_json_object`, `parse_json_array`, etc.
* See https://github.com/meltano/sdk/blob/97d97de313511be4c21b842bf2ba0283d4bf5c02/singer_sdk/mapper.py#L297-L307

Issue #1500, opened 2023-03-15
feat: Support for masking & JSON-based hashing


sbalnojan:
### Feature scope

Taps (catalog, state, stream maps, etc.)

### Description

Given data rows like:
`{"foo": "bar", "nested": {"foo": "bar"}}`
`{"foo": "bar2", "nested": {"foo": "bar"}}`
`{"foo": "ar", "nested": {"foo": "bar"}}`

it sounds reasonable to do some masking using the stream maps capability to map these to:

`{"foo": "b", "nested": {"foo": "bar"}}`
`{"foo": "b", "nested": {"foo": "bar"}}`
`{"foo": "a", "nested": {"foo": "bar"}}`

I thought the following would work, but apparently, simpleeval doesn't support some parts of it:

```
last_name:   "config.get(last_name) if config.get(last_name) else 'not-found'"
 last_name_old: last_name
 stream_map_config:
Book: mask-1
Tire: mask-1
Dorian: mask-2
```

For the nested object (where e.g. my tap-csv implementation delivers a string = '{"foo": "bar"}') I would like to go into that string in a native JSON-like manner and hash "foo". However, I couldn't figure out a way of turning the string into a dictionary object, like e.g.

`md5(json.loads(info)['addr'])`

Do these two features make sense? Or is there another way of doing them that I am missing? @edgarrmondragon @aaronsteers

aaronsteers:
@sbalnojan - I think I understand the use case but I'm having a hard time following the sample case. 

In terms of simpleeval support, at least in theory it should support any syntax that python supports.

So, taking the first character of a string column like last_name should be doable with something like `last_name[0]` or `last_name.left(1)`.

You shouldn't need to use the `json` module - instead you can access nested data from the record using Python bracket syntax.

Does that help?

aaronsteers:
> For the nested object (where e.g. my tap-csv implementation delivers a string = '{"foo": "bar"}') I would like to go into that string in a native JSON-like manner and hash "foo". However, I couldn't figure out a way of turning the string into a dictionary object, like e.g.
> 
> md5(json.loads(info)['addr'])

The `json` module isn't imported into context but it easily could be...

sbalnojan:
@aaronsteers ah yes, so I'm looking for a native JSONy way to have at least the ease of manipulating the data the transferwise mapper has. And that one supports JSON-path navigation (https://github.com/transferwise/pipelinewise-transform-field).

[...]

PR #1503, opened 2023-03-16
feat: import json module for stream maps, add tests

aaronsteers:
First draft. 

Resolves:

- #1502

Related to:

- #1500

### DIFF
--- a/singer_sdk/mapper.py
+++ b/singer_sdk/mapper.py
@@ -9,9 +9,9 @@ import abc
 import copy
 import datetime
 import hashlib
+import json
 import logging
 import sys
-import json
 from typing import Any, Callable, Dict, Union
 
 from singer_sdk._singerlib.catalog import Catalog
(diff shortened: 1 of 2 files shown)


singer_sdk/mapper.py
```diff
@@ -326,7 +338,8 @@ def _eval(
         Raises:
             MapExpressionError: If the mapping expression failed to evaluate.
         """
-        names = record.copy()  # Start with names from record properties
+        names = self.module_names  # Start with the module names
+        names.update(record.copy())  # Add top-level record properties
```
edgarrmondragon, line 342: If the record has a field named `json` it'd override the `json` module. I also don't think we need all the members of `json`, just `dumps` and `loads`.

edgarrmondragon: We could add them to the functions mapping:

```python
funcs["json.loads"] = json.loads
funcs["json.dumps"] = json.dumps
```

aaronsteers:
@edgarrmondragon - This is another PR I opened a while ago and wasn't high enough priority to push to delivery. I think I'll close for now, since it needs a tiny bit more work and I don't currently have capacity to get it completed. If anyone else wants to pick this up, they can use this as a starting point for resolving this one:

- #1502

PR #1505, opened 2023-03-16
chore: Enable `RSE` Ruff checks

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1505.org.readthedocs.build/en/1505/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -235,6 +235,7 @@ select = [
     "INP",  # flake8-no-pep420
     "PIE",  # flake8-pie
     "PT",   # flake8-pytest-style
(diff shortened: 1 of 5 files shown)


WillDaSilva (commented):
We may want to disable `R100` (`raise` in `except` handler without `from`), since raising in an except handler without from sets `__context__`, versus raising with `from` which sets `__cause__` and `__context__`. Either way you get the data about the chain of exceptions, but they have different semantics, and universally using `from` when raising within an except handler is not correct.

WillDaSilva (approved):

PR #1526, opened 2023-03-24
fix: Handle missing `type` value when checking JSON schema types

edgarrmondragon:
Uncovered in https://github.com/MeltanoLabs/tap-dbt/pull/119

----
:books: Documentation preview :books:: https://meltano-sdk--1526.org.readthedocs.build/en/1526/


mjsqu:
Looks good, in the case we found in `tap-dbt`, this would loop over the `"anyOf"` list and skip past `"enum"` without returning anything but output `True` when encountering the `"type"` key. The test looks exactly like the case in `tap-dbt`. Great work @edgarrmondragon 👍

PR #1563, opened 2023-03-30
feat(templates): Migrate shell scripts for cookiecutter e2e tests to Nox

### LINKED ISSUE #1423: [Feature]: Migrate shell scripts for cookiecutter e2e tests to Nox
### Feature scope

Other

### Description

Follow-on https://github.com/meltano/sdk/pull/1410

mkranna:
Closes #1423: Migrate shell scripts for cookiecutter e2e tests to Nox.

- First, I added a new parametrised session `test_cookiecutter` to the main project `noxfile.py`, which replaces the `test_cookiercutter.sh` (parametrised by the json-files in  e2e-tests/cookiecutter folder). To run *all* cookiecutter replay files, use the following terminal command: `nox  --session=test_cookiecutter`.

- Second, I updated the `cookiecutter-e2e.yml` to simply call the nox session as above. It is no longer required to maintain a list of the replay files in that YML.

- Comment: the advantage of the current `cookiecutter-e2e.yml` is that is it no longer lists replay files, but on the other hand the tests takes longer since they are not run in parallel. Alternatively, I can call tests in parallel as it was done before.

----
:books: Documentation preview :books:: https://meltano-sdk--1563.org.readthedocs.build/en/1563/


noxfile.py
```diff
@@ -170,3 +172,73 @@ def docs_serve(session: Session) -> None:
...
+@nox.parametrize("replay_file_path", glob.glob("./e2e-tests/cookiecutters/*.json"))
+@session(python=main_python_version)
+def test_cookiecutter(session: Session, replay_file_path) -> None:
+    """Uses the tap template to build an empty cookiecutter.
+
+    Runs the lint task on the created test project.
+    """
+    args = session.posargs or ["1"]
+
+    cc_build_path = "/tmp"  # noqa: S108
```
edgarrmondragon, line 186: I think you can use [`Session.create_tmp`](https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.create_tmp) here. Wdyt?

noxfile.py
```diff
@@ -170,3 +172,73 @@ def docs_serve(session: Session) -> None:
...
+
+    session.run(
+        "cookiecutter",
+        "--replay-file",
+        str(replay_file),
+        str(tap_template),
+        "-o",
+        cc_build_path,
+    )
+    os.chdir(cc_test_output)
```
edgarrmondragon, line 221: [`Session.chdir`](https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.chdir) might be safer

noxfile.py
mkranna: Thanks! But there is a little issue with build artifacts: we need to know which folder has the cookiecutters. 

Would  tempfile.gettempdir() work? Since it returns the name of the directory used for temporary files.

noxfile.py
mkranna: Thanks! Already fixed.

edgarrmondragon (approved):
This looks great!

Screenshot 2023-03-31 at 11 26 54

Thanks @mkranna!

PR #1569, opened 2023-03-31
chore: Enable `PLC`, `PLE`, `PLR` and `PLW` Ruff checks

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1569.org.readthedocs.build/en/1569/


pyproject.toml
WillDaSilva, line 1: I'd recommend increasing the number of parameters allowed by PLR0913, or disabling it. It defaults to 5, and unfortunately makes no distinction between positional and keyword arguments.

```
[tool.ruff.pylint]
max-args = ...
```

WillDaSilva (approved):

pyproject.toml
edgarrmondragon: It does make a distinction for keyword-only arguments so I made the change where it's safe, and also set `max-args` to 9.

PR #1572, opened 2023-03-31
feat: add `batch_config` handling in `append_builtin_config()`

### LINKED ISSUE #1571: Add `batch_config` to handling in `append_builtin_config()`
It looks like as of yet, we don't declare or advertise the "batch_config" config options, although we do provide instructions in the `batch.md` docs.

I may have a PR up for this shortly.

aaronsteers:
Resolves #1571

I discovered this when trying to configure another tap's config json schema.

----
:books: Documentation preview :books:: https://meltano-sdk--1572.org.readthedocs.build/en/1572/


singer_sdk/configuration/_dict_config.py
```diff
@@ -99,3 +99,18 @@ def merge_config_sources(
         config.update(read_json_file(config_path))
 
     return config
+
+
+def merge_missing_config_jsonschema(
```
aaronsteers, line 104: Refactored to a helper function in this private module. Previously it was a local function in the base class's method.

singer_sdk/helpers/capabilities.py
```diff
@@ -50,6 +50,34 @@
...
+        wrapped=ObjectType(
+            Property(
+                "encoding",
+                description="Specifies the format and compression of the batch files.",
+                wrapped=ObjectType(
+                    Property("format", StringType, allowed_values=["jsonl"]),
+                    Property(
+                        "compression",
+                        StringType,
+                        allowed_values=["gzip", "none"],
```
aaronsteers, line 66: I think omitting "compression" is the same as "none". While "none" isn't strictly handled, the condition I found checks for "gzip" and is a no-op otherwise. @edgarrmondragon - Thoughts?

edgarrmondragon: Yeah, this works fine and is explicit 👍

edgarrmondragon (approved):

aaronsteers:
@edgarrmondragon - I'm not able to add any further commits to this branch. What are your thoughts about getting this merged? 

I think we had a failure specific to python3.8 but I didn't dig deep into that issue to analyze root cause.

As a quick first step, do you think you could update this branch from `main` and see if test is still failing?

Based on that info, I may better be able to say if I would have to open a new PR, or conversely if you feel comfortable to get it merged on your side.

singer_sdk/tap_base.py
```diff
@@ -180,6 +182,28 @@ def capabilities(self) -> list[CapabilitiesEnum]:
...
+
+        For all except very advanced use cases, we recommend leaving these
+        implementations "as-is", since this provides the most choice to users and is
+        the most "future proof" in terms of taking advantage of built-in capabilities
+        which may be added in the future.
+
+        Args:
+            config_jsonschema: [description]
+        """
+        super().append_builtin_config(config_jsonschema)
```
aaronsteers, line 201: The `mypy` complaint is on this line. Error is:

> singer_sdk/tap_base.py:201: error: Argument 2 for "super" not an instance of argument 1  [misc]

Probably fixed by using a more specific `super()` invocation.

edgarrmondragon: @aaronsteers Yeah, that was it 🙂

aaronsteers: Oh, I see! Thanks @edgarrmondragon for fixing that! 👍 🚀

aaronsteers:
@edgarrmondragon - Thanks very much for your help on this. I'm excited to see it merge.

I noticed that it is still in auto-merge state and I don't have permission to disable that. 

I'm hesitating to mark resolved the final conversation thread above because that (probably?) will trigger the merge. But if you feel comfortable with this, you can go ahead and do so. 👍  For my side, it looks like this is ready to go thanks to your additional commits. 🙏

kgpayne (approved):

PR #1601, opened 2023-04-11
feat: Allow skipping child streams by returning an empty child context from parent stream

edgarrmondragon:


singer_sdk/streams/core.py
```diff
@@ -1238,20 +1246,24 @@ def _get_state_partition_context(self, context: dict | None) -> dict | None:
...
+    def get_child_context(self, record: dict, context: dict | None) -> dict | None:
         """Return a child context object from the record and optional provided context.
 
         By default, will return context if provided and otherwise the record dict.
 
         Developers may override this behavior to send specific information to child
         streams for context.
 
+        Return ``None`` if no child streams should be synced, for example if the
+        parent record was deleted and the child records can no longer be synced.
```
aaronsteers, line 1258: Do we also need to handle cases where `get_records()` returns a tuple combining the record and child_context?

edgarrmondragon: Good catch. They're handled upstream here: https://github.com/meltano/sdk/blob/0e2d2b6729f715a3665401899966b602092b80ef/singer_sdk/streams/core.py#L1096-L1106

but we should update the docs and annotation for `get_records()`.

edgarrmondragon: @aaronsteers Done 32fb526ce37ce2aed11ee7457115b0eff63d0669

aaronsteers (approved):
Nice work!! 🚀

PR #1603, opened 2023-04-12
feat: Allow `allowed_values` and `examples` in any JSON schema type constructor

edgarrmondragon:
Docs updates:

- [`PropertiesList` usage example](https://meltano-sdk--1603.org.readthedocs.build/en/1603/typing.html#usage-example)
- [`StringType` with and without initialization](https://meltano-sdk--1603.org.readthedocs.build/en/1603/classes/typing/singer_sdk.typing.StringType.html)

----
:books: Documentation preview :books:: https://meltano-sdk--1603.org.readthedocs.build/en/1603/

### DIFF
--- a/singer_sdk/typing.py
+++ b/singer_sdk/typing.py
@@ -459,20 +459,20 @@ class ArrayType(JSONTypeHelper[list], Generic[W]):
         return {"type": "array", "items": self.wrapped_type.type_dict, **self.extras}
 
 
-class Property(JSONTypeHelper, Generic[W]):
+class Property(JSONTypeHelper[T], Generic[T]):
     """Generic Property. Should be nested within a `PropertiesList`."""
 
     # TODO: Make some of these arguments keyword-only. This is a breaking change.
     def __init__(
         self,
         name: str,
-        wrapped: W | type[W],
+        wrapped: JSONTypeHelper[T] | type[JSONTypeHelper[T]],
         required: bool = False,  # noqa: FBT001, FBT002
-        default: _JsonValue | None = None,
+        default: T | None = None,
         description: str | None = None,
         secret: bool | None = False,  # noqa: FBT002
-        allowed_values: list[Any] | None = None,
-        examples: list[Any] | None = None,
+        allowed_values: list[T] | None = None,
+        examples: list[T] | None = None,
     ) -> None:
         """Initialize Property object.
 
@@ -661,7 +661,7 @@ class ObjectType(JSONTypeHelper):
             merged_props.update(w.to_dict())
             if not w.optional:
                 required.append(w.name)
-        result: dict = {"type": "object", "properties": merged_props}
+        result: dict[str, Any] = {"type": "object", "properties": merged_props}
 
         if required:
             result["required"] = required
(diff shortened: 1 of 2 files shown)


tests/core/test_jsonschema_helpers.py
```diff
@@ -434,6 +434,50 @@ def test_inbuilt_type(json_type: JSONTypeHelper, expected_json_schema: dict):
             },
             {is_integer_type},
         ),
+        (
+            Property(
```
TyShkan, line 438: Would be nice to add this example to "Usage example" to have it in docs

edgarrmondragon: @TyShkan Thanks for the suggestion! I've updated the usage example: https://meltano-sdk--1603.org.readthedocs.build/en/1603/typing.html#usage-example

singer_sdk/typing.py
```diff
@@ -10,6 +10,15 @@
 
         Property("id", IntegerType, required=True),
         Property("foo_or_bar", StringType, allowed_values=["foo", "bar"]),
+        Property(
+            "permissions",
+            ArrayType(
+                th.StringType(
```
TyShkan, line 16: It was my bad to bring `th.` prefix here 🙈

edgarrmondragon: Good catch, fixed: 905a77eb

singer_sdk/typing.py
```diff
@@ -159,25 +171,87 @@ def set_defaults(
...
+
+        Args:
+            instance: The instance to get the property value from.
+            owner: The class to get the property value from.
+
+        Returns:
+            The property value.
+        """
+        if instance is None:
+            instance = owner()
```
edgarrmondragon, line 201: @aaronsteers I'd like to get your thoughts on this approach. It replaces the `classproperty` descriptor with one that gets the attribute from

1) the instance of the class, e.g. `StringType().type_dict` or
2) an instance initialized with no arguments, e.g. `StringType.type_dict` (under the hood ``StringType().type_dict` is called).

also cc @WillDaSilva if you think this is not a good pattern

WillDaSilva: @edgarrmondragon Let me know if I understand the context correctly.

We currently use `classproperty` with the expectation that the class object will be used directly instead of instances of the class.

We want to allow customization by providing `allowed_values` and `examples` to the class, but we don't want to introduce a breaking change by requiring the class be initialized in all cases.

To accommodate this new feature while continuing to support the old way to using the classes directly (instead of via an instance) this change adds a new property type that uses the instance if it exists, and creates an instance (with the default keyword arguments) otherwise.

Seems reasonable. One hazard to watch out for is that none of the classes can ever have a parameter without a default. We _could_ add a metaclass to `JSONTypeHelper` that enforces that restriction so that we never accidentally forget and introduce a bug, but that may be overkill since our tests should always catch bugs like that.

edgarrmondragon: That's a very accurate description of the problem @WillDaSilva. I agree that we risk declaring a class with positional arguments but we don't often add new type helper classes and you're right that tests should catch them.

I've added more doctest for other type helper classes.

WillDaSilva (approved):

TyShkan (approved):

PR #1611, opened 2023-04-15
Taps should use the same connector

visch:
Right now we're generating a new connector, which negates defining self.connector() in the Tap. This fixes that issue

This came up while adding the SSH Tunnel feature here https://github.com/MeltanoLabs/tap-postgres/pull/94 . We're really specific about how sqlalchemy_url is passed to the `connector` and can't rely on what is set in `config` as we have to alter it dynamically. Most folks don't change the sqlalchemy_url which is why no one noticed

----
:books: Documentation preview :books:: https://meltano-sdk--1611.org.readthedocs.build/en/1611/

### DIFF
--- a/singer_sdk/tap_base.py
+++ b/singer_sdk/tap_base.py
@@ -595,10 +595,8 @@ class SQLTap(Tap):
         if self.input_catalog:
             return self.input_catalog.to_dict()
 
-        connector = self.default_stream_class.connector_class(dict(self.config))
-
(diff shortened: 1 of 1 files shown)


visch:
After all the test failures I've realized I didn't think this one out all the way for the SDK. I"ll just implement this in https://github.com/MeltanoLabs/tap-postgres/pull/94 and look later!

PR #1621, opened 2023-04-18
chore: Enable `G` (flake8-logging-format) Ruff checks

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1621.org.readthedocs.build/en/1621/


WillDaSilva (commented):
@edgarrmondragon Thoughts on using `{}` style interpolation instead of the default `%` style? https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

edgarrmondragon:
> @edgarrmondragon Thoughts on using `{}` style interpolation instead of the default `%` style? https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

@WillDaSilva We can't use that in the logging calls, only in the formatter instantiation (e.g. in the yaml logging config file):

<blockquote>

<p>Logging calls (<a class="reference internal" href="../library/logging.html#logging.Logger.debug" title="logging.Logger.debug"><code class="xref py py-meth docutils literal notranslate"><span class="pre">debug()</span></code></a>, <a class="reference internal" href="../library/logging.html#logging.Logger.info" title="logging.Logger.info"><code class="xref py py-meth docutils literal notranslate"><span class="pre">info()</span></code></a> etc.) only take
positional parameters for the actual logging message itself, with keyword
parameters used only for determining options for how to handle the logging call
(e.g. the <code class="docutils literal notranslate"><span class="pre">exc_info</span></code> keyword parameter to indicate that traceback information
should be logged, or the <code class="docutils literal notranslate"><span class="pre">extra</span></code> keyword parameter to indicate additional
contextual information to be added to the log). So you cannot directly make
logging calls using <a class="reference internal" href="../library/stdtypes.html#str.format" title="str.format"><code class="xref py py-meth docutils literal notranslate"><span class="pre">str.format()</span></code></a> or <a class="reference internal" href="../library/string.html#string.Template" title="string.Template"><code class="xref py py-class docutils literal notranslate"><span class="pre">string.Template</span></code></a> syntax,
because internally the logging package uses %-formatting to merge the format
string and the variable arguments. There would be no changing this while preserving
backward compatibility, since all logging calls which are out there in existing
code will be using %-format strings.</p>

</blockquote>

WillDaSilva (approved):

PR #1647, opened 2023-04-27
chore: Remove `poetry-lock` pre-commit check

edgarrmondragon:
Dependabot is still on Poetry 1.3 and our pre-commit check is on 1.4.0 so this is causing local pre-commit checks to fail and update the `poetry.lock` after a Dependabot PR was merged.

----
:books: Documentation preview :books:: https://meltano-sdk--1647.org.readthedocs.build/en/1647/


WillDaSilva (approved):
> Dependabot is still on Poetry 1.3

:disappointed:

PR #1648, opened 2023-04-28
feat(templates): Use Ruff to lint projects generated with Cookiecutter templates

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1648.org.readthedocs.build/en/1648/


WillDaSilva (approved):
Looks good to me. I found one issue where a fresh SQL target would fail a Ruff check, but I pushed a commit to fix it.

This is a significant reduction of the number of dev dependencies each SDK-based tap/target will have, which is great for avoiding dependency hell.

I noticed that the Ruff target version is `py37`. That makes sense for now, but we'll want to update that to `py38` on 2023-06-27 per https://devguide.python.org/versions/

edgarrmondragon:
> Looks good to me. I found one issue where a fresh SQL target would fail a Ruff check, but I pushed a commit to fix it.
> 
> This is a significant reduction of the number of dev dependencies each SDK-based tap/target will have, which is great for avoiding dependency hell.
> 
> I noticed that the Ruff target version is `py37`. That makes sense for now, but we'll want to update that to `py38` on 2023-06-27 per https://devguide.python.org/versions/

thanks @WillDaSilva, I added a test for that missing case to make sure both "per record" and SQL targets pass Ruff checks.

PR #1668, opened 2023-05-04
refactor: break out default batch file writer into separate class

kgpayne:
----
:books: Documentation preview :books:: https://meltano-sdk--1668.org.readthedocs.build/en/1668/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,12 +1,16 @@
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
 sphinx:
   builder: html
   configuration: docs/conf.py
   fail_on_warning: true
 
 python:
-  version: 3.8
   install:
     # - requirements: docs/requirements.txt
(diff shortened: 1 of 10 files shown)


kgpayne:
@edgarrmondragon this is to better reuse batch file writers on the target side. I.e. cases where the target supports batch files (in json or csv) but receives records. E.g. [target-snowflake](https://github.com/MeltanoLabs/target-snowflake/blob/main/target_snowflake/sinks.py#L355)

edgarrmondragon:
> @edgarrmondragon this is to better reuse batch file writers on the target side. I.e. cases where the target supports batch files (in json or csv) but receives records. E.g. [target-snowflake](https://github.com/MeltanoLabs/target-snowflake/blob/main/target_snowflake/sinks.py#L355)

@kgpayne This is looking great. Should we document these new batcher classes in the reference section (which is actually getting crowded imo)?

edgarrmondragon (approved):

PR #1672, opened 2023-05-05
feat: Support deselecting streams by default

### LINKED ISSUE #1203: [Feature]: Support de-selecting a stream by default
### Feature scope

### Description

Tap developers may want to disable certain streams in their taps for a few reasons:

- The stream requires higher access credentials than the rest
- The stream requires a different payment tier in the source system
- The stream has a large volume of data and not all users may be interested in it by default

The discovered catalog would flag the stream with `"selected_by_default": false`. This means that with no catalog file, the stream selection will be resolved to `False`.

End users could still toggle the stream selection as normal by passing a catalog file.

Related discussion:

- https://github.com/meltano/sdk/discussions/1652

edgarrmondragon:
* Closes #1203 

TODO:

- [x] Tests
- [x] Docs


edgarrmondragon:
Surprisingly, test coverage didn't drop but I still want to add a few test cases

kgpayne (approved):

PR #1680, opened 2023-05-08
fix: Pin `urllib3` to `<2` to avoid incompatibility issues with botocore

edgarrmondragon:
See https://meltano.slack.com/archives/C01PKLU5D1R/p1683304999507349

----
:books: Documentation preview :books:: https://meltano-sdk--1680.org.readthedocs.build/en/1680/


pyproject.toml
```diff
@@ -44,7 +44,7 @@ pendulum = "^2.1.0"
 click = "~=8.0"
 fs = "^2.4.16"
 PyJWT = "~=2.4"
-requests = "^2.25.1"
+requests = ">=2.25.1,<2.30"
```
WillDaSilva, line 47: A comment explaining the issue would be appreciated.

WillDaSilva (approved):

WillDaSilva (approved):
@edgarrmondragon How soon do you think this restriction could be removed? As time goes on more and more libraries will have dependencies on `urllib3 >= 2`.

edgarrmondragon:
@WillDaSilva Hopefully as soon as botocore loosens its constraint on urllib3: https://github.com/boto/botocore/blob/f49c300cbc363a83821abac236ac57b4d43d1e7c/setup.py#L28

prevostc:
> @WillDaSilva Hopefully as soon as botocore loosens its constraint on urllib3: https://github.com/boto/botocore/blob/f49c300cbc363a83821abac236ac57b4d43d1e7c/setup.py#L28
> 
> [boto/botocore#2926](https://github.com/boto/botocore/issues/2926)

I believe that should be fixed now https://github.com/boto/botocore/blob/1.34.20/setup.py#L29

edgarrmondragon:
> > @WillDaSilva Hopefully as soon as botocore loosens its constraint on urllib3: https://github.com/boto/botocore/blob/f49c300cbc363a83821abac236ac57b4d43d1e7c/setup.py#L28
> > [boto/botocore#2926](https://github.com/boto/botocore/issues/2926)
> 
> I believe that should be fixed now https://github.com/boto/botocore/blob/1.34.20/setup.py#L29

Thanks for the reference @prevostc!

Unfortunately, we use `poetry export` and it doesn't seem to like either of these dependency declarations:

1. `urllib3 = ">=1.26"`

   ```
   The `urllib3` package has the following compatible candidates `[Package('urllib3', '1.26.18')]`;  but, the exporter dependency walker previously elected `urllib3 (2.0.7)` which is not compatible with the dependency `urllib3 (>=1.25.4,<1.27)`. Please contribute to `poetry-plugin-export` to solve this problem.
   ```

   https://github.com/python-poetry/poetry-plugin-export/issues/239

2. ```toml
   urllib3 = [
       {version = ">=1.26,<1.27", markers = "python_version < \"3.10\""},
       {version = ">=1.26,<2.1", markers = "python_version >= \"3.10\""},
   ]
   ```

   ```
   Dependency walk failed at urllib3 (>=1.21.1,<3)
   ```

   https://github.com/python-poetry/poetry-plugin-export/issues/176

So, sadly this means we have to

* wait for poetry-plugin-export to fix this
* or move away from poetry-plugin-export for testing/CI
* or move away from poetry entirely

PR #1687, opened 2023-05-09
feat: Skip hex encoding for `bytes` values of JSON strings with `contentEncoded` property

### LINKED ISSUE #1675: feat: Skip default hex encoding if JSON Schema keyword `contentEncoding` present
### Feature scope

### Description

When a record element is an instance of python `bytes` and `is_boolean_type` returns `False` the element is encoded by default in `hex`.  If the record element was already encoded and the JSON Schema `contentEncoding` keyword is present in the Schema Properties stating what encoding was used on the element the Target has to know to decode from hex first before decoding from the encoding method given in the `contentEncoding` property.  I propose if the `contentEncoding` property is present in the schema that the element be passed as is.

**Current `singer_sdk.helpers._typing._conform_primitive_property`**
```
def _conform_primitive_property(  # noqa: PLR0911
    elem: t.Any,
    property_schema: dict,
) -> t.Any:

... some other stuff that has been excluded for readability 

    if isinstance(elem, bytes):
        # for BIT value, treat 0 as False and anything else as True
        bit_representation: bool
        if is_boolean_type(property_schema):
            bit_representation = elem != b"\x00"
            return bit_representation
        return elem.hex()

... more stuff that has been excluded for readability 

   return elem
```

**A proposed version just to get discussion started:**
```
def _conform_primitive_property(  # noqa: PLR0911
    elem: t.Any,
    property_schema: dict,
) -> t.Any:

... some other stuff that has been excluded for readability 

    if isinstance(elem, bytes):
        # for BIT value, treat 0 as False and anything else as True
        bit_representation: bool
        if is_boolean_type(property_schema):
            bit_representation = elem != b"\x00"
            return bit_representation
        if property_schema.get("contentEncoding"):
            return elem
        return elem.hex()

... more stuff that has been excluded for readability 

   return elem
```

BuzzCutNorman:
This feature enables a way for binary string values that are already encoded from being encoded again in hex by the _conform_primitive_property method.  If the JSON Schema keyword "contentEncoded" is present in the property_schema then the element is passed as is.   

A new typing check named is_encoded_string_type was put in place to assist in detecting that the element in question states in its JSON Schema it is type string and contentEncoding is present

Closes #1675

----
:books: Documentation preview :books:: https://meltano-sdk--1687.org.readthedocs.build/en/1687/


singer_sdk/helpers/_typing.py
```diff
@@ -489,6 +506,9 @@ def _conform_primitive_property(  # noqa: PLR0911
         if is_boolean_type(property_schema):
             bit_representation = elem != b"\x00"
             return bit_representation
+        # if property_schema.get("contentEncoding"):
```
edgarrmondragon, line 509:
```suggestion
```

singer_sdk/helpers/_typing.py
```diff
@@ -489,6 +506,9 @@ def _conform_primitive_property(  # noqa: PLR0911
...
+        if is_encoded_string_type(property_schema):
+            return elem
```
edgarrmondragon, line 511: Should this still be encoded as a string? I'm not sure how JSON libraries deal with bytes values.

BuzzCutNorman: I believe they should based on documentation if found at json-schema.org.  Please give it a look let me know what you think.

> JSON schema has a set of keywords to describe and optionally validate non-JSON data stored inside JSON strings. Since it would be difficult to write validators for many media types, JSON schema validators are not required to validate the contents of JSON strings based on these keywords. However, these keywords are still useful for an application that consumes validated JSON.

https://json-schema.org/understanding-json-schema/reference/non_json_data.html#media-string-encoding-non-json-data

singer_sdk/helpers/_typing.py
BuzzCutNorman: Thank you for pointing this out 🙏.  It has been removed.

BuzzCutNorman:
@edgarrmondragon Thank you for the example code comparing the `json` and `simplejson` libraries.  Going through them today using a simple mock-up program I found this feature could have another uses beside skipping the default hex encoding.  What I found was if a developer does the same thing I did and doesn't `.decode()` the `bytes` returned by `base64.b64encode` in `post_process` they end up here needing it to be decoded into a string.  I will add the code tomorrow but had to share.

singer_sdk/helpers/_typing.py
BuzzCutNorman: I agree that the values should be `str` before any JSON-encoding takes place.  Currently the '.hex()' default does this since it returns a string.  What if we ensure the values returned from the `bytes` portion of ` _conform_primitive_property` are valid ascii strings?  Since Batch messages don't go through this process and currently the `json.dumps` doesn't have a `default=str` we could extend the default encoding class to handle the byte type and pass that class and the encoder for it to use?  If you have that type of jsonencoder class would it make sense to use it for both record and batch message json.dumps?

BuzzCutNorman:
@edgarrmondragon 

I added the code I mentioned yesterday.  It attempts to do the following.  If you have a stream that has an element that was tagged as `{"type": ["null", "string"], "contentEncoded": "base64"}` and the element's value is a python `str` in the record dictionary before being sent through `_conform_primitive_property` then nothing is done with it.   If the element's value is a python `bytes` in the `_conform_primitive_property` method it checks that it `.isascii()` if it is then `elem.decode()` is used to turn the `bytes` to a `utf-8` encoded string and is returned.  If the element's value fails the `isascii()` check the default is to encode the element's value using `.hex()` which retuns the value as a hex enoded string.

Initially I thought this was going to be a great feature but the more I go through the review process I am starting to think it might not be that useful.  The reason I am starting to think this way is if I would have properly decoded the `bytes` returned by `base64.b64encode` in the first place I would have never run into the `bytes` being encoded in hex and sent as a string.   Which is kind of an edge case and the coding error should have easily been worked out.  In its current state this feature is a slight assist to developers who encode bytes and don't decode them.  I don't know if that is enough to warrant it being added.

Wdyt?

singer_sdk/helpers/_typing.py
```diff
@@ -489,6 +506,10 @@ def _conform_primitive_property(  # noqa: PLR0911
         if is_boolean_type(property_schema):
             bit_representation = elem != b"\x00"
             return bit_representation
+        # this is to detect bytes that have been encoded
+        # but not truned into utf-8 strings yet.
```
edgarrmondragon, line 510:
```suggestion
        # but not turned into utf-8 strings yet.
```

edgarrmondragon (changes requested):

edgarrmondragon:
@BuzzCutNorman this looks good, but after your last comment I worry that this may introduce a bit of tech debt in the handling of bytes values, where we should instead recommend developers to ensure raw values are never bytes.

> The reason I am starting to think this way is if I would have properly decoded the `bytes` returned by `base64.b64encode` in the first place I would have never run into the `bytes` being encoded in hex and sent as a string.

does this mean you're calling `base64.b64encode` manually in one of your taps?

cc @kgpayne wdyt?

BuzzCutNorman:
@edgarrmondragon 
> does this mean you're calling base64.b64encode manually in one of your taps?

Yes, I am in the post_process method. 😀
https://github.com/BuzzCutNorman/tap-mssql/blob/dc46ce1795f2f5dcc5dea475351a115c590c665e/tap_mssql/client.py#L445

BuzzCutNorman:
I am going to close and delete this PR since the issue was on my side.

PR #1708, opened 2023-05-16
fix: Finalize and write last state message with dedupe

### LINKED ISSUE #1704: bug: possible regression in state messages and progress_markers handling
### Singer SDK Version

0.23.0

### Is this a regression?

- [X] Yes

### Python Version

NA

### Bug scope

Taps (catalog, state, etc.)

### Operating System

linux

### Description

CI checks in https://github.com/MeltanoLabs/tap-github/pull/209 are now failing on a test related to `progress_markers`.

I've narrowed down the regression (?) to between sdk `v0.22.1` (pass) and `v0.23.0` (fail), and I think it is specifically https://github.com/meltano/sdk/pull/1436 which caused the regression. Tagging @aaronsteers as the author of that PR.

The test that fails was added to `tap-github` in [this PR](https://github.com/MeltanoLabs/tap-github/pull/164/files) after [this related PR](https://github.com/meltano/sdk/pull/1164) was merged in the SDK. In short, `progress_markers` [now appears](https://github.com/MeltanoLabs/tap-github/actions/runs/4978108124/jobs/8912840313?pr=209) in the last state message issued by the `sdk`, which used to signify that the message wasn't "final".

It looks like there are some changes in how state is handled, so I'm opening this as a discussion to understand if there was indeed a regression, or if the test in `tap-github` needs to be modified to accomodate the new behaviour in the sdk.

### Code

_No response_

kgpayne:
Closes #1704


aaronsteers:
@kgpayne - I did some digging and found that `Stream.finalize_state_progress_markers` actually calls a black box `finalize_state_progress_markers()`, which does not return a value.

https://github.com/meltano/sdk/blob/5ff3f4529b7e02831e38f2185df6342d06da21a7/singer_sdk/helpers/_state.py#L248-L270

Probably more scope than is needed, given that sending an extra `STATE` message is relatively harmless, but a long-term fix might migrate the helper function directly into the class - allowing it then to update the `Stream._is_state_flushed` property. Alternatively, the helper function could return a boolean value indicating "True" if changes are applied. Complicating matters (and expanding scope) is the fact that many write operations in the above helper function are not behind conditional checks.

Regarding performance of this operation, at least this operation is once-per-state-finalization-option and not once-per-record-write. Otherwise this would be more high-stakes code to update, performance wise. 🤷

kgpayne:
@aaronsteers the root cause was indeed missing `self._is_state_flushed = False` snippets. We are not marking state as flushable after calling `finalize_state_progress_markers`, which meant the final calls in stream and tap to `_write_state_message()` were skipped by the `if` statement in that method.

For now I was able to 'fix' the issue @laurentS is seeing in `tap-github` by i) setting `self._is_state_flushed = False` after every call to the `finalize_state_progress_markers` helper function (as state _may_ have changed) and ii) deduping in `Stream._write_state_message()` to eliminate duplicates. Its not elegant, but I wanted to verify I understood the root cause 🙂

Re: a refactor: we use the `finalize_state_progress_markers` helper function within the stream and the `Stream.finalize_state_progress_markers` from the Tap. Agree this could and probably should be refactored so that the `Stream.finalize_state_progress_markers` is the only one used, allowing us to toggle `_is_state_flushed` inline. Will keep hacking at this PR tomorrow.

FYI @edgarrmondragon - open to all feedback and suggestions 🙂

aaronsteers:
@kgpayne - Yeah, nice. Agreed with all the above. 👍

singer_sdk/streams/core.py
```diff
@@ -963,13 +967,27 @@ def reset_state_progress_markers(self, state: dict | None = None) -> None:
 
         reset_state_progress_markers(state)
 
+    def _finalize_state_progress_markers(
```
edgarrmondragon, line 970: Can we rename this? There's now three functions with the same name and signature, so it's hard to tell what each of them is doing:

* `singer_sdk.helpers._state.finalize_state_progress_markers`
* `singer_sdk.streams.core.Stream.finalize_state_progress_markers`
* `singer_sdk.streams.core.Stream._finalize_state_progress_markers`

singer_sdk/streams/core.py
```diff
@@ -980,10 +998,11 @@ def finalize_state_progress_markers(self, state: dict | None = None) -> None:
...
             for context in self.partitions or [{}]:
                 state = self.get_context_state(context or None)
-                finalize_state_progress_markers(state)
-            return
+                self._finalize_state_progress_markers(state)
+        else:
+            self._finalize_state_progress_markers(state)
 
-        finalize_state_progress_markers(state)
+        self._write_state_message()
```
edgarrmondragon, line 1005: Is it possible to call this outside of this scope. This method should _prepare_ the state to be emitted but it shouldn't actually write the message.

edgarrmondragon (changes requested):

laurentS:
@kgpayne @edgarrmondragon @aaronsteers is it worth bringing some form of the test that fails in `tap-github` upstream in the sdk to avoid regressions like this in the future? I'm thinking some simplified variant of [this one](https://github.com/MeltanoLabs/tap-github/pull/164/files#diff-4ad85a3deb7265065f80b3a0b6093a1dc32975536e497c12a5af0425f4d346c9R105).

edgarrmondragon:
> @kgpayne @edgarrmondragon @aaronsteers is it worth bringing some form of the test that fails in `tap-github` upstream in the sdk to avoid regressions like this in the future? I'm thinking some simplified variant of [this one](https://github.com/MeltanoLabs/tap-github/pull/164/files#diff-4ad85a3deb7265065f80b3a0b6093a1dc32975536e497c12a5af0425f4d346c9R105).

@laurentS I agree, we should take the opportunity to guard against future regressions 👍

singer_sdk/streams/core.py
edgarrmondragon: Nevermind

singer_sdk/streams/core.py
edgarrmondragon: Done in 9c353cce4b1b2d886d4463b9ef24630f932c76fe

edgarrmondragon (approved):
Ken's work is 👌 here. I only an automatic test so I'm confident merging.

PR #1733, opened 2023-05-24
feat: Test with Python 3.12 🐍

edgarrmondragon:
Blockers:

* [ ] #1858 [Pendulum](https://pypi.org/project/pendulum/) doesn't have wheels for 3.10+

----
:books: Documentation preview :books:: https://meltano-sdk--1733.org.readthedocs.build/en/1733/


edgarrmondragon:
Merging this with message `Test with Python 3.12` as:

- DuckDB (only used for testing) does not support Python 3.12 yet
- Pendulum can only be built on Python 3.12 (and has wheels) starting with prerelease 3.0.0b1

PR #1742, opened 2023-06-01
fix(templates): Recommend `meltano run` in target cookiecutter README 

pnadolny13:
----
:books: Documentation preview :books:: https://meltano-sdk--1742.org.readthedocs.build/en/1742/


edgarrmondragon (approved):
Thanks @pnadolny13!

PR #1745, opened 2023-06-02
refactor(taps): Change `SQLStream.schema` into a cached property

mjsqu:
Schema is called at least once per record and does not change within a stream, so it should be cached for performance benefits.

https://app.slack.com/client/TFG99TU9K/C013Z450LCD/thread/C013Z450LCD-1685703921.187269

----
:books: Documentation preview :books:: https://meltano-sdk--1745.org.readthedocs.build/en/1745/

### DIFF
--- a/singer_sdk/streams/sql.py
+++ b/singer_sdk/streams/sql.py
@@ -12,6 +12,7 @@ from singer_sdk._singerlib import CatalogEntry, MetadataMapping
 from singer_sdk.connectors import SQLConnector
 from singer_sdk.streams.core import Stream
 
+
 class SQLStream(Stream, metaclass=abc.ABCMeta):
     """Base class for SQLAlchemy-based streams."""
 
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
@mjsqu can you fix the failing mypy check? Some of the conversation in https://github.com/python/mypy/issues/5107#issuecomment-529372406 might be helpful.

Thanks!

mjsqu:
I had a quick attempt at this, but it didn't work particularly well - plus I think the proposed solution in the thread sets the lru_cache decorator to do nothing (`pass`), which removes the benefit?

I think I need to follow the example by @qbatten:
https://github.com/meltano/sdk/pull/1394/commits/0b5d6b0e4b07c93290f68b62e1f807f9e7217ccd

mjsqu:
@edgarrmondragon 🥳 

Thanks for the inspiration @qbatten

edgarrmondragon (approved):
Thanks @mjsqu!

PR #1749, opened 2023-06-05
fix: clear target test runner state after each run

pnadolny13:
The target test framework had weird behavior where all 14 default tests passed for me but when I looked closer only 1 table was getting created in the destination database. I figured out it was because the input file was getting stuck on the first test's input file.

The test overrides the runner `input_filepath` https://github.com/meltano/sdk/blob/fda24cfba259b917d0ab3a74b355346e7b0027e9/singer_sdk/testing/templates.py#L325 to pass in the path to the test's JSONL data but it only gets used on the first test because the second test uses the runner while it still has self._input state from the previous run so it [skips the logic](https://github.com/meltano/sdk/blob/fda24cfba259b917d0ab3a74b355346e7b0027e9/singer_sdk/testing/runners.py#LL241C17-L241C23) for reading in the new file. Ultimately all following tests get no input so they silently pass but dont actually do anything.

 I confirmed this fix worked in https://github.com/MeltanoLabs/target-snowflake/pull/38.

----
:books: Documentation preview :books:: https://meltano-sdk--1749.org.readthedocs.build/en/1749/

### DIFF
--- a/singer_sdk/testing/runners.py
+++ b/singer_sdk/testing/runners.py
@@ -270,6 +270,7 @@ class TargetTestRunner(SingerTestRunner):
         )
         self.stdout, self.stderr = (stdout.read(), stderr.read())
         self.state_messages.extend(self._clean_sync_output(self.stdout))
+        self.target_input = None
 
     def _execute_sync(
         self,


kgpayne:
@pnadolny13 @edgarrmondragon I think the issue and fix is a bit broader than resetting the `target_input` (though that obviously works).

A reset is needed because we made the `runner` instance a pytest fixture with `scope="class"`. This is desirable for Taps as it means they can run once (to hydrate the runners record/state/messages attributes) and then run multiple tests on the same record set. It was done to minimise the number of times the Taps actually has to run. However, with Target tests and especially with the file-based ones, we actually want to allow the Target to sync multiple times with the different file inputs. So I believe the change needed is to set the `runner` scope to `function` when constructing Target classes 🙂 This has a similar effect to resetting the runners input as each runner is 'fresh' and so doesn't have any previous state.

Will raise a PR in `target-snowflake` to test the fix.

pnadolny13:
Closing in favor of https://github.com/meltano/sdk/pull/1752 and I also found another bug for tests that follow a `pytest.raises(` we weren't getting a reset, I fixed that [in target-snowflake](https://github.com/MeltanoLabs/target-snowflake/pull/38/commits/dc18f25fd6ee2e3f4ffaeba362da1840a4f0aa1a) for now though.

PR #1752, opened 2023-06-06
fix: change runner scope to function for target tests

kgpayne:
----
:books: Documentation preview :books:: https://meltano-sdk--1752.org.readthedocs.build/en/1752/

### DIFF
--- a/samples/sample_target_csv/csv_target_sink.py
+++ b/samples/sample_target_csv/csv_target_sink.py
@@ -41,6 +41,7 @@ class SampleCSVTargetSink(BatchSink):
                 delimiter="\t",
                 quotechar='"',
                 quoting=csv.QUOTE_NONNUMERIC,
+                escapechar="\\",
             )
             for record in records_to_drain:
                 if newfile and not records_written:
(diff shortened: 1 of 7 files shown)


kgpayne:
This is now failing on legitimate target tests 😅🎉

kgpayne:
Related to #1749 and this conversation:

>@pnadolny13 @edgarrmondragon I think the issue and fix is a bit broader than resetting the `target_input` (though that obviously works).
>
> A reset is needed because we made the `runner` instance a pytest fixture with `scope="class"`. This is desirable for Taps as it means they can run once (to hydrate the runners record/state/messages attributes) and then run multiple tests on the same record set. It was done to minimise the number of times the Taps actually has to run. However, with Target tests and especially with the file-based ones, we actually want to allow the Target to sync multiple times with the different file inputs. So I believe the change needed is to set the `runner` scope to `function` when constructing Target classes 🙂 This has a similar effect to resetting the runners input as each runner is 'fresh' and so doesn't have any previous state.
>
> Will raise a PR in `target-snowflake` to test the fix.

edgarrmondragon (approved):

Issue #1689, opened 2023-05-10
feat: SkipAPIError to compliment RetriableAPIError and FatalAPIError


sebastianswms:
### Feature scope

Taps (catalog, state, stream maps, etc.)

### Description

Currently, the [`validate_response()`](https://github.com/meltano/sdk/blob/7430f778271abac272cef7ce98c6f148e6e08219/singer_sdk/streams/rest.py#L147) method in `rest.py` can throw two types of errors: `RetriableAPIError` and` FatalAPIError`. I propose a new type of error in addition to these two: `IgnorableAPIError`. This error would indicate that:

- The response being processed is invalid.
- The endpoint should _not_ be retried.
- The tap should _not_ fail.
- The record being processed should be skipped—no data is returned.
- The tap should continue executing.

If an override of `validate_response()` was to throw a `IgnorableAPIError`, the instance would merely be logged and no further action would be taken.

AutoIDM has had multiple use cases for this functionality before when writing taps for applications such as Google Ads or GitLab. In the past, we've done custom implementations, but it seems like something with wide enough applicability that it should be baked into the SDK.

edgarrmondragon:
Related:

* https://github.com/meltano/sdk/issues/280

edgarrmondragon:
Thanks for logging @sebastianswms! A few folks have requested this (see #280).

Would be willing to contribute it to the SDK?

edgarrmondragon:
Still relevant

Issue #1597, opened 2023-04-10
feat: Allow skipping child streams by returning an empty child context from parent stream


edgarrmondragon:
### Feature scope

Taps (catalog, state, stream maps, etc.)

### Description

Some pseudocodish approach would be:

```python
class ParentStream:
  # Custom tap implementation
  def get_child_context(self, record, context):
    if not record["has_children_records"]:

  # Part of the SDK
  def _sync_children(self, child_context):
    if not child_context:
      logger.warning(
        "Context for child streams of '%s' is empty, skipping sync of any child streams",
        self.name,
      )
      return
    for child_stream in self.child_streams:
      if child_stream.selected or child_stream.has_selected_descendents:
        child_stream.sync(context=child_context)
```

PR #1754, opened 2023-06-06
feat: Safely ignore certain API response codes

rawwar:
Implements #1689 

----
:books: Documentation preview :books:: https://meltano-sdk--1754.org.readthedocs.build/en/1754/


rawwar:
@edgarrmondragon , wondering if I should just call it `IgnorableException` instead of `IgnorableAPIResponseCodeException` so that other type of streams can also have some kind of ignore and continue mechanism. What do you think?

singer_sdk/exceptions.py
```diff
@@ -98,6 +98,20 @@ def __init__(self, message: str, response: requests.Response | None = None) -> N
...
+class IgnorableAPIResponseCodeException(Exception):
+    """Raised when an API response code is ignorable."""
+
+    def __init__(self, status_code: int | None = None) -> None:
+        """Extends the default with the ignored response's status code as an attribute.
+
+        Args:
+            status_code (int): Ignored status code.
+        """
+        message = f"Ignorable API response code: {status_code}"
```
laurentS, line 110: Hi @rawwar, thanks for implementing this. Looks like I tried over a year ago and never finished it for some mysterious reason :sweat_smile: I came back here looking for the same functionality again.

For `tap-github`, we are facing a bunch of situations where this would be very useful. But github works very hard to make their API a pain to use, so just mentioning the HTTP status code in the exception message isn't enough, it would be helpful to be able to include some extra text. To give a concrete example, trying to fetch the contributors list for the linux kernel, github returns:
`HTTP status 403 - Forbidden`
```json
{
  "message": "The history or contributor list is too large to list contributors for this repository via the API.",
  "documentation_url": "https://docs.github.com/rest/reference/repos#list-repository-contributors"
}
```
But in some other cases they also use that same error code for other problems. So I'd like to be able to write something like:
```python
raise IgnorableException(status_code=403, error_message="contributor list is too large")
```
That second argument could be optional, so it defaults to what you're suggesting above.

rawwar: Thank you. I'll do what you mentioned. Will be away for 4 days. But, once I am back.. will update the code

rawwar: updated

docs/classes/singer_sdk.exceptions.IgnorableException.rst
```diff
@@ -0,0 +1,8 @@
+singer_sdk.exceptions.IgnorableException
+=======================================================
```
edgarrmondragon, line 2:
```suggestion
singer_sdk.exceptions.IgnorableException
========================================
```

edgarrmondragon (changes requested):

edgarrmondragon (commented):
Hey @rawwar! Thanks for your continued contributions 😄 

## Naming

I feel like the purpose of this feature is to allow users to skip a stream _or partition_ if it's safe to do so. With that in mind, what do you think naming this something like `EndOfStreamError`?

That feels closer to how Python iterators use the `StopIteration` exception to signal they're consumed.

<details>

<summary>Iterator example</summary>

```python
class Iterator:
    def __init__(self, initial, max=10):
        self.current = initial
        self.max = max

    def __iter__(self):
        return self

    def __next__(self):
        if self.current < self.max:
            self.current += 1
            return self.current
        raise StopIteration


for i in Iterator(1):
    print(i)
```

</details>

## Make it agnostic

The status code feels too specific to REST streams, so wdyt about building the error message with the status code instead of passsing it as an argument.

## Skip partitions too?

The current implementation doesn't allow developers to skip a partition. This use case has come up a few times when a child stream partition should not be synced based on the parent record (https://github.com/meltano/sdk/issues/1597).

## Testing

With the above consideration in mind, we could use a snapshot test to validate that the partition/stream is skipped:

<details>

<summary>Unit test</summary>

```python
import io
from contextlib import redirect_stdout

import pytest
from freezegun import freeze_time

from singer_sdk.exceptions import IgnorableException
from singer_sdk.streams.core import Stream
from singer_sdk.tap_base import Tap


@freeze_time("2022-01-01T00:00:00Z")
@pytest.mark.snapshot()
def test_end_of_stream_error(snapshot):
    """Test end of stream error is raised."""
    snapshot_name = "end_of_stream_error.jsonl"

    class CustomTap(Tap):
        name = "custom_tap"

        def discover_streams(self):
            return [StreamWithPartitions(self)]

    class StreamWithPartitions(Stream):
        name = "partitions"
        partitions = [
            {"partition": "x"},
            {"partition": "y"},
            {"partition": "z"},
        ]
        schema = {
            "type": "object",
            "properties": {
                "id": {"type": "integer"},
                "name": {"type": "string"},
                "partition": {"type": "string"},
            },
        }

        def get_records(self, context):
            partition = context["partition"]

            if partition == "y":
                msg = "End of stream"
                raise IgnorableException(msg)

            yield {
                "id": 1,
                "name": "foo",
                "partition": partition,
            }
            yield {
                "id": 2,
                "name": "bar",
                "partition": partition,
            }

    tap = CustomTap(config=None)

    buf = io.StringIO()
    with redirect_stdout(buf):
        tap.sync_all()

    buf.seek(0)
    snapshot.assert_match(buf.read(), snapshot_name)
```

</details>

rawwar:
@edgarrmondragon , After thinking about what you mentioned, here are few doubts I have.. 
- 

> what do you think naming this something like EndOfStreamError
If we make this generic, how will developer consider ignoring only specific types of error codes? For example, If I choose to ignore 5xx errors and not any 4xx, Do I depend on catching the exception and checking its status code or some instance variable to decide how to move forward?

> The status code feels too specific to REST streams,
Sure. But, is it better to have levels of Exceptions or one general exception? Like, Having a base `IgnorableException` and use `IgnorableAPIException` as the classes for REST specific streams?

edgarrmondragon:
> For example, If I choose to ignore 5xx errors and not any 4xx, Do I depend on catching the exception and checking its status code or some instance variable to decide how to move forward?

@rawwar No, the tap developer would rather control when the exception _is raised_. Rather than catching it and deciding what do with it, the SDK handles and logs the exception then continues to the next stream or partition.

A user may want to do the following mapping:

* 5xx => RetriableAPIError, the server is flaky
* 401 => FatalAPIError, the access token is not valid
* 400 => IgnorableAPIError, bad request (maybe cause by the parent entity having no children)

The RESTStream class knows to retry the first and raise the second. With the changes here, the stream and tap would know to handle the third one gracefully so the program doesn't crash.

> Sure. But, is it better to have levels of Exceptions or one general exception? Like, Having a base `IgnorableException` and use `IgnorableAPIException` as the classes for REST specific streams?

Yeah, we could totally do that :+1:

PR #1769, opened 2023-06-16
feat: Validate parsed/transformed record against schema message

### LINKED ISSUE #1147: [Feature]: SQLTarget - Record missing key property should fail nicely
### Feature scope

### Description

https://github.com/MeltanoLabs/target-postgres/blob/4f677d3f30ceaf8415dd1a401d22ab8aaf06e2a2/target_postgres/tests/test_standard_target.py#L172 is the test

We should fail nicely, this test was taken from https://github.com/MeltanoLabs/target-sqlite/blob/master/tests/test_target_sqlite.py#L44

kgpayne:
Closes #1147


singer_sdk/testing/target_tests.py
```diff
@@ -108,6 +110,11 @@ class TargetRecordMissingKeyProperty(TargetFileTestTemplate):
 
     name = "record_missing_key_property"
 
+    def test(self) -> None:
+        """Run test."""
+        with pytest.raises(MissingKeyPropertiesError):
+            super().test()
```
edgarrmondragon, line 116: I'm a bit confused by this. Does it mean that the test succeeds if the record is missing a key property?

kgpayne: @edgarrmondragon this tests that the SDK implementation correctly raises a `MissingKeyPropertiesError` if the Tap sends a `RECORD` with missing key properties 🙂 Previously we were throwing a generic `Exception`, so this just makes the error more user-friendly.

edgarrmondragon (approved):

PR #1770, opened 2023-06-16
fix(targets): Check schema has arrived before any record

### LINKED ISSUE #1146: [Feature]: SQLTarget Record before Schema should fail nicely
### Feature scope

### Description

https://github.com/MeltanoLabs/target-postgres/blob/main/target_postgres/tests/test_standard_target.py#L64 is the record_before_schema test

We should fail nicely, this test was taken from https://github.com/MeltanoLabs/target-sqlite/blob/master/tests/test_target_sqlite.py#L23-L32

kgpayne:
Closes #1146


kgpayne:
cc @visch

edgarrmondragon (approved):

visch:
> cc @visch

Looks good I'll get an issue in to migrate `target-postgres` once this is released

Issue #1774, opened 2023-06-19
bug: anyOf sql types not parsed properly


pnadolny13:
This json type is returning a VARCHAR sql type when it should be returning a DATETIME:

`{"anyOf": [{"type": "string", "format": "date-time"}, {"type": "null"}]}`

It looks like the bug is in the fact that [to_sql_type](https://github.com/meltano/sdk/blob/e8772158aa7b4a0f87fb280d40bba68f8c0e72e3/singer_sdk/typing.py#LL845C8-L845C30) is calling [_jsonschema_type_check](https://github.com/meltano/sdk/blob/e8772158aa7b4a0f87fb280d40bba68f8c0e72e3/singer_sdk/typing.py#L809) to check if its considered a string and its returning false because the [anyof logic](https://github.com/meltano/sdk/blob/e8772158aa7b4a0f87fb280d40bba68f8c0e72e3/singer_sdk/typing.py#L828) is looking for the `{"type": "string", "format": "date-time"}` in `("string",)` when really it should be calling itself again for each anyof item.

PR #1775, opened 2023-06-19
fix: Add tests for SQL type conversion from JSON schemas

pnadolny13:
Closes https://github.com/meltano/sdk/issues/1774

I added tests that I let run in CI and they fail https://github.com/meltano/sdk/actions/runs/5315601259/jobs/9624162264 then I pushed the change that fixes them.

----
:books: Documentation preview :books:: https://meltano-sdk--1775.org.readthedocs.build/en/1775/

### DIFF
--- a/tests/core/test_typing.py
+++ b/tests/core/test_typing.py
@@ -295,24 +295,31 @@ def test_conform_primitives():
     assert _conform_primitive_property(1, {"type": "boolean"}) is True
 
 
-
-import sqlalchemy
 import pytest
+import sqlalchemy
+
+
 @pytest.mark.parametrize(
     "jsonschema_type,expected",
     [
-        ({'type': ['string', 'null']}, sqlalchemy.types.VARCHAR),
-        ({'type': ['integer', 'null']}, sqlalchemy.types.INTEGER),
-        ({'type': ['number', 'null']}, sqlalchemy.types.DECIMAL),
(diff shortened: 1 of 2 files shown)


tests/core/test_typing.py
```diff
@@ -292,3 +293,33 @@ def test_conform_primitives():
     assert _conform_primitive_property(None, {"type": "boolean"}) is None
     assert _conform_primitive_property(0, {"type": "boolean"}) is False
     assert _conform_primitive_property(1, {"type": "boolean"}) is True
+
+
+import pytest
+import sqlalchemy
+
```
edgarrmondragon, line 300: @pnadolny13 can you move these imports to the top of the file?

pnadolny13: @edgarrmondragon yep sorry I forgot to clean this up. Fixing now

tests/core/test_typing.py
```diff
@@ -6,6 +6,8 @@
 import logging
 import typing as t
 
+import sqlalchemy
```
edgarrmondragon, line 9:
```suggestion
import pytest
import sqlalchemy
```

and you'll also need to remove the `if t.TYPE_CHECKING:` block

edgarrmondragon (changes requested):

edgarrmondragon (approved):

edgarrmondragon:
This is great, thanks @pnadolny13!

Issue #1777, opened 2023-06-20
bug: target existing sink check returns incorrect boolean when `add_metadata_columns=True`


pnadolny13:
The way that the [get_sink method](https://github.com/meltano/sdk/blob/88d7586e538a0a38547757e11a8347ef6dd93ac0/singer_sdk/target_base.py#L169) does a comparison between the existing sink schema and the incoming singer message schema returns an incorrect boolean when add_metadata_columns config is true. The base sink [automatically adds](https://github.com/meltano/sdk/blob/88d7586e538a0a38547757e11a8347ef6dd93ac0/singer_sdk/sinks/core.py#L71) the metadata columns when the sink is initialized so the sink's schema is no longer matching the original singer message schema that was provided. If the target receives another exact same schema message in the same sync it will do a comparison between the singer schema and the sink schema that now has metadata columns, it will always decide to drain and recreate the sink because it thinks the schema has changed.

For some taps that emit lots of schema messages, I think when child streams are implemented, i.e. tap-github the target becomes super inefficient because its constantly draining and recreating sinks even if they only have a few records and the schema hasnt changed.

Originally discovered during target-snowflake development and was fixed https://github.com/MeltanoLabs/target-snowflake/blob/86ca8067636d0c82350f7868785fcc998e113dcb/target_snowflake/target.py#L89 by removing metadata columns prior to doing the comparison.

pnadolny13:
I have a PR on the way that should fix this.

PR #1778, opened 2023-06-20
fix: Sink schema comparison before adding metadata columns

pnadolny13:
Closes https://github.com/meltano/sdk/issues/1777

- moves test mock classes from the csv test to conftest so its reusable. Let me know if this should go somewhere else.
- adds test for target_base get_sink method to assert that when add_metadata_columns is used it doesnt affect the comparison. The [tests fail in CI](https://github.com/meltano/sdk/actions/runs/5326800321/jobs/9649332587) at first, then pass once I added a fix
- I chose to fix it by adding an original_schema attribute to the sink object. The alternative, and what I did in target-snowflake originally, is to add logic to remove those columns before doing the comparison but that sounded like more work.

@edgarrmondragon let me know what you think!

----
:books: Documentation preview :books:: https://meltano-sdk--1778.org.readthedocs.build/en/1778/

### DIFF
--- a/singer_sdk/sinks/core.py
+++ b/singer_sdk/sinks/core.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import abc
+import copy
 import datetime
 import json
 import time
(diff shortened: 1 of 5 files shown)


edgarrmondragon (approved):
This is great @pnadolny13 😁

Issue #1046, opened 2022-10-06
[Feature]: Faster JSON dumps in `format_message` with an alternative JSON SerDe library


s7clarke10:
### Feature scope

Taps (catalog, state, stream maps, etc.)

### Description

Speed of writing messages can add up when you are dumping a large amount of data. One of the things in the original pipelinewise-singer-python was the use of orjson for doing json dumps as it is a lot faster.

The particular line in particular is this. 

https://github.com/meltano/sdk/blob/796ccda470529253b5211b70742c814d2b20ec07/singer_sdk/_singerlib/messages.py#L165

We have a variant of pipelinewise-singer-python which uses orjson in favour of json.dumps for improved speed. The code for that section is as follows.

```python
import orjson
import decimal

...

def format_message(message, option=0):
    def default(obj):
        if isinstance(obj, decimal.Decimal):
            return str(obj)
        raise TypeError
    
    return orjson.dumps(message.asdict(), option=option, default=default)
```

Sorry the formatting is all out, I couldn't work out how to get the code formatting working. 
Best look at the example below.
https://github.com/mjsqu/pipelinewise-singer-python/blob/2a839d245b92d1ec6d5801a30f5280fa77b22d0c/singer/messages.py#L295

I note that while orjson is faster it is not perfect, for example we need to convert decimal data as a string. There are some other considerations mentioned with the module but if they are not an issue, we have measured a considerable performance gain using this and verified this by the cProfile to look at the time consumed.

edgarrmondragon:
> One of the things in the original pipelinewise-singer-python was the use of orjson for doing json dumps as it is a lot faster.

That's right, although the SDK never used that since it depended on an older version of `pipelinewise-singer-python` (see https://github.com/meltano/sdk/pull/738) and there was a bug in orjson that was a deal-breaker for us (see https://github.com/ijl/orjson/issues/251). Seems to have been fixed, but `pipelinewise-singer-python` is stuck with an older version of orjson 🤷‍♂️.

But, after #979, we can perhaps explore using `orjson` directly.

s7clarke10:
That's great feedback @edgarrmondragon, also thanks for taking the time to provide the back story re orjson.

mjsqu:
This has been updated in my fork of `pipelinewise-singer-python`:

Method is `BatchMessage.format_message`

```
def format_message(message, option=0):
    def default(obj):
        if isinstance(obj, decimal.Decimal):
[...]

Issue #1659, opened 2023-05-02
Drop support for Python 3.7


edgarrmondragon:
- Update requirement in `pyproject.toml` and run `poetry lock`
- Remove from trove classifiers in `pyproject.toml`
- Update Ruff config in `pyproject.toml`
- Address any `sys.version_info <= (3, 8)` not auto-fixed by Ruff
- Repeat above steps for each of the cookiecutter templates

edgarrmondragon:
Closed by https://github.com/meltano/sdk/pull/1837

PR #1784, opened 2023-06-22
feat: Faster JSON dumps in `format_message` with `msgspec`

BuzzCutNorman:
This is a draft PR to assist discussion and testing related to Issue #1046

msgspec requires python 3.8 and above.  PR #2143 once merged will drop support for python 3.7 from then SDK.

----
:books: Documentation preview :books:: https://meltano-sdk--1784.org.readthedocs.build/en/1784/


s7clarke10:
Thanks for pulling this together @BuzzCutNorman .

s7clarke10:
If msgspec is faster and is now supporting the standard output of decimals correctly now, I would like to throw my support behind this PR. The author of msgspec has written the decimal support in specifically for Meltano at our request. The output should be the same as the standard json dumps but a lot faster.
There is an option to output decimal as strings as required for fuller support by certain targets and/or ensuring precision - thinking the informal Singer Decimal specification.

I know a number of the taps that myself and @mjsqu maintain have the ability to output in either decimal form or strings based on a setting and have hints when singer decimal records are used for the target to decode the string decimal.

edgarrmondragon:
@BuzzCutNorman thanks you so much for taking to benchmark the JSON serde library alternatives and creating this PR (and the others)!

This seems to require us dropping support for Python 3.7, or doing the work of adding the logic to use `simplejson` in 3.7 and `msspec` in all newer versions but that seems unnecessarily complicated.

We've already planned to drop 3.7 (#1659, #1837) for the v1 release, so this PR would have to be merged after that.

The timeline for that is still unclear but I'll update you as soon as things start moving faster in that direction.

BuzzCutNorman:
@edgarrmondragon A little more time to refine this I think is a good thing.  There are some items that I need some assistance with.

1) msgspec wants us to reuse an Encoder.  I created it where I found it worked but I don't know if that is the best place for it.  https://jcristharif.com/msgspec/perf-tips.html#reuse-encoders-decoders

2) msgspec does return the encoded json as a bytes string.  Currently I am decoding it to a str but I also found you can point that result to stdout.buffer and the target side is able to read from stdin as a string.  I don't know what scenario is best to go with.

edgarrmondragon:
> @edgarrmondragon A little more time to refine this I think is a good thing. There are some items that I need some assistance with.
> 
> 1. msgspec recommends reusing an Encoder.  I created it where I found it worked but I don't know if that is the best place for it.  [jcristharif.com/msgspec/perf-tips.html#reuse-encoders-decoders](https://jcristharif.com/msgspec/perf-tips.html#reuse-encoders-decoders)

@BuzzCutNorman I think it makes sense where it is now. There are other places where we encode JSON data, e.g. for JSON lines batches but I don't think they're in scope for now.

> 2. msgspec does return the encoded json as a bytes string.  Currently I am decoding it to a str but I also found you can write json bytes strting to stdout.buffer and the target side is able to read from stdin as a string.  I don't know what scenario is best to go with.

I think going with the raw bytes produced by msspec is probably better, if just for the slight performance penalty of decoding before writing to stdout but also because it's likely safer.

BuzzCutNorman:
@edgarrmondragon I updated `format_message` to return a `bytes` string and `write_message` to write to ` sys.stdout.buffer`.  The pytest tests and testing framework were updated to allow for `sys.stdout.buffer.write`.  In `io_base.py` the `SingerReader` class method `listen` was updated to pull from `sys.stdin.buffer` and `deserialize_json` updated to utilize `msgspec` to deserialize lines.  I have not found a way to reproduce the `parse_float` option currently used as of yet.

BuzzCutNorman:
I was able to change `float` to `Decimal` using `msgspec`.  The key is you need to generate a `Struct` from `SCHEMA` messages.  The code is not elegant but working.  If anyone has time, please take a look and let me know what you think.

s7clarke10:
Hi @BuzzCutNorman ,

I have added msgspec to the traditional singer-python frameworks used by some non-SDK taps. I am seeing some good performance from doing this.  https://github.com/s7clarke10/pipelinewise-singer-python

BuzzCutNorman:
Hi @s7clarke10 ,

That is great news.  Thanks for sharing how you are using `msgspec` in singer-python.  I will defiantly be looking it over to see how you leveraged `msgspec`.

s7clarke10:
> Hi @s7clarke10 ,
> 
> That is great news. Thanks for sharing how you are using `msgspec` in singer-python. I will defiantly be looking it over to see how you leveraged `msgspec`.

For testing this I have created a branch on [pipelinewise-tap-oracle](https://github.com/s7clarke10/pipelinewise-tap-oracle/tree/feature/use_msgspec_serializer) to use msgspec.

Of note is the use of a boolean config setting use_singer_decimal in the tap. As a user of a tap, I have a choice whether I wish to have decimal and float data emitted as a string - `"use_singer_decimal": true` (to avoid precision loss), or in their native formats for backwards compatibility. It seems to work well, and produces the same results when I was using orjson.

If that setting is omitted it will out default to outputting decimals and floats in their lossy native format - important for backwards compatibility.

BuzzCutNorman:
@edgarrmondragon, I have a question on how I should deal with the SCHEMA found in  `singer_sdk\testing\target_test_streams\optional_attributes.singer` .  In the SCHEMA message there is the `required` field which has a list of fields.  Should I assume the fields not listed could be `null`.  If that is the case should I also to update the SCHEMA message's `properties` for `optional` from `"optional": {"type": "string"}` to `"optional": {"type": ["string","null"]}` before passing it on to the target?

edgarrmondragon:
> @edgarrmondragon, I have a question on how I should deal with the SCHEMA found in `singer_sdk\testing\target_test_streams\optional_attributes.singer` . In the SCHEMA message there is the `required` field which has a list of fields. Should I assume the fields not listed could be `null`?

TL;DR yes!

Strictly speaking there's probably a distinction to be made between _missing_ and _null_ fields, e.g. a string field could be missing but not nullable but for the purposes of moving data it's equivalent and better communicated by the Singer catalog's inclusion and selection metadata.

BuzzCutNorman:
@edgarrmondragon 
> Strictly speaking there's probably a distinction to be made between missing and null fields, e.g. a string field could be missing but not nullable 

I used the [UNSET](https://jcristharif.com/msgspec/supported-types.html#unset) type to achieve that distinction.  During the runtime creation of a struct from the schema message I set all fields that could be null or missing have a default value of `UNSET`.  I then remove all items in the record dictionary that have UNSET as a value.  While I use `t.Optional` to union the python data type with null so null values are allowed.

BuzzCutNorman:
@edgarrmondragon Yesterday I finished updating the `decoder` in `io_base.py` to use the new `float_hook` argument that was added in msgspec 0.18.0.  The `float_hook` addition drastically reduced the amount of code changes needed in the SingerReader class to utilize msgspec.

BuzzCutNorman:
@edgarrmondragon, I was wondering since there is a `SingerReader` class in `singer_sdk.io_base.` What do you think of a `SingerWriter` class being added to that file and moving  the msgspec `encoder`, `write_message()`, and `format_message` there from `singer_sdk._singerlib.messages`.   All the `stdin` and `stdout` methods, msgspec `encoder`, and msgspec `decoder` would be in one file.  Then `SingerWriter` could be added as an inherited class of `Tap` just like `SingerReader` is an inherited class of `Target`.  The additions to `singer_sdk.io_base` might something like this.
```
import msgspec
from singer_sdk._singerlib import Message, SingerMessageType

encoder = msgspec.json.Encoder(enc_hook=str, decimal_format="number")
msg_buffer = bytearray(64)

class SingerWriter:
    """Interface for all plugins writting Singer messages to stdout."""

    def format_message(self, message: Message) -> bytes:
        """Format a message as a JSON string.

        Args:
            message: The message to format.

        Returns:
            The formatted message.
        """
        encoder.encode_into(message.to_dict(), msg_buffer)
        msg_buffer.extend(b"\n")
        return msg_buffer

    def write_message(self, message: Message) -> None:
        """Write a message to stdout.

        Args:
            message: The message to write.
        """
        sys.stdout.buffer.write(self.format_message(message))
        sys.stdout.flush()
```

edgarrmondragon:
> @edgarrmondragon, I was wondering since there is a `SingerReader` class in `singer_sdk.io_base.` What do you think of a `SingerWriter` class being added to that file and moving the msgspec `encoder`, `write_message()`, and `format_message` there from `singer_sdk._singerlib.messages`. All the `stdin` and `stdout` methods, msgspec `encoder`, and msgspec `decoder` would be in one file. Then `SingerWriter` could be added as an inherited class of `Tap` just like `SingerReader` is an inherited class of `Target`. The additions to `singer_sdk.io_base` might something like this.
> 
> ```
> import msgspec
> from singer_sdk._singerlib import Message, SingerMessageType
> 
> encoder = msgspec.json.Encoder(enc_hook=str, decimal_format="number")
> msg_buffer = bytearray(64)
> 
> class SingerWriter:
>     """Interface for all plugins writting Singer messages to stdout."""
> 
>     def format_message(self, message: Message) -> bytes:
>         """Format a message as a JSON string.
> 
>         Args:
>             message: The message to format.
> 
>         Returns:
>             The formatted message.
>         """
>         encoder.encode_into(message.to_dict(), msg_buffer)
>         msg_buffer.extend(b"\n")
>         return msg_buffer
> 
>     def write_message(self, message: Message) -> None:
>         """Write a message to stdout.
> 
>         Args:
>             message: The message to write.
>         """
>         sys.stdout.buffer.write(self.format_message(message))
>         sys.stdout.flush()
> ```

@BuzzCutNorman Yeah, I like the symmetry of that 👍

BuzzCutNorman:
> @BuzzCutNorman Yeah, I like the symmetry of that 👍
@edgarrmondragon I just finished adding the `SingerWriter` class to `io_base`.  Please take a look when you get a chance.

edgarrmondragon:
> > @BuzzCutNorman Yeah, I like the symmetry of that 👍
> 
> @edgarrmondragon I just finished adding the `SingerWriter` class to `io_base`. Please take a look when you get a chance.

@BuzzCutNorman If it's not too much effort, would be willing to split that into it's own PR? Would make it easier to review both this PR and that change alone, and allow us to ship it earlier.

BuzzCutNorman:
>>> @BuzzCutNorman Yeah, I like the symmetry of that 👍
>>
>>@edgarrmondragon I just finished adding the SingerWriter class to io_base. Please take a look when you get a chance.
>
>@BuzzCutNorman If it's not too much effort, would be willing to split that into it's own PR? Would make it easier to review both this PR and that change alone, and allow us to ship it earlier.

@edgarrmondragon I split out of the SingerWriter related changes into PR #2058.

BuzzCutNorman:
@edgarrmondragon I believe this PR is passing all the tests now.  I did have to setup `xfail` for a few legacy tests.  They run fine alone but fail when run with all the tests.  I know this PR cannot be added until python 3.7 support is dropped so when you think it best, please give it a review.

BuzzCutNorman:
@edgarrmondragon Yay!! 🥳there are lightning bolts ⚡ in the [CodSpeed Performace Analysis](https://github.com/meltano/sdk/pull/1784/checks?check_run_id=19458784139) report for the new JSON encode decode test.

BuzzCutNorman:
@edgarrmondragon I merged SDK 0.35.0a3 alpha which had PR #2143, which dropped Python 3.7, merged into it.  That was the only perquisite for this PR.

pyproject.toml
```diff
@@ -51,6 +51,7 @@ joblib = ">=1.0.1"
 jsonpath-ng = ">=1.5.3"
 jsonschema = ">=4.16.0"
 memoization = { version = ">=0.3.2,<0.5.0", python = "<4" }
+msgspec = "^0.18.6"
```
edgarrmondragon, line 54:
```suggestion
msgspec = ">=0.18,<0.19"
```

singer_sdk/helpers/_compat.py
```diff
@@ -4,6 +4,8 @@
 
 import datetime
 import sys
+from importlib import metadata
+from typing import final  # noqa: ICN003
```
edgarrmondragon, line 8: We can get rid of these

singer_sdk/helpers/_compat.py
```diff
@@ -32,6 +34,8 @@
 time_fromisoformat = datetime.time.fromisoformat
 
 __all__ = [
+    "metadata",
+    "final",
```
edgarrmondragon, line 38: Ditto

singer_sdk/mapper_base.py
```diff
@@ -112,7 +112,7 @@ def invoke(  # type: ignore[override]
         about: bool = False,
         about_format: str | None = None,
         config: tuple[str, ...] = (),
-        file_input: t.IO[str] | None = None,
+        file_input: t.BinaryIO | None = None,
```
edgarrmondragon, line 115:
```suggestion
        file_input: t.IO[bytes] | None = None,
```

is more generic

singer_sdk/target_base.py
```diff
@@ -540,7 +540,7 @@ def invoke(  # type: ignore[override]
         about: bool = False,
         about_format: str | None = None,
         config: tuple[str, ...] = (),
-        file_input: t.IO[str] | None = None,
+        file_input: t.BinaryIO | None = None,
```
edgarrmondragon, line 543:
```suggestion
        file_input: t.IO[bytes] | None = None,
```

singer_sdk/helpers/_compat.py
BuzzCutNorman: I will give it a try.  I removed them when I merged in SDK 0.35.0a3 alpha and pytest was not happy.

BuzzCutNorman: Yep, pytest blew up again.  I will sync the changes so you can take a look.

BuzzCutNorman: I think I see the issue now.  It is in io_base `SingerReader.listen` is using `@Final` and I was importing it from `_compact`

BuzzCutNorman: I changed `@Final` to `@t.final` and `pytest` is happy 😃 now 🥳.

BuzzCutNorman:
@edgarrmondragon I found where I caused the issue in `io_base.py`.  I changed `@Final` decorator for `SingerWriter.listen` to `@t.final` and `pytest` is happy 😃 now 🥳

tests/samples/test_target_sqlite.py
```diff
@@ -521,6 +527,10 @@ def test_sqlite_generate_insert_statement(
     assert dml == expected_dml
 
 
+@pytest.mark.xfail(
+    raises=msgspec.DecodeError,
+    reason="Passes when run by itself, Always fails when run with all tests.",
+)
```
edgarrmondragon, line 533: This smells of dependencies between tests. @BuzzCutNorman can you create an issue for us to address this? 🙏

BuzzCutNorman: @edgarrmondragon I open an issue as requested, it is #2287.

edgarrmondragon: Thanks!

s7clarke10:
Just curious, what happened with the msgspec work to speed up serialization?

edgarrmondragon:
> Just curious, what happened with the msgspec work to speed up serialization?

I still need to come up with a plan to add this with the least amount of disruption, maybe starting with opt-in.

BuzzCutNorman:
> > Just curious, what happened with the msgspec work to speed up serialization?
> 
> I still need to come up with a plan to add this with the least amount of disruption, maybe starting with opt-in.

@edgarrmondragon,  I am happy to help break this PR into smaller PRs that can be merged in any order you would like.  Please just let me know how I can assist with the process.

edgarrmondragon:
> > > Just curious, what happened with the msgspec work to speed up serialization?
> > 
> > 
> > I still need to come up with a plan to add this with the least amount of disruption, maybe starting with opt-in.
> 
> @edgarrmondragon, I am happy to help break this PR into smaller PRs that can be merged in any order you would like. Please just let me know how I can assist with the process.

Thanks @BuzzCutNorman! I'll take some time this week to go through this and #2259 to see where it makes sense to split and have a reasonable package layout.

BuzzCutNorman:
@edgarrmondragon I am thinking it is time to close this PR since you have two very good PRs in the works and to be honest this PR is a bit of a mess. WDYT?

edgarrmondragon:
I'm happy to close but give you proper credit in what ends up getting merged. Thanks @BuzzCutNorman!

Issue #915, opened 2022-08-18
[Feature]: Make samples in the repo into runnable packages


edgarrmondragon:
### Feature scope

Other

### Description

This implies adding `__main__.py` to the existing samples.

Doing so would make manually testing the samples much simpler:

```shell
poetry shell
python samples/tap_sample --config=ENV | python samples/sample_target --config=ENV
```

Instead of having to call the tap/target module (which may or may not contain `if __name__ ...`) explicitly:

```shell
poetry shell
python samples/tap_sample/tap.py --config=ENV | python samples/sample_target/target.py --config=ENV
```

Saves a few keystrokes and may also help with debugging the samples (see #902).

edgarrmondragon:
This is easy to do if anyone's interested. Simply add a `__main__.py` file to each sample directory with this contents

```python
from <sample name> import <Sample Plugin Class>

<Sample Plugin Class>.cli()
```

edgarrmondragon:
Closed by https://github.com/meltano/sdk/pull/2156

PR #1800, opened 2023-06-29
feat: publish supported python versions in `--about`

### LINKED ISSUE #1403: feat: support reading standard `__version__` module attr when reporting `--about` version
### Feature scope

CLI (options, error messages, logging, etc.)

### Description

Python packaging is messy, and so our implementation of sourcing the plugin version (to report in `--about`) from package metadata does not work in several scenarios (discovered in this [PR](https://github.com/meltano/sdk/pull/1254)). It would help if our implementation would check the standard module `__version__` attribute before trying other means to infer a version.

`poetry-dynamic-versioning` supports substituting the value of `__version__` by default, so this would be the primary mechanism used in MeltanoLabs variants for reporting the plugins version.

e.g.

```python
# tap-snowflake/__init__.py
__version__ = "0.0.0"  # poetry-dynamic-versioning will replace this value on release

### LINKED ISSUE #1628: feat: --about to list the python versions supported by the connector
It would be great if SDK plugins would advertise their supported python versions. This way we can scrape them for:

- listing on the hub
- analyzing the ecosystem i.e. identifying taps that should be upgraded to support newer python versions or remove support for deprecated versions

I'm not exactly sure what format this should be in but something like whats in the pypoetry.toml would probably work `python = "<3.12,>=3.7.1"`

cc @WillDaSilva

### PR DESCRIPTION
**Pull request by kgpayne**
Closes #1628
Closes #1403

### DIFF
--- a/singer_sdk/plugin_base.py
+++ b/singer_sdk/plugin_base.py
@@ -209,19 +209,17 @@ class PluginBase(metaclass=abc.ABCMeta):
         Returns:
             A list of supported Python versions.
         """
-        package_metadata = {}
-        supported_python_versions = []
-
         try:
             package_metadata = metadata.metadata(package)
         except metadata.PackageNotFoundError:
             return None
 
         reported_python_versions = SpecifierSet(package_metadata["Requires-Python"])
-        for version in CHECK_SUPPORTED_PYTHON_VERSIONS:
-            if version in reported_python_versions:
-                supported_python_versions.append(version)
-        return supported_python_versions
+        return [
+            version
+            for version in CHECK_SUPPORTED_PYTHON_VERSIONS
+            if version in reported_python_versions
+        ]
 
     @classmethod
     def get_plugin_version(cls) -> str:
--- a/singer_sdk/plugin_base.py
+++ b/singer_sdk/plugin_base.py
@@ -56,6 +56,9 @@ CHECK_SUPPORTED_PYTHON_VERSIONS = (
     # future supported versions
     "3.12",
     "3.13",
+    "3.14",
+    "3.15",
+    "3.16",
 )
 
 
--- a/singer_sdk/plugin_base.py
+++ b/singer_sdk/plugin_base.py
@@ -207,7 +207,7 @@ class PluginBase(metaclass=abc.ABCMeta):
         """
         try:
             supported_python_versions = []
-            package_metadata = metadata(package)
+            package_metadata = metadata.metadata(package)
             reported_python_versions = SpecifierSet(package_metadata["Requires-Python"])
             for version in CHECK_SUPPORTED_PYTHON_VERSIONS:
                 if version in reported_python_versions:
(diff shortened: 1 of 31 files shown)

### CODE REVIEW

File: singer_sdk/plugin_base.py [hunk 1]
```diff
@@ -44,7 +65,10 @@
 class PluginBase(metaclass=abc.ABCMeta):
     """Abstract base class for taps."""
 
-    name: str  # The executable name of the tap or target plugin.
+    name: str  # The executable name of the tap or target plugin. e.g. tap-foo
+    package_name: str | None = (
+        None  # The package name of the plugin. e.g meltanolabs-tap-foo
+    )
```
WillDaSilva, line 71: As a hack day project, maybe we could implement something that determines the package name given the executable name. As far as I know there's no easy way to do this, but I think it's still possible.

singer_sdk/plugin_base.py
```diff
@@ -171,14 +195,36 @@ def _get_package_version(package: str) -> str:
...
+        Args:
+            package: The package name.
+
+        Returns:
+            A list of supported Python versions.
+        """
+        try:
+            supported_python_versions = []
+            package_metadata = metadata.metadata(package)
+            reported_python_versions = SpecifierSet(package_metadata["Requires-Python"])
```
WillDaSilva, line 211: Any idea why, despite the updated test, Codecov is not seeing this line as covered?

WillDaSilva (approved):

singer_sdk/plugin_base.py
kgpayne: Our tests for `--about` test the formatters with example data - we don't have a test for how the data is sourced. I suspect this is because of the complexity of installing example packages to test against. Will look at how easy this is to do with pytest as I agree it would be good to test the sourcing side 👍

kgpayne:
@edgarrmondragon FYI. We talked about this in our ☕️ chat, and you mentioned using a different approach in `tap-linkedin`? How do you feel about adding an optional `package_name` attr to facilitate lookups of the package version and other package metadata?

singer_sdk/plugin_base.py
```diff
@@ -44,7 +68,10 @@
 class PluginBase(metaclass=abc.ABCMeta):
     """Abstract base class for taps."""
 
-    name: str  # The executable name of the tap or target plugin.
+    name: str  # The executable name of the tap or target plugin. e.g. tap-foo
+    package_name: str | None = (
+        None  # The package name of the plugin. e.g meltanolabs-tap-foo
+    )
```
edgarrmondragon, line 74: Can we make this so they'll be documented in the docs site?

```suggestion
    #: The executable name of the tap or target plugin. e.g. tap-foo
    name: str

    #: The package name of the plugin. e.g meltanolabs-tap-foo
    package_name: str | None = None
```

edgarrmondragon (approved):
LGTM! Just a single nit/suggestion.

singer_sdk/plugin_base.py
edgarrmondragon: We do have https://github.com/meltano/sdk/issues/915 so in the future we could make the samples into installable packages and this behavior could be tested with them.

Whether we decide to distribute them separately is a different question, imo.

edgarrmondragon:
@kgpayne I pushed a fix for mypy. Feel free to merge!

PR #1809, opened 2023-07-05
fix(targets): Deserialize floats as `decimal.Decimal`

edgarrmondragon:


edgarrmondragon:
@kgpayne can you give this PR a 👀 when you got a minute?

pnadolny13 (approved):
@edgarrmondragon LGTM!

PR #1820, opened 2023-07-07
fix(targets): Include SCHEMA message count in target logs

edgarrmondragon:
From a (private) [Slack conversation](https://meltano.slack.com/archives/C05FP72408K/p1688678721243459), it seems that we're missing the SCHEMA message count which makes it seem that the total is missing something.

Example of new output:

```
2023-07-07 12:46:08,109 | INFO     | target-csv           | Target 'target-csv' completed reading 262 lines of input (2 schemas, 257 records, 0 batch manifests, 3 state messages).
```


pnadolny13 (approved):
@edgarrmondragon LGTM!

Issue #1807, opened 2023-07-05
feat: cookiecutter template should prompt for License options


pnadolny13:
We're finding that many open source SDK connector repos dont have a license. Many times the developer simply forgets to add one after using the cookiecutter to get started. Having a license be part of the getting started workflow should lead to more repos containing licenses.

Potential Solutions:
- Add a few common permissive licenses to the cookiecutter template - MIT, Apache 2.0
- Prompt the user to select a license while collecting info about the connector. If they want a custom license we can write a blank file to the repo as a placeholder
- Alternatively we could add TODOs to the README, and docs to the guide to remind developers about the license, etc. This is less ideal though, maybe we could do it in addition to the cookiecutter options.

edgarrmondragon:
@pnadolny13 This was added back in #1709.

<img width="797" alt="Screenshot 2023-07-05 at 11 21 33" src="https://github.com/meltano/sdk/assets/16805946/0ae0e674-90cd-4160-99c5-28690f6462d1">

Maybe this issue can instead be about adding _more_ license options?

pnadolny13:
@edgarrmondragon oh shoot I didnt find that in my search. I was creating this because we had the action item from the singer guild meeting last time. I'll convert this to be for adding MIT license as an option.

PR #1829, opened 2023-07-10
fix(templates): Update cookiecutter copyright assignment to cookiecutter user

### LINKED ISSUE #1828: bug: Copyright assignment of new taps/sinks to Meltano.
### Singer SDK Version

0.29.0

### Is this a regression?

- [ ] Yes

### Python Version

3.10

### Bug scope

Taps (catalog, state, etc.)

### Operating System

Linux

### Description

Cookiecuttter templates include Apache-2 as a default license (:100:), however the copyright assignment and date of copyright [are not templated](https://github.com/meltano/sdk/blob/8fb30f37ac9fa107237d3c197c5c9b4e8b931132/cookiecutter/tap-template/%7B%7Bcookiecutter.tap_id%7D%7D/%7B%25if%20'Apache-2.0'%20%3D%3D%20cookiecutter.license%20%25%7DLICENSE%7B%25endif%25%7D#L189) and so are always: `Copyright 2021 Meltano`. Since you don't seem like a CLA shop (:grinning:), I'm assuming this is a bug.

One way to handle this might be:
`Copyright {% now 'utc', '%Y' %} {% admin_name %}`

This would set the correct copyright year and be directionality correct for the rightsholder.

This would ask no new questions as part of the onboarding process.  One possible downside to this approach might be that if someone's doing this for work, odds are the copyright assignment is to their employer. 

Should #1807 (MIT License) be implemented, the same approach should be applied there as well.

### Code

_No response_

riordan:
Addresses
- #1828 

Assigns copyright in the current year (not 2021) and to the `admin_name` set in the cookiecutter workflow.

---

Should this be merged, I'd happily submit a PR to do the same for:
- #1807 

This is, of course, conditional on whether or not we want to encourage MIT licensing within the Meltano/Singer community (as Apache-2 comes with explicit patent usage license, whereas MIT does not).

Closes #1828

----
:books: Documentation preview :books:: https://meltano-sdk--1829.org.readthedocs.build/en/1829/

### DIFF
--- a/cookiecutter/tap-template/{{cookiecutter.tap_id}}/{%if 'Apache-2.0' == cookiecutter.license %}LICENSE{%endif%}
+++ b/cookiecutter/tap-template/{{cookiecutter.tap_id}}/{%if 'Apache-2.0' == cookiecutter.license %}LICENSE{%endif%}
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   
+
    Copyright {% now 'utc', '%Y' %} {{ cookiecutter.admin_name }}
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):
Thanks @riordan, this LGTM!

If you're willing to submit a PR for https://github.com/meltano/sdk/issues/1807 then I'd be happy to review it.

PR #1835, opened 2023-07-12
fix(taps): Avoid setting up mapper in discovery mode

edgarrmondragon:
Closes #1821

----
:books: Documentation preview :books:: https://meltano-sdk--1835.org.readthedocs.build/en/1835/

PR #1842, opened 2023-07-14
fix: Ensure all expected tap parameters are passed to `SQLTap` initializer

edgarrmondragon:
Fixes the broken initialization of SQL taps following https://github.com/meltano/sdk/pull/1835.

The subclass initializer doesn't introduce any new arguments, so this PR changes it to use `*args` and `**kwargs` so we don't miss any changes to the parent class' initializer.

cc @BuzzCutNorman

----
:books: Documentation preview :books:: https://meltano-sdk--1842.org.readthedocs.build/en/1842/


edgarrmondragon (commented):
`Tap.__init__` and `SQLTap.__init__` have the same signature so this feels like a safe change. Merging.

PR #1843, opened 2023-07-14
fix: Target template should not reference `tap_id`

pnadolny13:
```
Unable to create file '.github/workflows/{% if cookiecutter.include_ci_files == 'GitHub' %}test.yml{%endif%}'
Error message: 'collections.OrderedDict object' has no attribute 'tap_id'
```

----
:books: Documentation preview :books:: https://meltano-sdk--1843.org.readthedocs.build/en/1843/


edgarrmondragon:
Hmm there seems to be a bug in the latest release of Cookiecutter. I logged an issue https://github.com/cookiecutter/cookiecutter/issues/1900.

I'll try downgrading the version we use in CI.

edgarrmondragon (approved):

Issue #1836, opened 2023-07-13
bug: `_parse_timestamps_in_record` throws exception for key not present in schema


menzenski:
### Singer SDK Version

0.19.0

### Is this a regression?

- [ ] Yes

### Python Version

NA

### Bug scope

Targets (data type handling, batching, SQL object generation, etc.)

### Operating System

macOS, Linux

### Description

We use the meltanolabs variant of the target-postgres loader with the ets variant of the tap-spreadsheets-anywhere extractor.

Tap-spreadsheets-anywhere supports dynamic schema inference from randomly selected sample files. This is very useful as we do a lot of loading CSV+JSONL files from AWS S3, for instance.

Our meltano runs are frequently erroring when a KeyError is thrown here https://github.com/meltano/sdk/blob/d372080fbfcbcd1f529ed3839c8e34f3a6aa0cc8/singer_sdk/sinks/core.py#L362 - the dynamic schema generation may not include all fields actually present in the data (especially for JSON type documents) - this causes the KeyError and that brings the whole pipeline to a halt.

It would be great if this function could catch the KeyError and only raise an exception if truly necessary - in the scenarios I've seen, I believe the KeyError could have been swallowed with no impact.

### Code

```python
meltano-workflow-h7c8z-extract-and-load-2678913587: {"consumer": false, "producer": true, "string_id": "tap-spreadsheets-anywhere-s3", "cmd_type": "elb", "stdio": "stderr", "name": "tap-spreadsheets-anywhere-s3", "event": "INFO Syncing file \"sub/dir/ectory/2022/04/01/filename.CSV\".", "level": "info", "timestamp": "2023-07-13T17:38:00.017169Z"}
meltano-workflow-h7c8z-extract-and-load-2678913587: {"consumer": true, "producer": false, "string_id": "target-postgres-staging", "cmd_type": "elb", "stdio": "stderr", "name": "target-postgres-staging", "event": "Traceback (most recent call last):", "level": "info", "timestamp": "2023-07-13T17:38:00.160258Z"}
meltano-workflow-h7c8z-extract-and-load-2678913587: {"consumer": true, "producer": false, "string_id": "target-postgres-staging", "cmd_type": "elb", "stdio": "stderr", "name": "target-postgres-staging", "event": "  File \"/project/.meltano/loaders/target-postgres/venv/bin/target-postgres\", line 8, in <module>", "level": "info", "timestamp": "2023-07-13T17:38:00.160499Z"}
[...]

PR #1844, opened 2023-07-14
fix(targets): Safely skip parsing record field as date-time if it is missing in schema

edgarrmondragon:
Closes https://github.com/meltano/sdk/issues/1836

* [x] Add tests

----
:books: Documentation preview :books:: https://meltano-sdk--1844.org.readthedocs.build/en/1844/


menzenski (approved):

singer_sdk/sinks/core.py
```diff
@@ -365,12 +365,15 @@ def _parse_timestamps_in_record(
             schema: TODO
             treatment: TODO
         """
-        for key in record:
+        for key, value in record.items():
+            if key not in schema["properties"]:
+                self.logger.debug("No schema for record field '%s'", key)
+                continue
```
edgarrmondragon, line 371: Using the `in` operator is a bit faster than try-except:

```python
import timeit

setup = """
big_dict = {
    i: i * i
    for i in range(50_000)
}
"""

try_except = """
for i in range(70_000):
    try:
        x = big_dict[i]
    except KeyError:
        continue
"""

in_operator = """
for i in range(70_000):
    if i in big_dict:
        x = big_dict[i]
"""

print(timeit.timeit(setup=setup, stmt=try_except, number=1000))   # 4.430988875003095
print(timeit.timeit(setup=setup, stmt=in_operator, number=1000))  # 2.137157458000729

```

PR #1864, opened 2023-07-20
feat: SQLTarget connector instance shared with sinks

### LINKED ISSUE #1772: feat: Target level SQLConnector instance shared with Sinks
### Feature scope

Targets (data type handling, batching, SQL object generation, tests, etc.)

### Description

Currently an SQL target will open and hold a session to the target for each stream it is given.  This means if a tap has 100 streams the SQL target will open 100 session and hold them until the `meltano run` is finished.  It would be nice if SQL targets could utilize a single connector and leverage the SQLAlchemy pool of connection it has.

This can be accomplished If an instance of the target's connector class is initialized and held as a property and that connector property is passed on to `SQLSink` instances when they are initialized by the `add_sink` method.  The property code might look like this:
```
class SQLTarget(Target):
    """Target implementation for SQL destinations."""

    default_connector_class = SQLConnector
    _target_connector: SQLConnector = None

    @property
    def target_connector(self) -> SQLConnector:
        """The connector object.

        Returns:
            The connector object.
        """
        if self._target_connector is None:
            self._target_connector = self.default_connector_class(dict(self.config))
        return self._target_connector
```

The `add_sink` method would need to pass `self.target_connector` to the `SQLSink` when it is initialized. A couple of quick notes:

- The SQLSink class has a `connector` parameter while the base `Sink` class does not. 
- The `add_sink` method is decorated with `@final` so it cannot be overloaded in the SQLTarget class.

The code could look something like this:

```
   @final
    def add_sink(

		... comments and code removed to show the change better
		
        sink = sink_class(
            target=self,
            stream_name=stream_name,
            schema=schema,
        )
		
		... some code removed to assist with clarity
		
        return sink
```

An example of this working in a target can be found in [target-mssql--buzzcutnorman:target.py](https://github.com/BuzzCutNorman/target-mssql/blob/main/target_mssql/target.py) and [target-postgress--buzzcutnorman:target.py](https://github.com/BuzzCutNorman/target-postgres/blob/main/target_postgres/target.py)

BuzzCutNorman:
This adds a variable `_target_connector` to `SQLTarget` to house an instance of the `default_sink_class.connector_class` and a property to access it. The method `get_sink` was brought down from `Target` and altered to call `add_sqlsink`.  The method `add_sqlsink` is a copy of the `add_sink` method in `Target` it just passes the shared connector instance to the `SQLSink` isntances it initalizes for each stream. 

need some assistance with making this mypy compliant.

Closes #1772

----
:books: Documentation preview :books:: https://meltano-sdk--1864.org.readthedocs.build/en/1864/


edgarrmondragon:
@BuzzCutNorman can you try fixing the mypy failure? Otherwise I can take look in the coming days.

BuzzCutNorman:
@edgarrmondragon I was able to get the mypy failures resolved.

tests/core/test_target_base.py
```diff
@@ -53,3 +56,68 @@ def test_validate_record():
...
+        key_properties=key_properties,
+        connector=target.target_connector,
+    )
+    target._sinks_active["foo"] = sink
+    sink_returned = target.get_sink(
+        "foo",
+        schema=input_schema_2,
+        key_properties=key_properties,
+    )
+    assert sink_returned == sink
```
edgarrmondragon, line 89: Can we change this to

```suggestion
    assert sink_returned is sink
```

to ensure we're getting the exact same instance?

tests/core/test_target_base.py
```diff
@@ -53,3 +56,68 @@ def test_validate_record():
...
+        "foo",
+        schema=input_schema_2,
+        key_properties=key_properties,
+    )
+
+    sink_returned = target.get_sink(
+        "foo",
+    )
+
+    assert sink_returned == sink
```
edgarrmondragon, line 117: Ditto

```suggestion
    assert sink_returned is sink
```

singer_sdk/target_base.py
```diff
@@ -617,3 +635,114 @@ def _merge_missing(source_jsonschema: dict, target_jsonschema: dict) -> None:
         super().append_builtin_config(config_jsonschema)
 
     pass
+
+    @final
+    def add_sqlsink(
```
edgarrmondragon, line 640: Wdyt about renaming this to `register_sql_sink`?

edgarrmondragon (commented):
Thank you @BuzzCutNorman, overall this is looking great. I still have to go through the API and see how this fits with our plans for connectors (at first glance it seems it does!).

singer_sdk/target_base.py
BuzzCutNorman: I like the name `register_sql_sink`, but I don't know about renaming the method.  Since I copied `add_sink` it made sense to be `add_sqlsink` so the names were close enough you would know `add_sink` was the original source since we couldn't overload it.  I would be total ok with changing `add_sink` to `register_sink` and `add_sqlsink` to `register_sql_sink`. Wdyt

edgarrmondragon: Hmm, yeah it should be safe to rename `add_sink` but even though folks are not supposed to override it, they might still be overriding a different method that calls it. We could leave it as an alias of `register_sink` and add a deprecation warning but that seems unnecessarily complicated.

Let's just leave `add_sqlsink` and `add_sink` as they are.

edgarrmondragon (commented):
It doesn't look like we need to update the cookiecutter templates, can you confirm that?

BuzzCutNorman:
You are correct the cookiecutter templates do not need updating.  I my initial version they did need to be updated.  Then thanks to the sdk test I realized there was a way to accomplish this without a adding a `default_connector_class` variable.  This also means target developers won't need to make any code changes to their targets for the target to benefit from this PR.

edgarrmondragon:
> You are correct the cookiecutter templates do not need updating. In my initial version they did need to be updated. Then thanks to the sdk test I realized there was a way to accomplish this without adding a `default_connector_class` variable. This also means target developers won't need to make any code changes to their targets for the target to benefit from this PR.

Thanks for confirming!

edgarrmondragon (approved):

edgarrmondragon:
Thanks @BuzzCutNorman!

PR #1876, opened 2023-07-25
feat: Configurable batch size and max wait limit for targets

### LINKED ISSUE #1626: feat: Configurable batch size and max wait limit for targets
### Feature scope

## Prior art

Many targets in the Singer ecosystem ([target-snowflake](https://github.com/transferwise/pipelinewise-target-snowflake/) is one) support the following settings or some variant of them:

| Property                            | Type    | Description                                                   |
|-------------------------------------|---------|---------------------------------------------------------------|
| batch_size_rows                     | Integer | (Default: 100000) Maximum number of rows in each batch. At the end of each batch, the rows in the batch are loaded into Snowflake. |
| batch_wait_limit_seconds            | Integer | (Default: None) Maximum time to wait for batch to reach `batch_size_rows`. |

## Current SDK support

The Singer SDK currently only allows target _developers_ to configure the maximum size of a batch via an attribute:

https://github.com/meltano/sdk/blob/43f7323b70092292f6090e2f7e75a5fc5385d66f/singer_sdk/sinks/core.py#L45

By making this configurable, users would have better control of the resources in a constrained environment (i.e. low memory).

## Proposal

* Add a new pair of built-in settings `batch_size_rows` and `batch_wait_limit_seconds`
* Update `Sink.is_full` to use `batch_size_rows` and fall back to `Sink.MAX_SIZE_DEFAULT` if `batch_size_rows` is not configured
* Track the start time of each batch and implement `Sink.is_too_old` to check against `batch_wait_limit_seconds`
* Update the `TargetBase` class to make use of the new method that checks the batch TTL

## Related

* https://github.com/meltano/sdk/issues/134
*

BuzzCutNorman:
This is an attempt at implementing the batch size and wait limit for targets as outlined.

Closes #1626

----
:books: Documentation preview :books:: https://meltano-sdk--1876.org.readthedocs.build/en/1876/


singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
...
+    def perf_diff_allowed_max(self) -> float:
+        """The maximum postive variace allowed, # 3/4 better than wanted."""
+        return self.max_perf_counter * 0.75
+
+    @property
+    def perf_diff(self) -> float:
+        """Difference between wanted elapsed time and actual elpased time."""
+        diff = 0
+        if self._lap_time:
+            diff = self.max_perf_counter - self.lap_time
```
edgarrmondragon, line 90:
```suggestion
        diff = self.max_perf_counter - self.lap_time if self._lap_time else 0
```

singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
+"""performace timers which deal with dynamic timing events."""
```
edgarrmondragon, line 1:
```suggestion
"""Performance timers which deal with dynamic timing events."""
```

singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
...
+        return -1.0 * (self.max_perf_counter * 0.33)
+
+    @property
+    def perf_diff_allowed_max(self) -> float:
+        """The maximum postive variace allowed, # 3/4 better than wanted."""
+        return self.max_perf_counter * 0.75
+
+    @property
+    def perf_diff(self) -> float:
+        """Difference between wanted elapsed time and actual elpased time."""
```
edgarrmondragon, line 87:
```suggestion
        """Difference between wanted elapsed time and actual elapsed time."""
```

singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(
+        self,
+        max_size: int,
+        max_perf_counter: float,
+    ) -> None:
+        self._sink_max_size: int = max_size
+        self._max_perf_counter: float = max_perf_counter
+
+    SINK_MAX_SIZE_CELING: int = 100000
```
edgarrmondragon, line 62:
```suggestion
    SINK_MAX_SIZE_CEILING: int = 100000
```

singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
...
+        return self._sink_max_size
+
+    @property
+    def max_perf_counter(self) -> float:
+        """How many seconds can pass before a insert."""
+        return self._max_perf_counter
+
+    @property
+    def perf_diff_allowed_min(self) -> float:
+        """The mininum negative variance allowed, 1/3 worse than wanted."""
```
edgarrmondragon, line 77:
```suggestion
        """The minimum negative variance allowed, 1/3 worse than wanted."""
```

singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
...
+        return -1.0 * (self.max_perf_counter * 0.33)
+
+    @property
+    def perf_diff_allowed_max(self) -> float:
+        """The maximum postive variace allowed, # 3/4 better than wanted."""
```
edgarrmondragon, line 82:
```suggestion
        """The maximum positive variance allowed, # 3/4 better than wanted."""
```

singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
...
+    @property
+    def perf_diff(self) -> float:
+        """Difference between wanted elapsed time and actual elpased time."""
+        diff = 0
+        if self._lap_time:
+            diff = self.max_perf_counter - self.lap_time
+        return float(diff)
+
+    def counter_based_max_size(self) -> int:  # noqa: C901
+        """Caclulate performance based batch size."""
```
edgarrmondragon, line 94:
```suggestion
        """Calculate performance based batch size."""
```

singer_sdk/helpers/_perftimer.py
```diff
@@ -0,0 +1,118 @@
...
+                correction = -5000
+            elif self.sink_max_size >= 10000:  # noqa: PLR2004
+                correction = -1000
+            elif self.sink_max_size >= 1000:  # noqa: PLR2004
+                correction = -100
+            elif self.sink_max_size > 10:  # noqa: PLR2004
+                correction = -10
+        if (
+            self.perf_diff >= self.perf_diff_allowed_max
+            and self.sink_max_size < self.SINK_MAX_SIZE_CELING
```
edgarrmondragon, line 107:
```suggestion
            and self.sink_max_size < self.SINK_MAX_SIZE_CEILING
```

tests/core/test_perf_timer.py
```diff
@@ -0,0 +1,98 @@
...
+    assert batchtimer.perf_diff > batchtimer.perf_diff_allowed_max
+    assert batchtimer.counter_based_max_size() == 20
+    batchtimer._sink_max_size = 100
+    assert batchtimer.counter_based_max_size() == 200
+    batchtimer._sink_max_size = 1000
+    assert batchtimer.counter_based_max_size() == 2000
+    batchtimer._sink_max_size = 10000
+    assert batchtimer.counter_based_max_size() == 20000
+    batchtimer._sink_max_size = 100000
+    assert batchtimer.sink_max_size == batchtimer.SINK_MAX_SIZE_CELING
```
edgarrmondragon, line 81:
```suggestion
    assert batchtimer.sink_max_size == batchtimer.SINK_MAX_SIZE_CEILING
```

tests/core/test_perf_timer.py
```diff
@@ -0,0 +1,98 @@
...
+    assert batchtimer.counter_based_max_size() == 100000
+
+
+def test_batchperftimer_counter_based_max_size_reductive():
+    batchtimer: BatchPerfTimer = BatchPerfTimer(100000, 1)
+    batchtimer._lap_time = 1.5
+    assert batchtimer.perf_diff < batchtimer.perf_diff_allowed_min
+    assert batchtimer.sink_max_size == batchtimer.SINK_MAX_SIZE_CELING
```
edgarrmondragon, line 89:
```suggestion
    assert batchtimer.sink_max_size == batchtimer.SINK_MAX_SIZE_CEILING
```

singer_sdk/sinks/sql.py
```diff
@@ -336,6 +399,14 @@ def bulk_insert_records(
...
             result = conn.execute(insert_sql, new_records)
 
+        # Finish Line for max_size perf counter
+        if self.sink_timer is not None:
+            if self.sink_timer.start_time is not None:
+                self.sink_timer.stop()
+                self.batch_size_rows = self.sink_timer.counter_based_max_size()
+
+            # Starting Line for max_size perf counter
+            self.sink_timer.start()
```
edgarrmondragon, line 409: Wdyt about moving this up the stack? Perhaps to https://github.com/meltano/sdk/blob/f6bbf0c5ddba689ee1ab6df1f110529f35adf12f/singer_sdk/target_base.py#L498

That way all targets, not just SQL, benefit from this.

edgarrmondragon (changes requested):
Thanks for your work on this @BuzzCutNorman!

I think this https://github.com/meltano/sdk/blob/f6bbf0c5ddba689ee1ab6df1f110529f35adf12f/singer_sdk/sinks/core.py#L128-L135

has to be updated to take the timer into account, wdyt?

BuzzCutNorman:
> Thanks for your work on this @BuzzCutNorman!
> 
> I think this
> 
> https://github.com/meltano/sdk/blob/f6bbf0c5ddba689ee1ab6df1f110529f35adf12f/singer_sdk/sinks/core.py#L128-L135
> 
> has to be updated to take the timer into account, wdyt?

@edgarrmondragon I tried to account for this in the following way.  I reflect the timer batch size row changes in `max_size` by changing `self.max_size` to point to `self.batch_size_rows` not `self.MAX_SIZE_DEFAULT`  
https://github.com/BuzzCutNorman/sdk/blob/74b16533cbab619bcc516b124f522e0c180c8073/singer_sdk/sinks/sql.py#L126-L133
Then have `_batch_size_rows` grab the size from the target config or `MAX_SIZE_DEFAULT` during initialization. 
https://github.com/BuzzCutNorman/sdk/blob/74b16533cbab619bcc516b124f522e0c180c8073/singer_sdk/sinks/sql.py#L53-L55
The timer then updates `self.batch_size_rows` via the setter associated to it.
https://github.com/BuzzCutNorman/sdk/blob/74b16533cbab619bcc516b124f522e0c180c8073/singer_sdk/sinks/sql.py#L406

singer_sdk/sinks/sql.py
BuzzCutNorman: I like it.  The code has been moved up.  Please let me know if this matches up with what you had in mind.

BuzzCutNorman:
@edgarrmondragon if you would please review this PR again when you get a chance.

edgarrmondragon:
We probably want to append the new settings: https://github.com/meltano/sdk/blob/4b5b3c6d30a16747893ad8d934eec2ee155fa757/singer_sdk/target_base.py#L577-L607

singer_sdk/sinks/core.py
```diff
@@ -224,6 +277,26 @@ def key_properties(self) -> list[str]:
...
+            timer_msg: str = (
+                f"batch_size_rows: {self.batch_size_rows}, "
+                f"min: {self.sink_timer.perf_diff_allowed_min}, "
+                f"now: {self.sink_timer.perf_diff:.4f}, "
+                f"max {self.sink_timer.perf_diff_allowed_max}, "
+            )
+            self.logger.info(timer_msg)
+            if self.sink_timer.start_time is not None:
+                self.sink_timer.stop()
+                self.batch_size_rows = self.sink_timer.counter_based_max_size()
```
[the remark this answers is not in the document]
edgarrmondragon: @BuzzCutNorman My concern is more around how we use the timer to artificially fill up the row count so sink is drained. I guess it's mostly an implementation detail, so wdyt about documenting how the batch size limit and wait limit are used under the hood in `docs/implementation`?

BuzzCutNorman:
> @BuzzCutNorman My concern is more around how we use the timer to artificially fill up the row count so sink is drained. I guess it's mostly an implementation detail

@edgarrmondragon Sorry I had tunnel vision.  I think I see what you mentioned above.  In the original proposal you have `is_too_old` which is true when the difference between the `start_time` of a batch and the time at evaluation is greater than `batch_wait_limit_seconds`.  If `is_too_old` is true then drain the batch. While I am artificially filling up the row count.  

Please let me know if I am on the same page now. 😊

edgarrmondragon:
> > @BuzzCutNorman My concern is more around how we use the timer to artificially fill up the row count so sink is drained. I guess it's mostly an implementation detail
> 
> @edgarrmondragon Sorry I had tunnel vision. I think I see what you mentioned above. In the original proposal you have `is_too_old` which is `True` when the difference between the `start_time` of a batch and the time at evaluation is greater than `batch_wait_limit_seconds`. If `is_too_old` is true then drain the batch. While I am artificially filling up the row count.
> 
> Please let me know if I am on the same page now. 😊

Yup, that's correct.

BuzzCutNorman:
@edgarrmondragon I added in `is_too_old` then created a new check in `_process_record_message` using it.  Working with `batch_size_rows` and `batch_wait_limit_seconds` I didn't find a way to deal with widely varying ingest and write speeds.  To accomplish this, I added in `batch_dynamic_management` which takes `batch_size_rows` and `batch_wait_limit_seconds` and uses them as the max limits for `counter_based_max_size`.  I updated `counter_based_max_size` not to return a value but update the internal `sink_max_size` which if `batch_dynamic_management` is set will be used in `is_full` instead of `max_size` to determine if `current_size` is at the point to drain.  In doing this I can set `batch_size_rows: 1000000` and `batch_wait_limit_seconds: 30` and a quick loading table I test with drains at 110000 rows while a slow loading test table will chug along at 400 rows a batch.

When you get a chance, please review these changes and let me know what you think.

singer_sdk/helpers/capabilities.py
```diff
@@ -106,6 +106,20 @@
...
+        description="Maximum number of rows in each batch.",
+    ),
+).to_dict()
+BATCH_WAIT_LIMIT_SECONDS_CONFIG = PropertiesList(
+    Property(
+        "batch_wait_limit_seconds",
+        IntegerType,
+        description="Maximum time to elapse for a batch to fill, drain, and load.",
+    ),
+).to_dict()
```
edgarrmondragon, line 122: Can you the other setting here?

singer_sdk/target_base.py
```diff
@@ -597,6 +608,8 @@ def _merge_missing(source_jsonschema: dict, target_jsonschema: dict) -> None:
                     target_jsonschema["properties"][k] = v
 
         _merge_missing(ADD_RECORD_METADATA_CONFIG, config_jsonschema)
+        _merge_missing(BATCH_SIZE_ROWS_CONFIG, config_jsonschema)
+        _merge_missing(BATCH_WAIT_LIMIT_SECONDS_CONFIG, config_jsonschema)
```
edgarrmondragon, line 612: You might also want to add the other setting here

singer_sdk/sinks/core.py
```diff
@@ -132,7 +148,83 @@ def is_full(self) -> bool:
...
             True if the sink needs to be drained.
         """
-        return self.current_size >= self.max_size
+        return (
+            self.current_size >= self.sink_timer.sink_max_size
+            if self.batch_dynamic_management and self.sink_timer is not None
+            else self.current_size >= self.max_size
+        )
+
+    @property
```
edgarrmondragon, line 157: I don't think this needs to be a property, wdyt?

BuzzCutNorman: I think it makes sense as a property.  I like the cleanness of being able to access it via `.is_too_old`.  That keeps it consistent with `.is_full`.  It is a computed value.  I don't need to pass any parameter to obtain it.  I get a single value back which is used in a comparison when processing each record. If anything I am thinking maybe change it to `@cached_property` since it will never change so why compute it over and over again.  wdyt?

edgarrmondragon: Oh, I think `@cached_property` definitely would not work since the value is definitely dynamic.

I agree that even for the sake of consistency with `.is_full` it makes sense as a property 👍

BuzzCutNorman: Yep after I hit comment I realize how dumb that was.

edgarrmondragon:
@BuzzCutNorman I think this is really close 🤏 

What did you think of this?

> I guess it's mostly an implementation detail, so wdyt about documenting how the batch size limit and wait limit are used under the hood in `docs/implementation`?

https://github.com/meltano/sdk/pull/1876#discussion_r1296379407

BuzzCutNorman:
@edgarrmondragon before I get to far down the road.  Do you like `batch_dynamic_management` as the name for this setting?  I am not 100 percent sold on it but used it as something to work with.  I came up with these that could also work. `batch_size_rows_managed`,  `dynamic_batch_sized_rows`,  `optimize_batch_row_size_for_wait_limit`

BuzzCutNorman:
> Thanks for your work on this @BuzzCutNorman!
> 
> I think this
> 
> https://github.com/meltano/sdk/blob/f6bbf0c5ddba689ee1ab6df1f110529f35adf12f/singer_sdk/sinks/core.py#L128-L135
> 
> has to be updated to take the timer into account, wdyt?

@edgarrmondragon Ok, I finally caught up with what you were meaning with this comment.  It became clear when I set `batch_wait_limit_in_seconds` and didn't add a `batch_size_rows`.  Since it took less time to fill the batch to the default size of 10000 rows than the wait limit, it kept draining early.  I came up with two ways to fix this:

1. I add a `wait_limit_only` config option and tell `is_full` to always be false when that is set to True.  (Not my favorite)
2. Use four properties `is_full_rows`, `is_too_old`, `is_full_rows_and_too_old`, and `is_full_dynamic`.  Create an internal variable `_drain_property` that I set to the appropriate property with `set_drain_property` then have `is_full` return `_drain_property`. (I like this one)

Please let me what you think of the paths I came up with and defiantly let me know if you know of a better way to doing this. Thanks 🙏

edgarrmondragon:
@BuzzCutNorman I'm a bit confused now 😅. 

I'm failing to parse this

> When the Meltano Target configuration option `batch_dynamic_management` is set to `True` you are asking the `Sink.sink_timer` to find the maximum `rows` is full mark that keeps the time to fill a batch with records and write those records to the Target's target within the `time` in `seconds` given.

I wonder how https://github.com/transferwise/pipelinewise-target-snowflake manages this 🤔

BuzzCutNorman:
> I wonder how https://github.com/transferwise/pipelinewise-target-snowflake manages this 🤔

@edgarrmondragon I don't think they do I think they have dynamic batch management as an option.  Or utilize a performance counter class.  Looking just now this all I found in the `__init__.py` .
```
            flush = False
            if row_count[stream] >= batch_size_rows:
                flush = True
                LOGGER.info("Flush triggered by batch_size_rows (%s) reached in %s",
                            batch_size_rows, stream)
            elif (batch_wait_limit_seconds and
                  datetime.utcnow() >= (flush_timestamp + timedelta(seconds=batch_wait_limit_seconds))):
                flush = True
                LOGGER.info("Flush triggered by batch_wait_limit_seconds (%s)",
                            batch_wait_limit_seconds)
```

BuzzCutNorman:
> I'm failing to parse this
>
>>When the Meltano Target configuration option batch_dynamic_management is set to True you are asking the Sink.sink_timer >>to find the maximum rows is full mark that keeps the time to fill a batch with records and write those records to the Target's >>target within the time in seconds given.

@edgarrmondragon I might be overselling what the option is accomplishing.  When I look at the steps in the `drain_one` method they are as follows:
```
        draining_status = sink.start_drain()
        sink.process_batch(draining_status)
        sink._lap_manager()
        sink.mark_drained()
```
The `_lap_manager` starts the clock just after a `process_batch` finishes.  The batch performance timer is running when the batch is filled, the drain is started, and the process batch is run which I thought triggered writing the batched records to the target system. Once those steps are done `_lap_manager` stops the clock, calls `counter_based_max_size` which uses the difference between the lap time and `batch_wait_limit_in_seconds`, to determines if a correction needs to be made to the is full marker, then starts the clock again.

BuzzCutNorman:
@edgarrmondragon if you would please review this PR again when you get a chance.

docs/implementation/target_batch_full.md
```diff
@@ -0,0 +1,58 @@
...
... turns into `Sink.sink_timer.max_perf_counter` which is the time in seconds a full cycle should take.  The attribute `Sink.sink_timer.sink_max_size` starts at a predefined size of `100`.  During the `Target.drain_one(Sink)` process `Sink._lap_manager` is called and the timer method `counter_based_max_size` runs and checks if `Sink.sink_timer.perf_diff`, which is `max_perf_counter` - `lap_time`, is greater than `Sink.sink_timer.perf_diff_allowed_max` or less than `Sink.sink_timer.perf_diff_allowed_min`.  If `Sink.sink_timer.perf_diff` is greater than `Sink.sink_timer.perf_diff_allowed_max` the `Sink.sink_timer.sink_max_size` is increased as long as the `Sink.sink_timer.sink_max_size` is less than `Sink.sink_timer.SINK_MAX_SIZE_CEILING`. If `Sink.sink_timer.perf_diff` is less than `Sink.sink_timer.perf_diff_allowed_min` the `Sink.sink_timer.sink_max_size` is reduced.  If the `Sink.sink_timer.perf_diff` is between `Sink.sink_timer.perf_diff_allowed_max` and `Sink.sink_timer.perf_diff_allowed_min` no correction to `Sink.sink_timer.sink_max_size` is made since the optimal rows size has been reached.  This process is repeated when each `Sink` is initialized and starts processing records.
```
edgarrmondragon, line 58: I'm still finding this a bit confusing. I think we want the max batch size to have a default value, and the max wait time to default to `None` so that in the default case, only the current batch size is checked.

Essentially:

```mermaid
flowchart TD
    A[Process next row] --> B{current_size ><br/>max_size?}:::q
    B -- Yes --> C[Drain]:::drain
    B -- No ---> D{is max_wait_time<br/>set?}:::q
    D -- No ---> A
    D -- Yes --> F{elapsed ><br/>max_wait_time?}:::q
    F -- No ---> A
    F -- Yes --> C
    classDef q fill:#FFE666
    classDef drain fill:#FF6680
```

Does it make sense? I'm sorry, I feel like I'm missing something 🙏

BuzzCutNorman: The diagram makes sense. I will need to work through it again tomorrow.  I think you feeling like you are missing something is a result of my poor writing.   Below is the best flowchart I could make.

```mermaid
flowchart TD
    C1-->A2
    D2-->A3
    E3-->E2
    C3-->A4
    G4-->D3
    F2-->A1
    subgraph Target._process_record_message
    A1[Process next row] --> B1{is_full, <p>sink.current_size > = sink_timer.sink_max_size}
    B1-->|True| C1[drain]
    B1-->|False| A1
   end
   subgraph Target.drain_one
   A2([start])-->B2[sink.start_drain]
   B2-->C2[sink.process_batch]
   C2-->D2[sink._lap_manager]
   D2~~~E2[sink.mark_drained]
   E2-->F2([end])
   end
   subgraph Sink.lap_manager
   A3([start])-->B3[sink_timer.stop]
   B3-->C3[sink_timer.counter_based_max_size]
   C3~~~D3[sink_timer.start]
   D3-->E3([end])  
   end
   subgraph sink_timer. counter_based_max_size
   A4([start])-->B4{perf_diff < self.perf_diff_allowed_min}
   B4-->|True| C4[correction = decrease amount]
   B4-->D4{perf_diff >= perf_diff_allowed_max<p>and<p>sink_max_size < SINK_MAX_SIZE_CEILING}
   D4-->|True| E4[correction = increase amount]
   D4-->F4[sink_max_size += correction]
   F4-->G4([end])
   end
```

BuzzCutNorman: > in the default case,
This would be when `batch_dynamic_management` is set to `True` in the target config and no `batch_size_rows` or `batch_wait_limit_seconds` was preset in the target config?

BuzzCutNorman: @edgarrmondragon When you have a free moment would you please provide me clarification on what you see the "default case" to be. 🙏😃

PR #1878, opened 2023-07-27
feat: Add `_sdc_sync_started_at` metadata column to indicate the start of the target process

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1878.org.readthedocs.build/en/1878/

### DIFF
--- a/singer_sdk/plugin_base.py
+++ b/singer_sdk/plugin_base.py
@@ -24,7 +24,7 @@ from singer_sdk.exceptions import ConfigValidationError
 from singer_sdk.helpers._classproperty import classproperty
 from singer_sdk.helpers._compat import metadata
 from singer_sdk.helpers._secrets import SecretString, is_common_secret_key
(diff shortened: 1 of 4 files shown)


singer_sdk/plugin_base.py
```diff
@@ -155,6 +156,9 @@ def __init__(
         metrics._setup_logging(self.config)
         self.metrics_logger = metrics.get_metrics_logger()
 
+        # Initialization timestamp
+        self.__initialized_at = int(time.time() * 1000)
```
edgarrmondragon, line 160: @pnadolny13 I used an integer here since that feels easier to filter and dedup with in most targets but let me know if you think otherwise

pnadolny13: @edgarrmondragon yeah that makes sense to me!

pnadolny13 (approved):
@edgarrmondragon great work! Looks good to me

Issue #1645, opened 2023-04-27
feat: add cookiecutter template for mappers


kgpayne:
### Feature scope

Other

### Description

Using [MeltanoLabs/meltano-map-transform](https://github.com/MeltanoLabs/meltano-map-transform) and [MeltanoLabs/map-gpt-embeddings](https://github.com/MeltanoLabs/map-gpt-embeddings) as examples, it would be great to have a cookiecutter template for `mapper` in addition to existing `tap` and `target` options.

Related to #801

PR #1892, opened 2023-08-02
feat: Mapper cookiecutter

pnadolny13:
Closes https://github.com/meltano/sdk/issues/1645

I was trying to build a mapper related to https://github.com/meltano/internal-data/issues/131 and noticed it was pretty difficult and manual. Creating at least a draft of this was going to save me time vs manually copy/pasting an existing mapper.

I wasnt sure what the cookiecutter.test.yml was actually doing, it looks like theyre outdated (e.g. `include_cicd_sample_template` vs what I think is now called `include_ci_files`). Let me know if I should put that back in.

I tried to update everything I could find but will manually test that it works as expected.

### DIFF
--- a/.github/workflows/cookiecutter-e2e.yml
+++ b/.github/workflows/cookiecutter-e2e.yml
@@ -74,7 +74,10 @@ jobs:
         path: |
           /tmp/tap-*
           /tmp/target-*
+          /tmp/mapper-*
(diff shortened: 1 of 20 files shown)


edgarrmondragon (approved):
Nice! Thanks @pnadolny13.

Issue #1783, opened 2023-06-21
SQL targets should allow configurable insert vs upsert behavior by default


pnadolny13:
Similar to https://github.com/meltano/sdk/issues/1781 we should add a default configuration enum to SQL targets to configure whether to run the target in insert (append only) mode or upsert mode. Some targets will want this behavior so they'll implement it in a custom way, its better to get in front of it so we have a consistent config option and enum for all targets.

In addition we should consider recommending insert as the default option even if upsert is support and a key property is provided. This avoids challenges with deduplicating within a sync (i.e. unsorted streams, no timestamp or incrementing sort key, etc.).

pnadolny13:
There could also be the option to run in trunc and load mode. This mode would truncate the table prior to loading. Its a bit risky so we should document it well. Snowflake has the time travel feature that allows you to recover any dropped tables or truncated rows back to a certain timestamp (default 24hr retention) so its a little safer if the developer accidentally uses this feature and drops their data.

PR #1893, opened 2023-08-03
feat: Standard configurable load methods

pnadolny13:
Closes https://github.com/meltano/sdk/issues/1783

I was attempting to add this `overwrite` feature to target-pinecone https://github.com/MeltanoLabs/target-pinecone/issues/4 so I decided it would probably best to draft it in the SDK while I was at it.

- Adds a standard config `load_method` for: append-only, upsert, overwrite
- Implements the overwrite workflow in SQL targets
- By default overwrite isnt supported until the target developer activates it. Is this what we want or should it default to `True`?
- I wasnt sure if the truncate logic should go in `setup` of the SQLSink class or `prepare_table`. For now I chose `prepare_table`
- Ideally a target would throw a nice NotImplementedError if the user choses one of the load methods that arent supported but I'm not totally sure how to do that as of right now. It might require more thought. 

@edgarrmondragon let me know what you think overall and also specifically with the naming of everything.

cc @tayloramurphy in case you have opinions on naming or default behavior.


tayloramurphy:
@pnadolny13 I like overwrite! It matches Snowflake too https://docs.snowflake.com/en/sql-reference/sql/insert#optional-parameters

edgarrmondragon (commented):
@pnadolny13 I like the approach 👌. Perhaps we can add a test for the SQLite sample.

pnadolny13:
@edgarrmondragon I added a test for the sqlite sample, and refactored a little. I couldnt get custom query parameterization to work for some reason so I used f strings and had to add noqa flags for ruff to pass.

singer_sdk/connectors/sql.py
```diff
@@ -766,6 +808,8 @@ def prepare_table(
                 as_temp_table=as_temp_table,
             )
             return
+        if self.config["load_method"] == TargetLoadMethods.OVERWRITE:
+            self.truncate_table(full_table_name=full_table_name)
```
edgarrmondragon, line 812: My main concern here would be what would happen if loading fails for some unrelated reason, then the data in the table is lost, right?

pnadolny13: Yeah thats a good point. I was originally thinking that since the user is choosing to overwrite completely, thats the risk theyre taking. If you overwrite and your input is empty then you get an empty table.

Although we could be nicer than that and use different workflow that renames the table instead of truncating. Then if it fails we can rollback and rename the original table back before exiting and if it succeeds we'd drop the original table completely once the sync is successful. Is that what you were thinking?

pnadolny13: In a future iteration we can implement this more forgiving workflow to:

1. rename existing table as a backup
2. recreate new table
3. if success then delete renamed backup table
4. if fail then rollback by deleting new table and renaming the backup table back to the original name

This is my thoughts on how we would implement that:
1. in `prepare_table` we add logic to do this
2. also in `prepare_table`
3. we could implement the [clean_up method](https://github.com/meltano/sdk/blob/36f74363f0eaa9c2e2114af6a0d7de2790a20133/singer_sdk/sinks/core.py#L489). 
4. I'm not sure where a try/catch rollback step would fit in right now

@edgarrmondragon any thoughts on where 4 could be done? I'm wondering now if we ultimately want to use this workflow in the future then using a drop/recreate workflow now instead of introducing a new truncate method might be better, since the truncate method would need to be deprecated almost immediately. I considered that at first but didnt know if there were any risks of dropping a table vs truncating it, it felt like truncate was safer. Thoughts?

pnadolny13:
@edgarrmondragon I swapped the logic in this to use a drop/recreate flow instead of truncating. I started to implement the rename + rollback workflow but I wasnt able to find anywhere to easily put logic to rollback if an exception is thrown. I tried using the [clean_up](https://github.com/meltano/sdk/blob/36f74363f0eaa9c2e2114af6a0d7de2790a20133/singer_sdk/sinks/core.py#L489) method which worked when the sync was successful to remove the backup table before exiting but I couldnt figure out how to easily catch exceptions and revert to the backup table prior to exiting.

edgarrmondragon:
@pnadolny13 I think I was probably overstating the risk of data loss caused by pipeline failures when syncing in _overwrite_ mode. If the user is OK with overriding existing data, then they're probably also OK-ish with losing data if the pipeline fails, as long as they can load data on the next run.

edgarrmondragon:
> @edgarrmondragon I swapped the logic in this to use a drop/recreate flow instead of truncating. I started to implement the rename + rollback workflow but I wasnt able to find anywhere to easily put logic to rollback if an exception is thrown. I tried using the [clean_up](https://github.com/meltano/sdk/blob/36f74363f0eaa9c2e2114af6a0d7de2790a20133/singer_sdk/sinks/core.py#L489) method which worked when the sync was successful to remove the backup table before exiting but I couldnt figure out how to easily catch exceptions and revert to the backup table prior to exiting. I create a [draft PR](https://github.com/meltano/sdk/pull/1941) to this branch with the code I was tinkering with.

@pnadolny13 Let'd punt the rollback support and create an issue for it. We'll worry about it if users are annoyed by losing data 🙂

edgarrmondragon (approved):

PR #1911, opened 2023-08-16
Document get_schema_names behavior

visch:
Mostly pushing up to call this out, not sure of what the fix should be yet. 

----
:books: Documentation preview :books:: https://meltano-sdk--1911.org.readthedocs.build/en/1911/

### DIFF
--- a/singer_sdk/connectors/sql.py
+++ b/singer_sdk/connectors/sql.py
@@ -496,6 +496,7 @@ class SQLConnector:
         result: list[dict] = []
         engine = self._engine
         inspected = sqlalchemy.inspect(engine)
+        # get_schema_names returns all databases in the DB, not just schemas
         for schema_name in self.get_schema_names(engine, inspected):
             # Iterate through each table and view
             for table_name, is_view in self.get_object_names(


singer_sdk/connectors/sql.py
```diff
@@ -496,6 +496,7 @@ def discover_catalog_entries(self) -> list[dict]:
         result: list[dict] = []
         engine = self._engine
         inspected = sqlalchemy.inspect(engine)
+        # get_schema_names returns all databases in the DB, not just schemas
```
edgarrmondragon, line 499: It's not clear to me if the problem is that this method returns database names for some dialects or if it returns schema names for _all_ databases.

visch: It looks like it returns database names and schema names. It's confusing. I know for sure is that right now we're pulling all databases anyone has access to for both tap-mysql and tap-postgres. 

Here's a dump

```
2023-08-16 17:31:07,255 | INFO     | sqlconnector         | self.get_schema_names(engine, inspected)=['information_schema', 'melty', 'mysql', 'performance_schema', 'sys', 'testdb']
```

From patching in this code

```

    def discover_catalog_entries(self) -> list[dict]:
        """Return a list of catalog entries from discovery.

        Returns:
            The discovered catalog entries as a list.
        """
        result: list[dict] = []
        engine = self._engine
        inspected = sqlalchemy.inspect(engine)
        self.logger.info(f"{self.get_schema_names(engine, inspected)=}")
        for schema_name in self.get_schema_names(engine, inspected):
 ```
 
 Actual databases here that were manually created are `melty` and `testdb` 
 
 I'm realizing now what's going on. In Mysql schemas == databases.

edgarrmondragon: Do you think it's worth calling this behaves differently in different SQLAlchemy dialect?

visch: I don't know I thought this was a bug but now that I understand it it's just a mysql specific oddity which is kind of expected of all sql alchemy dialect stuff. I'll just close this

PR #1925, opened 2023-08-23
Add viztracer command for testing targets

mjsqu:
When it comes to testing targets it took me a long while to work out I could use the `--input` switch to pass a file of records to the target rather than trying to work out how to run `viztracer` on a command containing a pipe 🤯. If the added line can be proven to work by a reviewer and it's OK to be added to the documentation then it would have saved me heaps of time yesterday 😄

### DIFF
--- a/docs/dev_guide.md
+++ b/docs/dev_guide.md
@@ -264,6 +264,7 @@ Then simply run your package's CLI as normal, preceded by the `viztracer` comman
 
 ```console
 $ poetry run viztracer my-tap
+$ poetry run viztracer -- my-target --config=config.json --input=messages.json
 ```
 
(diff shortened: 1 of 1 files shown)


mjsqu:
I also had some success by restricting the output of the command using `--include_files` - the full command I was running:

```
poetry run viztracer --include_files ${HOME}/.cache/pypoetry/virtualenvs/meltanolabs-target-snowflake* -o results.html -- target-snowflake --config=ENV --input=messages.json
```

edgarrmondragon (approved):
> If the added line can be proven to work by a reviewer

Confirmed 🙂

Thanks @mjsqu!

Issue #1332, opened 2023-01-18
[Bug]: Improve error message when replication key is not present in schema


edgarrmondragon:
### Singer SDK Version

0.17.0

### Python Version

NA

### Bug scope

Taps (catalog, state, stream maps, etc.)

### Operating System

NA

### Description

Right now, users only get the following rather unhelpful message:

```
ValueError: Could not detect type from empty type_dict. Did you forget to define a property in the stream schema?
```

Maybe the stream could catch `ValueError` and resurface it with a better message that contains the name of the field.

edgarrmondragon:
Still relevant

mjsqu:
Interested in working on this one, just going to jot some notes:

- The message comes from `singer_sdk.helpers._typing.EmptySchemaTypeError`
- This error is raised in `is_datetime_type`, called from`singer_sdk.streams.core.is_timestamp_replication_key`:

```py
    @property
    def is_timestamp_replication_key(self) -> bool:
        """Check is replication key is a timestamp.

        Developers can override to `True` in order to force this value, although this
        should not be required in most use cases since the type can generally be
        accurately detected from the JSON Schema.

        Returns:
            True if the stream uses a timestamp-based replication key.
        """
        if not self.replication_key:
            return False
        type_dict = self.schema.get("properties", {}).get(self.replication_key)
        return is_datetime_type(type_dict)

```

mjsqu:
Suggested steps to fix:

- add a new Exception to singer_sdk.exceptions, or leverage an existing one if it exists
- check the value of `type_dict` in the function `is_timestamp_replication_key` and raise the new Exception if `None`

PR #1927, opened 2023-08-27
fix: Handle replication key not found in stream schema

mjsqu:
Closes: #1332

Changes:

- Add new `InvalidReplicationKeyException` to `singer_sdk.exceptions`
- Update `streams.core`. A `get()` is used to obtain the `replication_key` from `self.schema`, if the key is not present in the schema, the variable `type_dict` would equal `None`. A check has been added, if the value of the `type_dict` is `None` then the new exception is raised
- The resulting error message is of the format:

```log
"incorrect_replication_key" is not in schema for stream name: my_stream"
```

- Add a test to `tests/core/test_streams.py` which checks that a `SimpleTestStream` with `replication_key` overridden to `"INVALID"` raises the correct `InvalidReplicationKeyException` exception.

### DIFF
--- a/singer_sdk/streams/core.py
+++ b/singer_sdk/streams/core.py
@@ -214,7 +214,8 @@ class Stream(metaclass=abc.ABCMeta):
             True if the stream uses a timestamp-based replication key.
 
         Raises:
-            InvalidReplicationKeyException: If the schema does not contain the replication key.
+            InvalidReplicationKeyException: If the schema does not contain the 
+            replication key.
         """
         if not self.replication_key:
             return False
--- a/singer_sdk/streams/core.py
+++ b/singer_sdk/streams/core.py
@@ -212,7 +212,7 @@ class Stream(metaclass=abc.ABCMeta):
 
         Returns:
             True if the stream uses a timestamp-based replication key.
-        
+
         Raises:
             InvalidReplicationKeyException: If the schema does not contain the replication key.
         """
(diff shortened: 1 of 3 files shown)


singer_sdk/streams/core.py
```diff
@@ -211,10 +212,19 @@ def is_timestamp_replication_key(self) -> bool:
...
+            InvalidReplicationKeyException: If the schema does not contain the
+                replication key.
         """
         if not self.replication_key:
             return False
         type_dict = self.schema.get("properties", {}).get(self.replication_key)
+        if type_dict is None:
+            msg = (
+                f"{self.replication_key} is not in schema for stream name: {self.name}"
+            )
```
edgarrmondragon, line 226:
```suggestion
            msg = (
                f"Field '{self.replication_key}' is not in schema for stream '{self.name}'"
            )
```

edgarrmondragon (commented):
Thanks for contributing @mjsqu! Just a small nit.

tests/core/test_streams.py
```diff
@@ -275,6 +278,23 @@ def test_stream_starting_timestamp(
...
+
+    class InvalidReplicationKeyStream(SimpleTestStream):
+        replication_key = "INVALID"
+
+    stream = InvalidReplicationKeyStream(tap)
+
+    with pytest.raises(
+        InvalidReplicationKeyException,
+        match=(
+            f"{stream.replication_key} is not in schema for stream name: {stream.name}"
```
edgarrmondragon, line 292:
```suggestion
            f"Field '{stream.replication_key}' is not in schema for stream '{stream.name}'"
```

edgarrmondragon (approved):

edgarrmondragon:
Thanks @mjsqu!

PR #1940, opened 2023-09-07
fix: Fix pytest plugin declaration so it can be used without requiring using `pytest_plugins`

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1940.org.readthedocs.build/en/1940/


edgarrmondragon:
This is a breaking change, so I'm moving it to the v1 branch

edgarrmondragon:
https://github.com/meltano/sdk/pull/1943

PR #1943, opened 2023-09-07
fix: Fix pytest plugin declaration so it can be used without requiring `pytest_plugins`

edgarrmondragon:


edgarrmondragon:
@kgpayne This removes the need for users to do

```python
# tests/conftest.py

pytest_plugins = ("singer_sdk.testing.pytest_plugin",)
```

PR #1951, opened 2023-09-11
fix: Ensure SQL streams are sorted when a replication key is set

edgarrmondragon:
----
:books: Documentation preview :books:: https://meltano-sdk--1951.org.readthedocs.build/en/1951/

### DIFF
--- a/tests/samples/test_tap_sqlite.py
+++ b/tests/samples/test_tap_sqlite.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import json
 import typing as t
 
+import pytest
 from click.testing import CliRunner
+from freezegun import freeze_time
 
 from samples.sample_tap_sqlite import SQLiteTap
 from samples.sample_target_csv.csv_target import SampleTargetCSV
@@ -11,6 +13,7 @@ from singer_sdk import SQLStream
 from singer_sdk._singerlib import MetadataMapping, StreamMetadata
 from singer_sdk.testing import (
     get_standard_tap_tests,
+    tap_sync_test,
     tap_to_target_sync_test,
 )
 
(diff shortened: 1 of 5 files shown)


edgarrmondragon:
SQLAlchemy 1.* tests are failing because inspected columns have a different order, so the SCHEMA messages, and thus the snapshots, are different.

tests/samples/test_target_sqlite.py
```diff
@@ -84,8 +85,10 @@ def test_sync_sqlite_to_sqlite(
     )
     orig_stdout.seek(0)
     tapped_config = dict(sqlite_sample_target.config)
-    catalog = _get_tap_catalog(SQLiteTap, config=tapped_config, select_all=True)
-    tapped_target = SQLiteTap(config=tapped_config, catalog=catalog)
+    tapped_target = SQLiteTap(
+        config=tapped_config,
+        catalog=sqlite_sample_db_catalog.to_dict(),
+    )
```
edgarrmondragon, line 91: The data tapped from the loaded SQLite db had a different catalog, i.e. streams that are incremental during the first sync, are "full table" in the downstream sync, so the message count was different and the test was failing.

tests/samples/test_tap_sqlite.py
```diff
@@ -116,3 +119,24 @@ def test_sync_sqlite_to_csv(sqlite_sample_tap: SQLTap, tmp_path: Path):
...
+            state_messages.append(message)
+
+    return state_messages
+
+
+def test_sqlite_state(sqlite_sample_tap_state_messages):
+    assert all(
+        "progress_markers" not in bookmark
+        for message in sqlite_sample_tap_state_messages
+        for bookmark in message["value"]["bookmarks"].values()
```
edgarrmondragon, line 141: This ensures that the stream is resumable

pnadolny13 (approved):
@edgarrmondragon this looks good to me!

kgpayne (approved):

PR #1954, opened 2023-09-13
fix(targets): missing begin()s related to SQLAlchemy 2.0

andyoneal:
adds begin() to three other functions where schema is added or columns are altered. because of the changes made to auto-commit in SQLAlchemy 2.0, without begin() the ddl is rolled back.

matches style used in _create_empty_column()

----
:books: Documentation preview :books:: https://meltano-sdk--1954.org.readthedocs.build/en/1954/


edgarrmondragon (approved):
Thanks @andyoneal! This revealed some untested paths so I pushed tests to your branch.

Issue #1961, opened 2023-09-15
bug: schema comparison before adding metadata columns


pnadolny13:
https://github.com/meltano/sdk/issues/1777 was resolved by https://github.com/meltano/sdk/pull/1778 but theres a very similar bug in [_process_schema_message](https://github.com/meltano/sdk/blob/97d97de313511be4c21b842bf2ba0283d4bf5c02/singer_sdk/target_base.py#L377C1-L378C1) in the target where the comparison always returns false when SDC metadata columns are included.

My shopify stream has a bunch of schema messages that cause it to constantly log the fact that it needs to reinitialize a sink because the schema changed. When I added a breakpoint I found that the only difference is the SDC properties.

```
2023-09-15 16:09:49,858 | INFO     | target-postgres      | Initializing target sink for stream 'shopify_transactions'...
2023-09-15 16:09:50,537 | INFO     | target-postgres      | Schema has changed for stream 'shopify_transactions'. Mapping definitions will be reset.
2023-09-15 16:09:50,537 | INFO     | target-postgres      | Schema or key properties for 'shopify_transactions' stream have changed. Initializing a new 'shopify_transactions' sink...
2023-09-15 16:09:50,537 | INFO     | target-postgres      | Initializing 'target-postgres' target sink...
2023-09-15 16:09:50,537 | INFO     | target-postgres      | Initializing target sink for stream 'shopify_transactions'...
2023-09-15 16:09:53,004 | INFO     | target-postgres      | Schema has changed for stream 'shopify_transactions'. Mapping definitions will be reset.
2023-09-15 16:09:53,004 | INFO     | target-postgres      | Schema or key properties for 'shopify_transactions' stream have changed. Initializing a new 'shopify_transactions' sink...
2023-09-15 16:09:53,004 | INFO     | target-postgres      | Initializing 'target-postgres' target sink...
2023-09-15 16:09:53,004 | INFO     | target-postgres      | Initializing target sink for stream 'shopify_transactions'...
2023-09-15 16:09:54,800 | INFO     | target-postgres      | Schema has changed for stream 'shopify_transactions'. Mapping definitions will be reset.
2023-09-15 16:09:54,800 | INFO     | target-postgres      | Schema or key properties for 'shopify_transactions' stream have changed. Initializing a new 'shopify_transactions' sink...
2023-09-15 16:09:54,800 | INFO     | target-postgres      | Initializing 'target-postgres' target sink...
[...]

PR #1962, opened 2023-09-15
perf: Ensure `raw_schema` in stream mapper is immutable

pnadolny13:
Closes https://github.com/meltano/sdk/issues/1961

When trying to figure out how to fix https://github.com/meltano/sdk/issues/1961 I realized it was caused by the [raw_schema](https://github.com/meltano/sdk/blob/97d97de313511be4c21b842bf2ba0283d4bf5c02/singer_sdk/mapper.py#L87) in the mapper being a shared object with the [transformed_schema](https://github.com/meltano/sdk/blob/97d97de313511be4c21b842bf2ba0283d4bf5c02/singer_sdk/mapper.py#L89) which is passed to `get_sink` then `add_sink`, ultimately to the sink itself [where SDC columns are added](https://github.com/meltano/sdk/blob/97d97de313511be4c21b842bf2ba0283d4bf5c02/singer_sdk/sinks/core.py#L76). The rest of the codebase expects transformed_schema to be mutable but raw_schema to be immutable (or at least it reads that way to me). They both share the same reference to the original schema dict so assumptions break. The main case I ran into was in https://github.com/meltano/sdk/issues/1961 where target-postgres was constantly creating new sink instances for me when the schema wasnt actually changing.

### DIFF
--- a/singer_sdk/mapper.py
+++ b/singer_sdk/mapper.py
@@ -84,7 +84,7 @@ class StreamMap(metaclass=abc.ABCMeta):
             flattening_options: Flattening options, or None to skip flattening.
         """
         self.stream_alias = stream_alias
-        self.raw_schema = raw_schema
+        self.raw_schema = copy.deepcopy(raw_schema)
         self.raw_key_properties = key_properties
         self.transformed_schema = raw_schema
         self.transformed_key_properties = key_properties


edgarrmondragon:
This reduced running time from 15s to 3s in my very simple load test 😮

edgarrmondragon (approved):

edgarrmondragon:
Included in https://pypi.org/project/singer-sdk/0.32.0b2/

PR #2003, opened 2023-10-07
feat(taps): add ability to do list comprehensions in stream map expressions

### LINKED ISSUE #2002: feat: add ability to perform comprehensions in stream maps
### Feature scope

### Description

It would be helpful to have a way to parse out specific information from json fields. Currently my team is working on an integration where some information in an array of objects are considered too sensitive to bring into the data warehouse.

For instance, if a field `custom_fields` in stream `test_data` contains the json object `[{"key": "a", "value": 1}, {"key": "b", "value": 2}, {"key": "c", "value": 10}]`, and the fields `b` & `c` are deemed sensitive and should not be brought into the data warehouse, then the following stream map definition would be ideal.

```
stream_maps:
    test_data:
        custom_fields: __NULL__
        custom_field_dict: "[x['value'] for x in custom_fields if x['key'] == 'a'][0]"
```

however this is not currently possible, it fails with the error

```
singer_sdk.exceptions.MapExpressionError: Failed to evaluate simpleeval expressions [x['value'] for x in custom_fields if x['key'] == 'a'][0]
```

Adding support for this should be fairly easy. `simpleeval` supports compound types through the `EvalWithCompoundTypes` class.

I believe the following changes would have to be made in `mapper.py`

```
# add new function
def compound_eval(expr, operators=None, functions=None, names=None):
    s = simpleeval.EvalWithCompoundTypes(operators=operators, functions=functions, names=names)
    return s.eval(expr)
```

```
# replace this (328 - 332):
            result: str | int | float = simpleeval.simple_eval(
                expr,
                functions=self.functions,
                names=names,
            )
# with this:
            result: str | int | float = compound_eval(
                expr,
                functions=self.functions,
                names=names,
            )
```

haleemur:
Closes #2002

* add `simpleeval` as a dependency & remove locally vendored `simpleeval` (since simpleeval itself moved to poetry and is no longer dependent on deprecated packages https://gitlab.com/meltano/sdk/-/issues/213)
* use `ast` to parse expressions used for filtering and property mapping in stream_maps.
* use `simpleeval.EvalWithCompoundTypes` instead of `simpleEval.SimpleEval`
* add `bool` as a new transformed type for mapped properties
* add test for list comprehension in mapped properties.
* update tap test factory to check against final record schema (after stream maps are applied)
* add a performance test through https://github.com/meltano/sdk/pull/2003

----
:books: Documentation preview :books:: https://meltano-sdk--2003.org.readthedocs.build/en/2003/


edgarrmondragon:
Hey @haleemur,

Thanks for contributing! Ping me when this is ready to review 🙂

tayloramurphy:
Thanks for the contribution @haleemur ! @edgarrmondragon can you open a separate issue for adding this as an example to https://sdk.meltano.com/en/latest/stream_maps.html ?

tests/core/test_simpleeval.py
edgarrmondragon, line 1: This test file was also vendored, so it can probably be removed if we're going forward adding the dependency.

edgarrmondragon (changes requested):

haleemur:
@edgarrmondragon 

I'll take a look at why mypy is complaining on 3.8, but otherwise, I think the PR is ready.

When looking at the benchmark run in the workflows a significant improvement is noted (likely due to parsing only once)

* [current branch median](https://github.com/meltano/sdk/actions/runs/6474806593/job/17580538748?pr=2003#step:7:532): 174.63
* [main branch median](https://github.com/meltano/sdk/actions/runs/6474067943/job/17578183393#step:7:624): 656.25
* the benchmark median-duration is 3.7 times shorter


Locally run benchmarks reveal a similar story.

`meltano:main` locally run benchmark
```
------------------------------------------------------- benchmark: 1 tests ------------------------------------------------------
Name (time in ms)                         Min       Max      Mean  StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------
test_bench_simple_map_transforms     318.3243  320.8534  319.9556  0.9875  320.2287  1.1652       1;0  3.1254       5           1
---------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================================================= 1 passed, 473 skipped, 2 deselected, 8 warnings in 3.55s =============================================================

```

`haleemur:simpleeval-compound-data-types` locally run benchmark
```
----------------------------------------------------- benchmark: 1 tests -----------------------------------------------------
Name (time in ms)                        Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------
test_bench_simple_map_transforms     84.7507  86.4868  85.4988  0.5254  85.5117  0.7484       3;0  11.6961      12           1
------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================================================= 1 passed, 390 skipped, 2 deselected, 8 warnings in 2.48s =============================================================
```

edgarrmondragon (approved):
This is amazing. Thanks @haleemur!

PR #2011, opened 2023-10-10
test(taps): Add benchmark test on map transforms via `pytest-benchmark`

haleemur:
relates to https://github.com/meltano/sdk/issues/2002.

As suggested in https://github.com/meltano/sdk/issues/2002#issuecomment-1751826751, 

This PR adds a benchmark test using pytest-benchmark to test the performance of stream maps using simpleeval, and ensure that we do not harm application performance by using `EvalWithCompoundTypes`.

The PR does the following:
* separate some logic into function `_run_transform` out of `_test_transform` to use for benchmarking
* add a benchmark test for the `repositories` stream with stream size of 4000 records
* add `pytest-benchmark` as a development dependency 
* add nox session for benchmarking (`benches`)

_I'm less sure if I've implemented the following two correctly as I'm not familiar with github actions_

* add `benches to the github-actions workflow matrix
* add action to upload benchmark results

----
:books: Documentation preview :books:: https://meltano-sdk--2011.org.readthedocs.build/en/2011/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -101,11 +101,19 @@ jobs:
         name: coverage-data
         path: ".coverage.*"
 
-    - name: Upload benchmark result
-      uses: benchmark-action/github-action-benchmark@v1
-      if: always() && (matrix.session == "benches")
+    - name: Comment with Pytest-Benchmark Result
+      if: always() && (matrix.session == 'benches')
+      uses: nils-braun/pytest-benchmark-commenter@v2.0.2
       with:
-        output-file-path: "bench.json"
+        token: ${{ secrets.GITHUB_TOKEN }}
+        benchmark-file: output.json
+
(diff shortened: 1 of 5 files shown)


haleemur:
@edgarrmondragon I added the benchmark test, but having some trouble with the github action to post the benchmark result as a comment on the PR

edgarrmondragon (approved):

edgarrmondragon:
> @edgarrmondragon I added the benchmark test, but having some trouble with the github action to post the benchmark result as a comment on the PR
> 
> for now, am removing the gh action to try to comment on the PR

Thank you so much for the contribution @haleemur! I think this is good enough as a first iteration. I think I can integrate [Codspeed](https://docs.codspeed.io/benchmarks/python) with very minor changes to comment in PRs.

PR #2018, opened 2023-10-11
feat(templates): Drop Python 3.7 support in cookiecutter templates

visch:
----
:books: Documentation preview :books:: https://meltano-sdk--2018.org.readthedocs.build/en/2018/


edgarrmondragon:
@visch I'm not against, but why? Is it blocking something? 😅

cookiecutter/mapper-template/{{cookiecutter.mapper_id}}/pyproject.toml
```diff
@@ -21,7 +21,7 @@ packages = [
 {%- endif %}
 
 [tool.poetry.dependencies]
-python = ">=3.7.1,<4"
+python = ">=3.8.1,<4"
```
edgarrmondragon, line 24: Why not

```suggestion
python = ">=3.8,<4"
```

visch:
> @visch I'm not against, but why? Is it blocking something? 😅
> 
> I ask because I've been reluctant to backport #1837 from the `v1` branch to `main` because 3.7 is still ~1/3 of downloads:
> 
> sdk-monthly-downloads

Not a huge deal for me I just hit that a down stream depending i wanted to use for my tap required python 3.8 or higher (latest version of fsspec atleast). Since I had to update the cookie cutter on my side I figured it'd help some others down the road that's it!

visch:
> @visch I'm not against, but why? Is it blocking something? 😅
> 
> I ask because I've been reluctant to backport #1837 from the `v1` branch to `main` because 3.7 is still ~1/3 of downloads:
> 
> sdk-monthly-downloads

Just remembered one other thing, I didn't touch the main pyproject toml file exactly for the reason you mentioned :)  so we don't break anyone here but anyone new has a later version of Python

edgarrmondragon:
> I just hit that a down stream dependency i wanted to use for my tap required python 3.8 or higher (latest version of fsspec atleast)

Yeah, I've hit those too and Poetry doesn't have good hints about what to do in those cases because it tries to add the latest version by default. I've found that adding the dependency manually and pointing to a version _right before_ 3.7 support was dropped, works fine.

For the case of `fsspec` that'd be [2023.3.0](https://github.com/fsspec/filesystem_spec/releases/tag/2023.3.0).

That said, it probably doesn't hurt to start expecting new packages built with the templates to require Python >= 3.8.

cookiecutter/target-template/{{cookiecutter.target_id}}/pyproject.toml
```diff
@@ -20,7 +20,7 @@ packages = [
 {%- endif %}
 
 [tool.poetry.dependencies]
-python = ">=3.7.1,<4"
+python = ">=3.8.1,<4"
```
edgarrmondragon, line 23:
```suggestion
python = ">=3.8,<4"
```

cookiecutter/tap-template/{{cookiecutter.tap_id}}/pyproject.toml
```diff
@@ -20,7 +20,7 @@ packages = [
 {%- endif %}
 
 [tool.poetry.dependencies]
-python = ">=3.7.1,<4"
+python = ">=3.8.1,<4"
```
edgarrmondragon, line 23:
```suggestion
python = ">=3.8,<4"
```

edgarrmondragon (approved):

PR #2035, opened 2023-11-08
fix: retry all 5xx status codes

asamasoma:
The documentation for the `validate_response` function states that 5xx error codes should raise a `RetriableAPIError`. 

However, because `max(HTTPStatus)` evaluates to 511, the current implementation does not raise the expected error for valid custom error codes such as Cloudflare's "521 Web Server Is Down".

This PR removes the cap on `RetriableAPIError` status codes so that valid custom error codes will raise the expected error.


edgarrmondragon (approved):
Thanks @asamasoma!

asamasoma:
> Thanks @asamasoma!

Wow, fastest merge ever! Thanks for accepting my PR!

PR #2041, opened 2023-11-12
fix: converting floats to decimals before schema validation

taylorbarstow:
Fixes a [known challenge](https://github.com/python-jsonschema/jsonschema/issues/185) with certain validations of floating points using JSON schema, by converting floats to decimals prior to validation. 

For example, this fixes errors like: 

```
ValidationError: 92.6 is not a multiple of 0.1
```

There is a relevant discussion here: https://gitlab.com/meltano/target-csv/-/issues/3

Certain targets (for example: [gupy-io/target-s3-parquet](https://github.com/gupy-io/target-s3-parquet)) rely on this project to do schema validation, so by handling this here, developers can continue to use that pattern while also avoiding these challenges with floating point validation. 

----
:books: Documentation preview :books:: https://meltano-sdk--2041.org.readthedocs.build/en/2041/

### DIFF
--- a/singer_sdk/helpers/_typing.py
+++ b/singer_sdk/helpers/_typing.py
@@ -6,6 +6,7 @@ import copy
 import datetime
 import logging
 import typing as t
+from decimal import Decimal
 from enum import Enum
 from functools import lru_cache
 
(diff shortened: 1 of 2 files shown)


edgarrmondragon:
This seems to be intended to fix https://github.com/gupy-io/target-s3-parquet.

Repeating here my comment from Slack:

<blockquote>
<p>
That target seems to be using an old version of the SDK: <a href="https://github.com/gupy-io/target-s3-parquet/blob/a8022b4eef8cf0e803d579a6e8c464d5f5df9b31/pyproject.toml#L15">https://github.com/gupy-io/target-s3-parquet/blob/a8022b4eef8cf0e803d579a6e8c464d5f5df9b31/pyproject.toml#L15</a>.
</p>

<p>
My guess is that simply bumping the dependency will fix this issue, since this was addressed in #1809 and shipped with <a href="https://github.com/meltano/sdk/releases/tag/v0.29.0">v0.29.0</a>.
</p>
</blockquote>

PR #2054, opened 2023-11-17
test: Benchmarks for record datetime parsing

### LINKED ISSUE #2048: chore: Set up benchmarks for record and/or datetime parsing
Similar to https://github.com/meltano/sdk/pull/2011

BuzzCutNorman:
Closes #2048

### DIFF
--- a/tests/core/sinks/test_validation.py
+++ b/tests/core/sinks/test_validation.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import datetime
 
+import pytest
+
 from tests.conftest import BatchSinkMock, TargetMock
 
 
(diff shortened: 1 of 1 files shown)


BuzzCutNorman:
@edgarrmondragon  I am not quite sure why `test_validate_record` is passing while `test_bench_parse_timestamps_in_record` and `test_bench_validate_and_parse` are failing.  I would think all three would pass or fail the same way.  When you have some time would you please review the `codspeed` errors.

tests/core/sinks/test_validation.py
```diff
@@ -39,3 +41,62 @@ def test_validate_record():
...
+    number_of_runs = 10000
+
+    sink: BatchSinkMock = bench_sink
+    record: dict = bench_record
+
+    def run_parse_timestamps_in_record():
+        for _ in range(number_of_runs):
+            _ = sink._parse_timestamps_in_record(
+                record, sink.schema, sink.datetime_error_treatment
+            )
```
edgarrmondragon, line 86: > @edgarrmondragon I am not quite sure why `test_validate_record` is passing while `test_bench_parse_timestamps_in_record` and `test_bench_validate_and_parse` are failing. I would think all three would pass or fail the same way. When you have some time would you please review the `codspeed` errors.

@BuzzCutNorman I think date-time strings are parsed _in place_, so the first time `_parse_timestamps_in_record` runs they're converted, then the second call fails because they're no longer strings. Does that make sense?

BuzzCutNorman: @edgarrmondragon that does make sense.  I just tried using `copy.copy`, `copy.deepcopy` and also moving them inside the function.  I am still seeing the error.  I think still not correctly isolating `record` from getting updated between bench tests.

BuzzCutNorman: I think I see it now.

BuzzCutNorman: @edgarrmondragon Please take a look and let me know what you think. 😊

edgarrmondragon (approved):
Nice. Thanks a lot @BuzzCutNorman!

PR #2058, opened 2023-11-20
refactor: Implement a `SingerWriter` class in `singer_sdk.io_base` and use it to emit Singer messages

BuzzCutNorman:
Creates a `SingerWriter` class in `singer_sdk.io_base` which the write_message(), and format_message functions from singer_sdk._singerlib.messages are placed into. After doinng this all the stdin and stdout methods, json encoders, and json decoder would be in one file.  SingerWriter was added as an inherited class of Tap just like SingerReader is an inherited class of Target.

----
:books: Documentation preview :books:: https://meltano-sdk--2058.org.readthedocs.build/en/2058/


singer_sdk/_singerlib/messages.py
```diff
@@ -200,25 +198,3 @@ class ActivateVersionMessage(Message):
...
-
-
-def write_message(message: Message) -> None:
-    """Write a message to stdout.
-
-    Args:
-        message: The message to write.
-    """
-    sys.stdout.write(format_message(message) + "\n")
-    sys.stdout.flush()
```
edgarrmondragon, line 224: These are arguably public API, so I don't want to remove them altogether

What do you think of inverting things here: have `SingerWriter` use these public functions.

I think we can have an intermediate generic "encoder" interface, with a default JSON implementation class and instance that is used by `format_message`:

```python
class BaseSingerMessageEncoder(ABC):
  def encode(self, message):
    ...

class JSONEncoder(BaseSingerMessageEncoder):
  ...

DEFAULT_ENCODER = JSONEncoder(...)

def format_message(message):
  return DEFAULT_ENCODER.encode(message)
```

edgarrmondragon (changes requested):

edgarrmondragon (approved):
Thanks @BuzzCutNorman!

PR #2060, opened 2023-11-21
refactor(templates): Simplify tap template file names with `post_gen_project.py` hook

vicmattos:
----
:books: Documentation preview :books:: https://meltano-sdk--2060.org.readthedocs.build/en/2060/


edgarrmondragon:
Thanks @vicmattos and congrats on your first contribution!

There seems to be a few things causing CI failures:

* `cookiecutter/tap-template/{{cookiecutter.tap_id}}/.github/workflows/test.yml` is failing causes it's not valid YAML, which is ok since it's a template but we have to configure the pre-commit hook to ignore it in: https://github.com/meltano/sdk/blob/206c38ebfc68dc88dced16f47b76d98e32f77ce2/.pre-commit-config.yaml#L16-L21
* The linter is complaining about some of the imports, and some lines are longer than the allowed limit. It might be worth running the template lint checks locally to iterate over and fix the errors, (hint: run `nox -rs test_cookiecutter`).

cookiecutter/tap-template/hooks/post_gen_project.py
```diff
@@ -0,0 +1,35 @@
...
+BASE_PATH = Path('{{cookiecutter.library_name}}')
+
+
+def delete_folder(pth: Path):
+    for sub in pth.iterdir():
+        if sub.is_dir():
+            delete_folder(sub)
+        else:
+            sub.unlink()
+    pth.rmdir()
```
edgarrmondragon, line 14: Why not use `shutil.rmtree`?

vicmattos: If I'm being honest, it's because I've been indoctrinated to never use `shutil` and got used to doing like this. 
It's much more concise with it. Thanks for pointing it out!

vicmattos:
Hey @edgarrmondragon! Thanks for your welcoming and feedback.

I've updated the `.pre-commit-config.yaml` to exclude the template files for CI.

I've also removed the changes I made inside the clients files. My main goal was to simplify navigating in the cookiecutter template. I can go back to simplifying their contents later.

I've ran the nox automation and seems that everything is correct. Could you double check?

edgarrmondragon (approved):

edgarrmondragon:
Thanks @vicmattos!

edgarrmondragon:
> I've also removed the changes I made inside the client files. My main goal was to simplify navigating in the cookiecutter template. I can go back to simplify their contents later.

That makes sense, do feel free to send any follow-on PRs 👍

PR #2066, opened 2023-11-22
perf: Validate records with `fastjsonschema`

### LINKED ISSUE #2045: refactor: Use `fastjsonschema` to validate records
### Feature scope

Other

### Description

[`fastjsonschema`](https://github.com/horejsek/python-fastjsonschema) seems to perform better than [`jsonschema`](https://github.com/python-jsonschema/jsonschema/).
```[tasklist]
### Tasks
- [ ] https://github.com/meltano/sdk/issues/2047
```

BuzzCutNorman:
Add dependency of `fastjsonshema` and refactor of `Sink` and `Sink._validate_and_parse` utilize `fastjsonschema`.

Closes #2045

----
:books: Documentation preview :books:: https://meltano-sdk--2066.org.readthedocs.build/en/2066/


BuzzCutNorman:
I have run into a `date-time` and `time` problem.  If a `date-time` or `time` does not end in with a time zone like `+00:00` or `Z` it is seen as not valid.

```
>>>
>>>
>>> import fastjsonschema
>>>
>>> validator = fastjsonschema.compile({
... "type": "object",
... "properties": {
... "id": {"type": "integer"},
... "datetime_col": {"type": "string", "format": "date-time"},
... "date_col": {"type": "string", "format": "date"},
... "time_col": {"type": "string", "format": "time"},
... },
... "required": ["id"],
... })
>>>
>>>
>>> validator({"id": 1, "datetime_col": "2023-01-03T08:10:00.000000+00:00", "date_col": "2023-01-03", "time_col": "08:10:00.000000+00:00"})
{'id': 1, 'datetime_col': '2023-01-03T08:10:00.000000+00:00', 'date_col': '2023-01-03', 'time_col': '08:10:00.000000+00:00'}
>>>
>>>
>>> validator({"id": 2, "datetime_col": "2023-01-03T08:10:00.000000Z", "date_col": "2023-01-03", "time_col": "08:10:00.000000Z"})
{'id': 2, 'datetime_col': '2023-01-03T08:10:00.000000Z', 'date_col': '2023-01-03', 'time_col': '08:10:00.000000Z'}
>>>
>>>
>>> validator({"id": 3, "datetime_col": "2023-01-03T08:10:00.000000", "date_col": "2023-01-03", "time_col": "08:10:00.000000"})
  File "<stdin>", line 1, in <module>
  File "<string>", line 24, in validate
fastjsonschema.exceptions.JsonSchemaValueException: data.datetime_col must be date-time
>>>
```

CC: @edgarrmondragon

edgarrmondragon:
> Going back to the `JSONSchema` documentation I found that `date-time`, `date`, and `time` values should conform to `RFC 3339, section 5.6`. Reading through the specification I found a that `RFC 3339, section 4.3` states how to deal with unknown local offset. In these instances, you represent the offset with `-00:00`. I have found that `fastjsonschema` finds `date-time` and `time` with the `-00:00` offset as valid.

I see. This makes me a bit worried that this may be a breaking change. Targets may be currently doing some false negatives and letting values like `2023-01-03T00:10:00.0000` go through.

BuzzCutNorman:
> I see. This makes me a bit worried that this may be a breaking change. Targets may be currently doing some false negatives and letting values like 2023-01-03T00:10:00.0000 go through.

@edgarrmondragon , I do have a test table (StackOverflow postlinks table) in which that very scenario is happening.  Dates with no time zone pass validation using `jsonschema` but are failing validation with `fastjsonschema`.  

I did some testing yesterday and even if we could append `-00:00` to `date-time` or `time` strings so they pass `fastjsonschema` validation, once the string is processed by `.fromisoformat` a `datetime` or `time` with the `UTC` time zone is generated instead of a `datetime` or `time` no time zone.  That solution will not work.

This morning I found that you can pass a custom format to replace a built-in format. I tested a regex string for iso datetime with an optional time zone which resolved the before mention issue with my test table. I also tested a simple lamda function that passed `date-time` when `value == value`.  There seems to be no performance impact by doing this.  There is also a flag to disable format checking `use_formats=False`.

BuzzCutNorman:
@edgarrmondragon I took a go at incorporating the changes you wrote in PR1471 into this PR.  When you get a chance, please take a look and let me know what you think.

edgarrmondragon (commented):
@BuzzCutNorman I like the approach of defaulting to no format checks since are the most problematic. We probably want consider an update to https://github.com/meltano/sdk/pull/1711, but that's outside of the scope of this issue.

tests/core/sinks/test_validation.py
```diff
@@ -83,7 +83,7 @@ def bench_record():
         "id": 1,
         "created_at": "2021-01-01T00:08:00-07:00",
         "updated_at": "2022-01-02T00:09:00-07:00",
-        "deleted_at": "2023-01-03T00:10:00.0000",
+        "deleted_at": "2023-01-03T00:10:00.0000-00:00",
```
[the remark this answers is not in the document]
BuzzCutNorman: Good call, the appended `-00:00` is no longer needed.

singer_sdk/sinks/core.py
```diff
@@ -330,7 +351,10 @@ def _validate_and_parse(self, record: dict) -> dict:
         Returns:
             TODO
         """
-        self._validator.validate(record)
+        try:
+            self._validator(record)
+        except fastjsonschema.JsonSchemaValueException as e:
+            self.logger.info("Record Message Validation Error: %s", e.message)
```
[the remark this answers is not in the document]
BuzzCutNorman: A covering test has been added to `test_validation`.  I changed the except in `Sink._validate_and_parse` to put the messages in variable error_message string which is given to `self.logger.info` to make the testing assertion of the full message cleaner.

singer_sdk/sinks/core.py
```diff
@@ -51,6 +49,8 @@ class Sink(metaclass=abc.ABCMeta):
     logger: Logger
 
     MAX_SIZE_DEFAULT = 10000
+    CHECK_RECORD_FORMATS = False
```
BuzzCutNorman, line 52: @edgarrmondragon I thinking this need to not be a constant since developers will be changing it as needed.  I was thinking of changing it to `allow_record_format_cheks`.  Should it stay here or move?  Please let me know what you think.

tests/core/sinks/test_format_checker.py
```diff
@@ -0,0 +1,223 @@
...
+                ),
+            }
+
+        def process_batch(self, context: dict) -> None:
+            pass
+
+        def process_record(self, record: dict, context: dict) -> None:
+            pass
+
+    return CustomSink(target, "test_stream", test_schema, None)
```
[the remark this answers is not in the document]
BuzzCutNorman: I added this but now I am getting error `fastjsonschema.exceptions.JsonSchemaDefinitionException: Unknown format: date`.  The `date` and `time` formats are part of Draft7.  I think validators are supposed to allow formats if it doesn't know of them?

>  A JSON Schema validator will ignore any format type that it does not understand.
https://json-schema.org/understanding-json-schema/reference/string#format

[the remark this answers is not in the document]
BuzzCutNorman: If I change `CHECK_RECORD_FORMATS = False` and found that this seems to turn off both schema and record format checks since I didn't get the error message anymore.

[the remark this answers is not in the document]
edgarrmondragon: Ah gotcha, how about adding `"$schema": "http://json-schema.org/draft-07/schema"` to the draft7 sink?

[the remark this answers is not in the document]
BuzzCutNorman: I can but `fastpythonschema` defaults to draft7 when no schema is specified.  Since I haven't seen `$schema` in any of schema messages generated by SDK taps I was reluctant to add it.

> When not specified, the latest implementation is used (draft-07).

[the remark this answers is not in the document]
edgarrmondragon: Ah gotcha, I missed that. Disregard then 😄.

singer_sdk/sinks/core.py
edgarrmondragon: Agreed. How about `validate_field_string_format`?

singer_sdk/sinks/core.py
```diff
@@ -52,6 +50,9 @@ class Sink(metaclass=abc.ABCMeta):
 
     MAX_SIZE_DEFAULT = 10000
 
+    record_validator = fastjsonschema
+    validate_field_string_format = False
```
edgarrmondragon, line 54: Let's make sure this shows up in the docs:

```suggestion
    validate_field_string_format = False
    """Enable format validation, for example for `date-time` string fields."""
```

edgarrmondragon: Yup, nice!

https://meltano-sdk--2066.org.readthedocs.build/en/2066/classes/singer_sdk.Sink.html#singer_sdk.Sink.validate_field_string_format

singer_sdk/sinks/core.py
```diff
@@ -52,6 +50,12 @@ class Sink(metaclass=abc.ABCMeta):
 
     MAX_SIZE_DEFAULT = 10000
 
+    record_validator = fastjsonschema
+    """Default JSONSschema validation library, Developers can override."""
```
edgarrmondragon, line 54: This doesn't look as great.

Screenshot 2023-12-01 at 1 48 04 p m

Either we shouldn't document it, or we should create a wrapper that implements certain interface (e.g. RecordValidator) but it's not too useful to document it as is.

Wdyt, the interface thing is probably out of scope?

BuzzCutNorman: @edgarrmondragon I added add an interface.  Please take a look and let me know what you think.

singer_sdk/sinks/core.py
```diff
@@ -95,11 +131,32 @@ def __init__(
         self._batch_records_read: int = 0
         self._batch_dupe_records_merged: int = 0
 
-        self._validator = Draft7Validator(
+        self._validator: RecordValidator = self.record_validator(
```
edgarrmondragon, line 134:
```suggestion
        self._validator: BaseRecordValidator | None = self.get_record_validator()
```

singer_sdk/sinks/core.py
```diff
@@ -95,11 +131,32 @@ def __init__(
...
+            .. validate = fastjsonschema.compile(definition, formats={
+                    'foo': r'foo|bar',
+                    'bar': lambda value: value in ('foo', 'bar'),
+                })
+
+        Returns:
+            A dictionary containing regex strings and callables.
+            See: https://horejsek.github.io/python-fastjsonschema/.
+        """
+        return {}
```
edgarrmondragon, line 158:
```suggestion
    def get_record_validator(self) -> BaseRecordValidator | None:
        """Get a record validator for this sink.

        Override this method to use a custom format validator or disable record validator, by returning `None`.

        Returns:
            An instance of a subclass of ``BaseRecordValidator``.
        """
        return FastJSONSchemaValidator(
            schema,
validate_formats=self.validate_field_string_format,
            format_validators={},
        )
```

singer_sdk/sinks/core.py
```diff
@@ -95,11 +131,32 @@ def __init__(
         self._batch_records_read: int = 0
         self._batch_dupe_records_merged: int = 0
 
-        self._validator = Draft7Validator(
+        self._validator: RecordValidator = self.record_validator(
             schema,
```
edgarrmondragon, line 135:
```suggestion
```

singer_sdk/sinks/core.py
```diff
@@ -95,11 +131,32 @@ def __init__(
...
-            format_checker=Draft7Validator.FORMAT_CHECKER,
+            validate_formts=self.validate_field_string_format,
+            validator_formats=self.get_validator_formats(),
```
edgarrmondragon, line 137:
```suggestion
```

singer_sdk/sinks/core.py
```diff
@@ -329,7 +386,11 @@ def _validate_and_parse(self, record: dict) -> dict:
         Returns:
             TODO
         """
-        self._validator.validate(record)
+        try:
+            self._validator.validate(record)
+        except fastjsonschema.JsonSchemaValueException as e:
+            error_message: str = f"Record Message Validation Error: {e.message}"
+            self.logger.info(error_message)
```
edgarrmondragon, line 393:
```suggestion
        except InvalidRecord:
            message = "Record message validation failed"
            self.logger.exception(message)
```

singer_sdk/sinks/core.py
```diff
@@ -329,7 +386,11 @@ def _validate_and_parse(self, record: dict) -> dict:
         Returns:
             TODO
         """
-        self._validator.validate(record)
+        try:
```
edgarrmondragon, line 389: With the proposed changes, we can add a condition to possibly skip validation:

```python
if self._validator:
    try:
        ...
```

singer_sdk/sinks/core.py
```diff
@@ -39,7 +39,39 @@
...
+
+    def validate(self, record: dict):  # noqa: ANN201
+        """Validate a record message.
+
+        Args:
+            record: Record message to validate.
+
+        Returns: Record validation results.
+        """
+        self.validator(record)
```
edgarrmondragon, line 74:
```suggestion
        try:
            self.validator(record)
        except fastjsonschema.JsonSchemaValueException as e:
            raise InvalidRecord(e.message) from e
```

singer_sdk/sinks/core.py
```diff
@@ -39,7 +39,39 @@
...
+        Args:
+            schema: Schema of the stream to sink.
+            validate_formts: Ture or False validate formats.
+            validator_formats: User defined validator formats.
+        """
+        self.validator = fastjsonschema.compile(
+            schema,
+            use_formats=validate_formts,
+            formats=validator_formats,
+        )
```
edgarrmondragon, line 64:
```suggestion
class InvalidRecord(Exception):
    pass


class BaseRecordValidator(abc.ABC):
    def __init__(schema: dict[str, t.Any]) -> None:
        self.schema = schema

    @abc.abstractmethod
    def validate(record: dict[str, t.Any]) -> None:
        ...
        

class FastJSONSchemaValidator(BaseRecordValidator):
    def __init__(
        self,
        schema: dict,
        *,
        validate_formats: bool = False,
        format_validators: dict[str, t.Callable],
    ):
        """Initialize the validator.

        Args:
            schema: Schema of the stream to sink.
            validate_formats: Whether JSON string formats (e.g. ``date-time``) should be validated.
            format_validators: User defined format validators.
        """
        super().__init__(schema)
        self.validator = fastjsonschema.compile(
            self.schema,
            use_formats=validate_formats,
            formats=format_validators,
        )
```

edgarrmondragon (changes requested):
@BuzzCutNorman I think we can hide the record validation implementation details from both the developer (through an abstract base class) and the user (through a generic exception).

BuzzCutNorman:
@edgarrmondragon Did you mean to remove the `get_validator_formats` that allows developers to set custom formats?
```
def get_validator_formats(self) -> dict:
        """Get formats for JSON schema validator.
        Override this method to add custom string format checkers to the JSON schema
        validator.
        This is useful when, for example, the target requires a specific format for a
        date or datetime field.
        Example:
            .. validate = fastjsonschema.compile(definition, formats={
                    'foo': r'foo|bar',
                    'bar': lambda value: value in ('foo', 'bar'),
                })
        Returns:
            A dictionary containing regex strings and callables
                https://horejsek.github.io/python-fastjsonschema/ 
        """
        return {}
```

edgarrmondragon:
@BuzzCutNorman I moved a few things around (e.g. exceptions) and added a few `@override` decorators to make mypy tell us when we change a method name in a parent class. CodeCov is unwilling to cooperate so I'll retry it later.

This feels really close. Thanks for your time, patience and openness!

edgarrmondragon:
@BuzzCutNorman [I've given this a try on MeltanoLabs/target-postgres](https://github.com/MeltanoLabs/target-postgres/pull/241) and it predictably fails on this test:

```python
def test_record_missing_required_property(postgres_target):
    with pytest.raises(jsonschema.exceptions.ValidationError):
        file_name = "record_missing_required_property.singer"
        singer_file_to_target(file_name, postgres_target)
```

Even changing the exception type to be `singer_sdk.exceptions.InvalidRecord` does not work because we stopped raising the exception.

So, what do you think about continue raising an exception, in particular `InvalidRecord`? That will still require a change on target-postgres, but it's the least surprising in my opinion because the sync will continue to fail at the same stage instead of further down the line when a missing key property is detected.

We still leave the door open to developers to disable schema validation with the Sink class-level attribute.

BuzzCutNorman:
> So, what do you think about we continue raising an exception, in particular InvalidRecord? That will still require a change on the target-postgres test, but it's the least surprising in my opinion because the sync will continue to fail at the same stage instead of further down the line when a missing key property is detected.
>
>We still leave the door open to developers to disable schema validation with the Sink class-level attribute.

Hmm.  I like flags (maybe to much).  What do you think about something like this

```
stop_on_invalid_record: bool = True
"""Enable/Disable the Target from stopping when JSONSchmea field validation error raised."""

        if self._validator is not None:
            # TODO: Check the performance impact of this try/except block. It runs
            # on every record, so it's probably bad and should be moved up the stack.
            try:
                self._validator.validate(record)
            except InvalidRecord as e:
                if stop_on_invaild_record
                    raise InvalidRecord(e)
                else
                    self.logger.exception("Record validation failed %s", e)
```

edgarrmondragon:
> Hmm. I like flags (maybe to much). What do you think about something like this
> 
> ```
> stop_on_invalid_record: bool = True
> """Enable/Disable the Target from stopping when JSONSchmea field validation error raised."""
> 
>         if self._validator is not None:
>             # TODO: Check the performance impact of this try/except block. It runs
>             # on every record, so it's probably bad and should be moved up the stack.
>             try:
>                 self._validator.validate(record)
>             except InvalidRecord as e:
>                 if stop_on_invaild_record
>                     raise InvalidRecord(e)
>                 else
>                     self.logger.exception("Record validation failed %s", e)
> ```
> 
> I like this idea because if you want field validation turned on and you want to see how a full load goes this is an easy way to let the developer do that while still getting the validation feedback. Plus it could be migrated easily to a config option by the developer to allow users of the target to control this behavior as well. 🤷

@BuzzCutNorman  Yeah, I like that idea too 👍

BuzzCutNorman:
@edgarrmondragon I added the flag `stop_on_field_validation_exception` and defaulted it to `True`.  When you get a chance please give it a try with against the MeltanoLabs target-postgres tests.

edgarrmondragon (approved):
LGTM @BuzzCutNorman! I'm gonna wait to merge this until probably next week, in case we have to release any patches for v0.34.0.

BuzzCutNorman:
@edgarrmondragon I was wondering if you wanted me to break out the abstraction of the JSON Schema library into its own PR?

edgarrmondragon:
Switching the codspeed workflow to run in Python 3.12 lets us see a _flamegraph diff_!!!

Screenshot 2024-01-02 at 11 19 21 p m

https://codspeed.io/meltano/sdk/branches/BuzzCutNorman:2045-validate-records-with-fastjsonschema

singer_sdk/sinks/core.py
```diff
@@ -67,46 +67,36 @@ def validate(self, record: dict[str, t.Any]) -> None:
...
-            format_checker = jsonschema.FormatChecker(formats=())
-
         try:
-            jsonschema_validator.check_schema(schema)
-        except jsonschema.SchemaError as e:
+            self.validator = fastjsonschema.compile(
+                self.schema, use_formats=validate_formats, formats=format_checker or {}
+            )
+        except fastjsonschema.JsonSchemaDefinitionException as e:
             error_message = f"Schema Validation Error: {e}"
```
edgarrmondragon, line 97: I missed this in the previous PR

```suggestion
            error_message = "Schema Validation Error"
```

that way you don't even need to update the tests.

singer_sdk/sinks/core.py
```diff
@@ -67,46 +67,36 @@ def validate(self, record: dict[str, t.Any]) -> None:
         """
 
 
-class JSONSchemaValidator(BaseJSONSchemaValidator):
+class FastJSONSchemaValidator(BaseJSONSchemaValidator):
```
edgarrmondragon, line 70: Should we keep the `JSONSchemaValidator` class? It doesn't hurt, I think.

BuzzCutNorman: I don't see that it would hurt keeping `JSONSchemaValidator`.  Let me change it back.

BuzzCutNorman:
@edgarrmondragon should we keep this PR or close it?

edgarrmondragon:
Superseded by https://github.com/meltano/sdk/pull/2231

PR #2070, opened 2023-11-27
refactor(targets): Parse dates with `datetime.fromisoformat`/`backports.datetime_fromisoformat`

### LINKED ISSUE #2046: refactor: Parse dates with a more performant library than `python-dateutil`
### Feature scope

CLI (options, error messages, logging, etc.)

### Description

Pendulum seems to perform significantly better at parsing datetimes. Doing this would improve target performance by making record parsing faster.

```[tasklist]
### Tasks
- [ ] https://github.com/meltano/sdk/issues/2048
- [ ] https://github.com/meltano/sdk/pull/2065
```

BuzzCutNorman:
Replaced `python-dateutil.parse` with `backports.datetime_fromisoformat`. 
 Changed except to ValueError.

Closes #2046

----
:books: Documentation preview :books:: https://meltano-sdk--2070.org.readthedocs.build/en/2070/


pyproject.toml
```diff
@@ -81,6 +81,7 @@ fs-s3fs = {version = ">=1.1.1", optional = true}
 # Testing dependencies installed as optional 'testing' extras
 pytest = {version=">=7.2.1", optional = true}
 pytest-durations = {version = ">=1.2.0", optional = true}
+backports-datetime-fromisoformat = "^2.0.1"
```
edgarrmondragon, line 84: We can avoid installers building this package from the source distribution on Python 3.11+ since [there are no wheels for those](https://pypi.org/project/backports-datetime-fromisoformat/#files):

```suggestion
backports-datetime-fromisoformat = { version = ">=2.0.1", python = "<3.11" }
```

And you'll probably want to run `poetry lock --no-update` after that 😉.

edgarrmondragon (commented):
I think the mypy failure can be fixed by adding `backports-datetime-fromisoformat` here:

https://github.com/buzzcutnorman/sdk/blob/a4bdd6674f5f59fda60fbad9dd01af046676dd6d/noxfile.py#L70

BuzzCutNorman:
I should have noticed this initially but `backports-datetime-fromisoformat` is not a datetime parser.  It does not look at the input and see if `datetime.datetime.fromisoformat()`, `datetime.date.fromisoformat()`, or `datetime.time.fromisoformat()` need to be run to successfully transform the value to a datetime, date, or time.  I will need to add some regex comparisons to determine which method the value should be run through to be successfully transformed.

edgarrmondragon (approved):
Thanks a bunch @BuzzCutNorman!

PR #2159, opened 2024-01-19
docs: Add doctest for `get_starting_timestamp`

edgarrmondragon:
----
📚 Documentation preview 📚: https://meltano-sdk--2159.org.readthedocs.build/en/2159/


edgarrmondragon:
erm this might be a breaking change if some return types change...

PR #2170, opened 2024-01-22
feat(taps): Generate fake data with stream maps

### LINKED ISSUE #1009: [Feature]: enhance mapper to produce realistic test data
### Feature scope

Other

### Description

As a user, I need to move data from a production database to a test database keeping realistic looking values for names, addresses, and so on, but not share actual personal information.

## proposal

Add a dependency on 'faker' library
https://github.com/joke2k/faker/blob/master/setup.py#L73-L77

Enable mapper configuration similar to md5 to mask values during a run from tap to target.

## benefits and use cases

As of today, it is possible to use a hashing function like md5() but with this feature, we will have data that "looks like" real data - especially good for test environments!

## links

https://app.slack.com/client/TFG99TU9K/C013EKWA2Q1/thread/C013EKWA2Q1-1664297456.961009

ReubenFrankel:
### Usage

The `fake` property (a `Faker` instance):

```yml
config:
  stream_maps:
    users:
      name: fake.name()
```

The `Faker` instance can be configured through `faker_config`:

#### Seed

```yml
config:
  faker_config:
    seed: 0
```

#### Locale

```yml
config:
  faker_config:
    locale: en_GB
```

#### Multiple locales
```yml
config:
  faker_config:
    locale:
    - en_GB
    - es_MX
    - ja_JP
```

---

Closes #1009

---

----
📚 Documentation preview 📚: https://meltano-sdk--2170.org.readthedocs.build/en/2170/


edgarrmondragon (commented):
Hey @ReubenFrankel!

TLDR: What do you think about making this an optional dependency and add it as an extra both to singer-sdk and the 3 cookiecutters.

**Pros:** The site packages for most tap/target installations saves on the ~10MB (uncompressed) of `faker`. I've been meaning to slim down the SDK dependencies, and though the biggest offender is sqlalchemy, faker is not small either.

**Cons:** Folks have to either install `faker` alongside the tap/target or install with the extra if it's supported, e.g. `tap-github[faker]`.

Having docs for this feature in https://sdk.meltano.com/en/latest/stream_maps.html that call out that users should install `faker` would be great.

Thanks!

ReubenFrankel:
@edgarrmondragon Sure, we can go with the extra approach. A couple of questions:

1. > **Cons:** Folks have to either install `faker` alongside the tap/target or install with the extra if it's supported, e.g. `tap-github[faker]`.

    Would that not mean that the tap developer just has to specify `singer-sdk[faker]` as a dependency (indirectly through the cookiecutter), and otherwise from a Meltano perspective the `pip_url` doesn't have to change for a plugin?
1. Are you happy with the `stream_map_config` approach for configuring the `Faker` instance? This does mean `faker.seed` and `faker.locale` become reserved config values for a tap using the `faker` extra. Alternatively, we could make it a top-level config option for taps with the `stream-maps` capability - like how the `batch` capability exposes all the `batch_config` options (although this might have to be a separate capability if `faker` becomes an extra.

ReubenFrankel:
@edgarrmondragon Sure, we can go with the extra approach. A couple of questions:

1. > **Cons:** Folks have to either install `faker` alongside the tap/target or install with the extra if it's supported, e.g. `tap-github[faker]`.

    Would that not mean that the tap developer just has to specify `singer-sdk[faker]` as a dependency (indirectly through the cookiecutter), and otherwise from a Meltano perspective the `pip_url` doesn't have to change for a plugin?
1. Are you happy with the `stream_map_config` approach for configuring the `Faker` instance? This does mean `faker.seed` and `faker.locale` become reserved config values for a tap using the `faker` extra. Alternatively, we could make it a top-level config option for taps with the `stream-maps` capability - like how the `batch` capability exposes all the `batch_config` options (although this might have to be a separate capability if `faker` becomes an extra).

edgarrmondragon:
> 1. Would that not mean that the tap developer just has to specify `singer-sdk[faker]` as a dependency (indirectly through the cookiecutter), and otherwise from a Meltano perspective the `pip_url` doesn't have to change for a plugin?

That's correct, that'd also work 👍

> 2\. Are you happy with the `stream_map_config` approach for configuring the `Faker` instance? This does mean `faker.seed` and `faker.locale` become reserved config values for a tap using the `faker` extra. Alternatively, we could make it a top-level config option for taps with the `stream-maps` capability - like how the `batch` capability exposes all the `batch_config` options (although this might have to be a separate capability if `faker` becomes an extra).

Yeah. I think I'd like them better as top-level configs.

edgarrmondragon:
@ReubenFrankel there seems to be a performance regression in setting up a mapper, reported by codspeed:
Screenshot 2024-01-29 at 10 34 32 a m

Do you think this drop in performance wouldn't be there if `faker` is not installed? (i.e. if users don't need `fake` and thus are not requesting `faker` be installed)

ReubenFrankel:
> @ReubenFrankel there seems to be a performance regression in setting up a mapper, reported by codspeed: Screenshot 2024-01-29 at 10 34 32 a m
> 
> Do you think this drop in performance wouldn't be there if `faker` is not installed? (i.e. if users don't need `fake` and thus are not requesting `faker` be installed)

When `faker` is not installed, both of these exceptions should be suppressed:

https://github.com/meltano/sdk/blob/3bad517a24588e73b4aa79a1dd9a1082cf1f464a/singer_sdk/mapper.py#L271-L272
https://github.com/meltano/sdk/blob/3bad517a24588e73b4aa79a1dd9a1082cf1f464a/singer_sdk/mapper.py#L336-L337

I would have thought that without `faker` installed, this would cause a very minimal performance delta for what is essentially two try-except statements.

edgarrmondragon:
@ReubenFrankel I've pushed 4e7ea15 to test performance without faker and it seems to get worse 😅. I confirmed by running locally with `poetry run pytest --benchmark-only tests`.

What do you think?

edgarrmondragon:
> @edgarrmondragon This was a misunderstanding on my part. I think the performance overhead of actually catching and handling exceptions is what was causing the issue here. Seems to be fixed now.
> 
> https://docs.python.org/3/faq/design.html#how-fast-are-exceptions
> 
> > A [`try`](https://docs.python.org/3/reference/compound_stmts.html#try)/[`except`](https://docs.python.org/3/reference/compound_stmts.html#except) block is extremely efficient if no exceptions are raised. Actually catching an exception is expensive.

Ah, thanks for looking into it @ReubenFrankel!

It's probably worth adding updating https://sdk.meltano.com/en/latest/stream_maps.html to instruct users on how to use the optional `faker` instance in their stream maps.

ReubenFrankel:
@edgarrmondragon What do you think of these docs additions?

ReubenFrankel:
Ah, yeah good call. I'll add both those. Regarding updating the capability:

1. Should we conditionally include the `faker_config` property on `importlib.util.find_spec("faker")`, or just add it regardless?
1. Is this documentation out-of-date? As a developer, I don't think I've ever defined those settings in a plugin I've built using the SDK, since they are automatically added given the `stream_maps` capability https://github.com/meltano/sdk/blob/fa888be4ba00f471f8c0a474800f9699a1ed3347/docs/stream_maps.md?plain=1#L93-L96

I'll also take a look at updating the cookiecutter tap and target templates to support `singer-sdk[faker]`.

edgarrmondragon:
> 1\. Should we conditionally include the `faker_config` property on `importlib.util.find_spec("faker")`, or just add it regardless?

@ReubenFrankel I think it's ok to just add it regardless. We can reiterate in the field `description` that it requires injecting `faker`.

> 2\. Sort of unrelated, but is this documentation out-of-date? As a developer, I don't think I've ever defined those settings in a plugin I've built using the SDK, since they are automatically added given the `stream_maps` capability

Ah, good catch. Feel free to get rid of that!

ReubenFrankel:
@edgarrmondragon This is ready for another look. There is something wrong with the current `duckdb` version for CI checks - looks like it [no longer exists on PyPI](https://pypi.org/project/duckdb/0.9.3.dev2938/)?

https://github.com/meltano/sdk/blob/63bb73bdeea6a40a94a691fa8fd8d24ad02f9f47/poetry.lock#L547

edgarrmondragon:
> @edgarrmondragon This is ready for another look. There is something wrong with the current `duckdb` version for CI checks - looks like it [no longer exists on PyPI](https://pypi.org/project/duckdb/0.9.3.dev2938/)?
> 
> https://github.com/meltano/sdk/blob/63bb73bdeea6a40a94a691fa8fd8d24ad02f9f47/poetry.lock#L547

Fixed on `main` by https://github.com/meltano/sdk/pull/2206 😮‍💨

edgarrmondragon (approved):
Thanks @ReubenFrankel!

ReubenFrankel:
@edgarrmondragon Thanks for finishing off the tests and docs. 👍

PR #2198, opened 2024-01-30
fix(targets): Quote column names in INSERT statement

mattwill09:
This is an update to fix how the insert into statements are being generated. If you have a source column that uses a reserved word then invalid SQL was being generated. 

Previous:
```sql
INSERT INTO analytics.foo
(id, in)
VALUES (:id, :in)
```

Updated:
```sql
INSERT INTO analytics.foo
("id", "in")
VALUES (:id, :in)
```

----
📚 Documentation preview 📚: https://meltano-sdk--2198.org.readthedocs.build/en/2198/

### DIFF
--- a/singer_sdk/sinks/sql.py
+++ b/singer_sdk/sinks/sql.py
@@ -283,7 +283,7 @@ class SQLSink(BatchSink):
         statement = dedent(
             f"""\
             INSERT INTO {full_table_name}
-            ({", ".join(quoted_name(name) for name in property_names)})
+            ({", ".join(quoted_name(name, True) for name in property_names)})
             VALUES ({", ".join([f":{name}" for name in property_names])})
             """,  # noqa: S608
         )
--- a/singer_sdk/sinks/sql.py
+++ b/singer_sdk/sinks/sql.py
@@ -10,6 +10,7 @@ from textwrap import dedent
 
 import sqlalchemy as sa
 from pendulum import now
+from sqlalchemy.sql import quoted_name
 from sqlalchemy.sql.expression import bindparam
 
 from singer_sdk.connectors import SQLConnector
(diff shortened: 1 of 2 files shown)


singer_sdk/sinks/sql.py
```diff
@@ -282,7 +282,7 @@ def generate_insert_statement(
         statement = dedent(
             f"""\
             INSERT INTO {full_table_name}
-            ({", ".join(property_names)})
+            ({", ".join([f'"{name}"' for name in property_names])})
```
edgarrmondragon, line 285: We could try using https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.quoted_name

edgarrmondragon (changes requested):
Thanks @mattwill09!

You may want to update the test in https://github.com/meltano/sdk/blob/af3b678ddda34112b56954d568fe152a7e45dfbf/tests/samples/test_target_sqlite.py#L490-L495.

singer_sdk/sinks/sql.py
mattwill09: I'm not typically a python developer, but that change looks to be easy enough. I'll make sure to get those tests fixed up too.

singer_sdk/sinks/sql.py
```diff
@@ -282,7 +283,7 @@ def generate_insert_statement(
         statement = dedent(
             f"""\
             INSERT INTO {full_table_name}
-            ({", ".join(property_names)})
+            ({", ".join(quoted_name(name) for name in property_names)})
```
edgarrmondragon, line 286:
```suggestion
            ({", ".join(quoted_name(name, True) for name in property_names)})
```

edgarrmondragon (changes requested):

singer_sdk/sinks/sql.py
```diff
@@ -282,7 +283,7 @@ def generate_insert_statement(
         statement = dedent(
             f"""\
             INSERT INTO {full_table_name}
-            ({", ".join(property_names)})
+            ({", ".join(quoted_name(name, True) for name in property_names)})
```
edgarrmondragon, line 286:
```suggestion
            ({", ".join(quoted_name(name, True) for name in property_names)})  # noqa: FBT003
```

edgarrmondragon:
Superseded by #2200. Thanks @mattwill09!

mattwill09:
Thanks for finding a proper solution for this. Much appreciated!

Issue #1904, opened 2023-08-09
feat: Option to disable schema validation in targets


pnadolny13:
The transferwise variant of [target-snowflake](https://github.com/transferwise/pipelinewise-target-snowflake) has a `validate_records` setting that is defaulted to false. In the SDK we automatically do validation and don't provide a way to turn it off. We've seen some people struggle to transition to the new meltanolabs target-snowflake because validation is failing, its likely correctly failing but the solution for users is to fix the tap which becomes too big of a barrier for them when switching.

```
(Default: False) Validate every single record message to the corresponding JSON schema.
This option is disabled by default and invalid RECORD messages will fail only at load time by Snowflake.
Enabling this option will detect invalid records earlier but could cause performance degradation.
```

cc @tayloramurphy @edgarrmondragon

edgarrmondragon:
This was made a lot easier to implement by https://github.com/meltano/sdk/issues/2136 (currently only available in prerelease v0.35.0a2).

One could override [`Sink.validate_schema`](https://sdk.meltano.com/en/v0.35.0a2/classes/singer_sdk.Sink.html#singer_sdk.Sink.validate_schema) to read from config:

```python
class MySink(Sink):
  @property
  def validate_schema(self) -> bool:
    return config.get("validate_records", False)
```

PR #2208, opened 2024-01-31
feat(targets): Allow users to disable schema validation in targets

edgarrmondragon:
Related:

* Closes https://github.com/meltano/sdk/issues/1904

----
📚 Documentation preview 📚: https://meltano-sdk--2208.org.readthedocs.build/en/2208/

### DIFF
--- a/singer_sdk/helpers/capabilities.py
+++ b/singer_sdk/helpers/capabilities.py
@@ -144,6 +144,14 @@ TARGET_HARD_DELETE_CONFIG = PropertiesList(
         default=False,
     ),
 ).to_dict()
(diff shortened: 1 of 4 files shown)


anden-akkio:
Imagine our surprise when we're figuring out how to do this and it turns out you guys merged something two weeks ago! Appreciate the implementation @edgarrmondragon, keep up the great work!

Context -- our use case is that we'd rather go through with an import even if the data doesn't match up perfectly with the schema, as we have downstream things that let our users clean up their data. We'd rather get an imperfect import than the import not going through at all! We also like fast imports and avoiding validation gets us out of a lot of CPU cycles :)

PR #2214, opened 2024-02-03
fix(taps): update _flatten_record function by adding  allow_nan=True in json.dumps

### LINKED ISSUE #2213: bug: ValueError: Out of range float values are not JSON compliant
### Singer SDK Version

0.34.1

### Is this a regression?

- [ ] Yes

### Python Version

3.9

### Bug scope

Taps (catalog, state, etc.)

### Operating System

MacOS

### Description

I get `ValueError: Out of range float values are not JSON compliant`  when trying to sync Amazon Document DB collection. 

The probably reason is that `simple_json.dumps()` has `allow_nan=False` default, while standard `dumps()` function from standard `json` library has `allow_nan=True`. 

As there is no way to pass configuration from outside could we add this functionality? I can introduce relevant fix. 

### Code

_No response_

dgawlowsky:
Fix #2213 

----
📚 Documentation preview 📚: https://meltano-sdk--2214.org.readthedocs.build/en/2214/


singer_sdk/helpers/_flattening.py
```diff
@@ -430,7 +430,7 @@ def _flatten_record(
             items.append(
                 (
                     new_key,
-                    json.dumps(v, use_decimal=True, default=str)
+                    json.dumps(v, use_decimal=True, default=str, allow_nan=True)
```
edgarrmondragon, line 433: Wdyt about making this configurable with two initial NaN options (e.g. `fail` and `allow`)?

dgawlowsky: I gave it a first try, please see move the conversation to #2222 and close this PR if possible.

dgawlowsky: @edgarrmondragon  can you take a look?

dgawlowsky: @edgarrmondragon  please close this PR and see the proposal in https://github.com/meltano/sdk/pull/2222

PR #2222, opened 2024-02-05
fix(taps): add argument for nan handling strategy to  _flatten_record

dgawlowsky:
----
📚 Documentation preview 📚: https://meltano-sdk--2222.org.readthedocs.build/en/2222/


dgawlowsky:
Fix https://github.com/meltano/sdk/issues/2213

singer_sdk/helpers/_flattening.py
```diff
@@ -20,6 +20,7 @@ class FlatteningOptions(t.NamedTuple):
     max_level: int
     flattening_enabled: bool = True
     separator: str = DEFAULT_FLATTENING_SEPARATOR
+    nan_strategy: t.Literal["fail", "allow"] = "fail"
```
edgarrmondragon, line 23: We may want to support msgspec for serialization in the near future (#1784), and it defaults to making the values null. Do you think there's a way to make these options future-compatible with msgspec?

https://jcristharif.com/msgspec/supported-types.html#float

dgawlowsky: Can you specify a little more details? I am not very familiar with `msgspec` lib. In simplejson which is currently used there is `ignore_nan` argument which could be used. I am thinking about adding one more nan_strategy "convert_null" which will set `ignore_nan` to `True`. WDYT?

dgawlowsky: @edgarrmondragon 
Added this! sorry for the delay!

dgawlowsky: Hi @edgarrmondragon! Can you share the status of the review? We are using meltano in one of our projects and will be deploying do production soon, hence it will be easier to pull library from official distribution. Thanks in advance!

larsrinn:
What's the state of this? I'm trying to develop a tap where there None-values are expected. Do you see a way around this PR to get it done quickly?

edgarrmondragon:
> What's the state of this? I'm trying to develop a tap where there None-values are expected. Do you see a way around this PR to get it done quickly?

@larsrinn The biggest blocker in this PR is a missing public setting to toggle this behavior, similar to other built-in ones like `stream_maps` and `add_record_metadata`.

PR #2232, opened 2024-02-10
fix(templates): Test workflow job now fails for unsupported Python versions in cookiecutter templates

### LINKED ISSUE #2223: bug: Test workflow does not fail for unsupported Python versions
### Singer SDK Version

0.35.0

### Is this a regression?

- [ ] Yes

### Python Version

NA

### Bug scope

Other

### Operating System

GitHub Actions

### Description

When an unsupported Python version is specified in the test workflow matrix, Poetry tries to find a different compatible Python version to install, and will use that instead if it one is available. This produces false-positives for test runs - for example, I expect the Python 3.7 test to fail here: https://github.com/ReubenFrankel/tap-f1/actions/runs/7801705432/job/21277456475#step:5:13

### Code

```Python
...
The currently activated Python version 3.7.17 is not supported by the project (>=3.8,<4).
Trying to find and use a compatible version. 
Using python3.10 (3.10.12)
Creating virtualenv tap-f1-gIbUqA1n-py3.10 in /home/runner/.cache/pypoetry/virtualenvs
Installing dependencies from lock file
...
```

ReubenFrankel:
Closes #2223

----
📚 Documentation preview 📚: https://meltano-sdk--2232.org.readthedocs.build/en/2232/

### DIFF
--- a/cookiecutter/mapper-template/{{cookiecutter.mapper_id}}/.github/workflows/test.yml
+++ b/cookiecutter/mapper-template/{{cookiecutter.mapper_id}}/.github/workflows/test.yml
@@ -8,6 +8,7 @@ on: [push]
 jobs:
   pytest:
     runs-on: ubuntu-latest
+    continue-on-error: true
     env:
       GITHUB_TOKEN: {{ '${{secrets.GITHUB_TOKEN}}' }}
     strategy:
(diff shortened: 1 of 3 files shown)


edgarrmondragon (approved):
Thanks @ReubenFrankel!

PR #2238, opened 2024-02-13
fix: Remove upper constraint on `urllib3`

edgarrmondragon:
----
📚 Documentation preview 📚: https://meltano-sdk--2238.org.readthedocs.build/en/2238/


edgarrmondragon:
Anytime someone runs `poetry lock` it'll start causing problems, so I'm not sure if this is worth it...

Issue #2227, opened 2024-02-07
Diference between handling activate_version messages between hard_delete true and false


vitoravancini:
Hello everyone,

It seems that with hard_delete = true when we receive an activate_version message we delete everything equal or lesser than the new version:
https://github.com/meltano/sdk/blob/7c77366e30efff5bc04cce709ecac359fdfabc2f/singer_sdk/sinks/sql.py#L407

and when we set hard_delete = false we soft delete everything less than the new version:  
https://github.com/meltano/sdk/blob/7c77366e30efff5bc04cce709ecac359fdfabc2f/singer_sdk/connectors/sql.py#L1214

Is this how it supposed to be? 
When using target_snowflake(based on singer_sdk) with tap mssql we drop everything but the last batch at the end of the execution when hard_delete = true. 
The tap emits another activate_message at the end with the same version, which triggers a sql like 

```
delete from {table} where _sdc_record_version <= {version}
```

that {version} beeing the same version of all the records inserted up to that point.

Should we make both logics(hard_delete true and false) the same?

thank you!

edgarrmondragon:
Should we make both logics(hard_delete true and false) the same?

@vitoravancini I think we should :)

PRs welcome!

vitoravancini:
Nice, will try to send it later today

PR #2242, opened 2024-02-14
fix(taps): Hard and soft deletes for handling `ACTIVATE_VERSION` messages now use the same `WHERE` clause 

vitoravancini:
Soft delete is deleting only versions smaller than the latest version .
Hard delete is deleting smaller and equal.
Making both strategies delete only smaller versions.

fixes https://github.com/meltano/sdk/issues/2227

----
📚 Documentation preview 📚: https://meltano-sdk--2242.org.readthedocs.build/en/2242/


edgarrmondragon (approved):
Thanks @vitoravancini!

PR #2243, opened 2024-02-15
fix: Force flattened record according to provided flattened schema

joaopamaral:
If a flattened schema is provided, and there is a dynamic object field, the flattened record can ignore the provided schema. This means that the resulting output may have more fields than what was originally specified in the provided schema. This can cause problems if the loader does not accept fields that are not listed in the schema.

So this PR is checking if a flattened is provided and forcing the flattened record to use it during the flatten record call.

Example:
```
RECORD:

{
    'key_1': 1,
    'key_2': {
        'key_3': 'value',
        'key_4': {
            'key_5': 1,
            'key_6': ['a', 'b']
        }
    }
}

SCHEMA:
{
    'key_1': {'type': ['null', 'integer']},
    'key_2__key_3': {'type': ['null', 'string']},
    'key_2__key_4': {'type': ['null', 'object']},
}

WRONG RESULT:
{
    'key_1': 1,
    'key_2__key_3': 'value',
    'key_2__key_4__key_5': 1,
    'key_2__key_4__key_6': '["a", "b"]'
}

EXPECTED RESULT:
{
    'key_1': 1, 
    'key_2__key_3': 'value', 
    'key_2__key_4': '{"key_5": 1, "key_6": ["a", "b"]}'
}
```

----
📚 Documentation preview 📚: https://meltano-sdk--2243.org.readthedocs.build/en/2243/


joaopamaral (commented):
added new test

singer_sdk/helpers/_flattening.py
```diff
@@ -376,6 +378,11 @@ def flatten_record(
     Returns:
         A flattened version of the record.
     """
+    if flattened_schema is None and max_level is None:
```
edgarrmondragon, line 381: I wonder the type annotation is accurate... In which cases would `flattened_schema` be `None`?

The only place where this function is called is https://github.com/meltano/sdk/blob/5734e694994562830a5f7d4894910f2a41ec5aac/singer_sdk/mapper.py#L119-L124

and there both `flattened_schema` and `max_level` are ensured to be non-null AFAICT.

edgarrmondragon (changes requested):
Thanks for contributing @joaopamaral! I left a few comments.

tests/core/test_flattening.py
```diff
@@ -0,0 +1,76 @@
...
+    }
+    if expected_exception:
+        with pytest.raises(expected_exception):
+            flatten_record(
+                record, max_level=max_level, flattened_schema=flattened_schema
+            )
+    else:
+        result = flatten_record(
+            record, max_level=max_level, flattened_schema=flattened_schema
+        )
```
edgarrmondragon, line 75: There's a pattern we've used elsewhere in the test suite based on `contextlib.nullcontext`[^1] that could probably be used here

[^1]: https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext

singer_sdk/helpers/_flattening.py
joaopamaral: > I wonder the type annotation is accurate... In which cases would flattened_schema be None?

Yes, makes sense. Let me remove this check and the None option for `flattened_schema`.

joaopamaral: Maybe it's better to create a flag to enforce the schema on the output or the `max_level` here will be ignored. Wdyt?

tests/core/test_flattening.py
joaopamaral: TIL, thanks for the tips @edgarrmondragon! I've removed the exception since `flattened_schema` and `max_level` are always non null.

tests/core/test_flattening.py
```diff
@@ -0,0 +1,74 @@
...
+                }
+            },
+            1,
+            {
+                "key_1": 1,
+                "key_2__key_3": "value",
+                "key_2__key_4": '{"key_5": 1, "key_6": ["a", "b"]}',
+            },
+            id="max level limiting flattened schema",
+        ),
```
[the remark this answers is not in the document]
edgarrmondragon: That 3rd case should not happen IIUC. It would require that the schema and record are processed with different max flattening level, so I think it's ok to leave out a warning.

edgarrmondragon (approved):
Thanks @joaopamaral!

PR #2248, opened 2024-02-15
feat(targets): Targets now accept a `batch_size_rows` setting to configure how many rows are loaded in each record batch

### LINKED ISSUE #2155: feat: allow target `MAX_SIZE_DEFAULT` to be externally configurable
In the base sink class we define a `MAX_SIZE_DEFAULT` thats set to 10,000 https://github.com/meltano/sdk/blob/e9f2c179fb3a8bcd103546209b4d1c4b8df78ec6/singer_sdk/sinks/core.py#L133. Thats a fine base default and target developers can override it if their target prefers smaller or larger batches.

The thing thats not supported today is the use case where I want to configure a target, say target-postgres, to output more frequently because my tap records are slow/expensive to get so the possibility of waiting until 10k records are retrieved, the batch is old enough to drain, or the sync completes, is undesirable. I might want to drain after every 50 records or so in some case.

One example that came up a while ago is from the map-gpt-embeddings plugin that requests from openai's api. It takes a while to embed all input data and each request costs money so preferably I'd drain as frequently as possible.

cc @edgarrmondragon

BuzzCutNorman:
This feature will allow target users the ability to set the batch size used to triggers a drain via the target's meltano confiig.

 ` batch_size_rows: 100000`

When a batch is full and a drain is triggered, the message logged will now state the size of the batch so the users has a visual confirmation that the batch size they configured is being utilized.

The code for this PR was pulled out of PR 1876.

Closes #2155

----
📚 Documentation preview 📚: https://meltano-sdk--2248.org.readthedocs.build/en/2248/


BuzzCutNorman:
@pnadolny13 and @edgarrmondragon I broke out the code from pr 1876 that gives the user of a target the ability to set the batch size in rows which triggers a drain.  Please take a look and let me know what you think.

singer_sdk/sinks/core.py
```diff
@@ -269,13 +265,35 @@ def current_size(self) -> int:
...
+            A batch_size_rows object.
+        """
+        return self._batch_size_rows
+
+    @property
+    def max_size(self) -> int:
+        """Get max batch size.
+
+        Returns:
+            Max number of records to batch before `is_full=True`
```
edgarrmondragon, line 289:
```suggestion
        Returns:
            Max number of records to batch before `is_full=True`

        .. versionremoved:: 0.36.0
        This property now takes into account the ``batch_size_rows`` target setting.
```

singer_sdk/sinks/core.py
```diff
@@ -269,13 +265,38 @@ def current_size(self) -> int:
...
+
+        .. versionremoved:: 0.36.0
```
edgarrmondragon, line 291:
```suggestion
        .. versionchanged:: 0.36.0
```

singer_sdk/sinks/core.py
```diff
@@ -269,13 +265,38 @@ def current_size(self) -> int:
...
+
+    @property
+    def max_size(self) -> int:
+        """Get max batch size.
+
+        Returns:
+            Max number of records to batch before `is_full=True`
+
+        .. versionchanged:: 0.36.0
+        This property now takes into account the ``batch_size_rows`` target setting.
```
edgarrmondragon, line 292:
```suggestion
        .. versionchanged:: 0.36.0
           This property now takes into account the ``batch_size_rows`` target setting.
```

edgarrmondragon (approved):
Thanks for taking the time for extracting this into its own PR @BuzzCutNorman!

BuzzCutNorman:
@edgarrmondragon Thanks 🙏 for all of the comment and documentation edits.  Load better 🥳 than what I had.  🎉

PR #2274, opened 2024-02-29
fix(templates): Test workflow now fails for any Python matrix job failure in cookiecutter template

ReubenFrankel:
Follow-up from #2232 

---

The current test workflow configuration for the `pytest` job uses `continue-on-error: true`, meaning that when any matrix job (i.e. test run for a specific Python version) fails, the workflow itself still shows as passing - for example: https://github.com/Matatika/tap-spotify/actions/runs/8089702498

[`fail-fast: false`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) is what should have been implemented initially (sorry, still fairly new to GitHub Actions) - for example: https://github.com/Matatika/tap-spotify/actions/runs/8089862693

----
📚 Documentation preview 📚: https://meltano-sdk--2274.org.readthedocs.build/en/2274/

### DIFF
--- a/cookiecutter/mapper-template/{{cookiecutter.mapper_id}}/.github/workflows/test.yml
+++ b/cookiecutter/mapper-template/{{cookiecutter.mapper_id}}/.github/workflows/test.yml
@@ -8,10 +8,10 @@ on: [push]
 jobs:
   pytest:
     runs-on: ubuntu-latest
-    continue-on-error: true
     env:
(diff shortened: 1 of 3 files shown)


edgarrmondragon (approved):
Ah, thanks for circling back!

PS: I should've caught this since we use it here and in Meltano 🤦‍♂️

PR #2288, opened 2024-03-07
fix(targets): Bumped min joblib dependency to `>=1.3.0` and replaced deprecated `parallel_backend` with `parallel_config`

BuzzCutNorman:
JobLlib which is used in `Target._drain_all` to accomplish threaded drains of all the sinks utilizes `parallel_backend` which  has been deprecated.  The suggested replaement is `parallel_config`.  

https://joblib.readthedocs.io/en/stable/index.html#deprecated-functionalities

----
📚 Documentation preview 📚: https://meltano-sdk--2288.org.readthedocs.build/en/2288/


edgarrmondragon:
I think this requires us to bump the min joblib constraint to `>=1.3.0`

https://joblib.readthedocs.io/en/stable/developing.html#release-1-3-0-2023-06-28

BuzzCutNorman:
> I think this requires us to bump the min joblib constraint to `>=1.3.0`
> 
> https://joblib.readthedocs.io/en/stable/developing.html#release-1-3-0-2023-06-28

Great catch 🥳 I missed that it isn't present until 1.3.0

poetry.lock
edgarrmondragon, line 1: Some wheels were lost here. Can you run

```
poetry cache clear pypi --all -n
poetry lock --no-update
```

?

BuzzCutNorman: those commands have been run.

edgarrmondragon:
Thanks!

Issue #2182, opened 2024-01-25
bug: Cookiecutter `tap-template` errors out if VSCode IDE is not selected


menzenski:
### Singer SDK Version

0.34.1

### Is this a regression?

- [X] Yes

### Python Version

3.11

### Bug scope

Cookiecutter templates

### Operating System

macOS Ventura 13.6

### Description

I went through the cookiecutter tap-template generation process and it errored out without creating a project directory. I believe this is due to this line? https://github.com/meltano/sdk/blob/main/cookiecutter/tap-template/hooks/post_gen_project.py#L31

It's also not clear what `__prompts__ [default]:` is doing - when I hit enter with the intent of selecting "default option", it errored, but I was able to provide a JSON object `{}` to move past that.

### Code

```Python
cookiecutter https://github.com/meltano/sdk --directory="cookiecutter/tap-template"
You've downloaded /Users/matt/.cookiecutters/sdk before. Is it okay to delete and re-download it? [yes]: yes
source_name [MySourceName]: MySource
admin_name [FirstName LastName]: Matt Menzenski
admin_email [firstname.lastname@example.com]: matt@example.com
tap_id [tap-my-source]: tap-my-source
library_name [tap_my_source]:
variant [None (Skip)]:
Select stream_type:
1 - REST
2 - GraphQL
3 - SQL
4 - Other
Choose from 1, 2, 3, 4 [1]: 1
Select auth_method:
1 - API Key
2 - Bearer Token
3 - Basic Auth
4 - OAuth2
5 - JWT
6 - Custom or N/A
Choose from 1, 2, 3, 4, 5, 6 [1]: 2
Select include_ci_files:
1 - GitHub
2 - None
Choose from 1, 2 [1]: 1
Select license:
1 - Apache-2.0
2 - None
Choose from 1, 2 [1]: 1
Select ide:
1 - VSCode
2 - None
Choose from 1, 2 [1]: 2
__prompts__ [default]:
Error: Unable to decode to JSON.
__prompts__ [default]:
Error: Unable to decode to JSON.
__prompts__ [default]: {}
Traceback (most recent call last):
  File "/var/folders/ms/gk5mg99d1d3_6sn7bqyhzx0m0000gp/T/tmptczel5gz.py", line 31, in <module>
    shutil.rmtree(".vscode")
  File "/usr/local/Cellar/python@3.9/3.9.18/Frameworks/Python.framework/Versions/3.9/lib/python3.9/shutil.py", line 724, in rmtree
[...]

PR #2342, opened 2024-03-25
fix(templates): Avoid failing if VSCode IDE config is not requested for target and mapper cookiecutter templates

ReubenFrankel:
Related to #2182 and #2183

These templates currently do not have a `.vscode` directory, and so when VSCode IDE config is not requested (i.e. `2 - None` option) the post generate project hook fails trying to remove a directory that doesn't exist:

```
FileNotFoundError: [Errno 2] No such file or directory: '.vscode'
ERROR: Stopping generation because post_gen_project hook script didn't exit successfully
```

This PR simply ignores that error for target and mapper templates, should it occur. I haven't added any VSCode config (e.g. `launch.json`) as it's probably not gonna be the same as the tap template (in hindsight, I might have been a bit quick to exclude `launch.json` from the `check-json` pre-commit hook).

### DIFF
--- a/cookiecutter/mapper-template/hooks/post_gen_project.py
+++ b/cookiecutter/mapper-template/hooks/post_gen_project.py
@@ -14,4 +14,4 @@ if __name__ == "__main__":
         shutil.rmtree(Path(".github"))
 
     if "{{ cookiecutter.ide }}" != "VSCode":
-        shutil.rmtree(".vscode")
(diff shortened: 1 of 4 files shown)


ReubenFrankel:
From https://meltano.slack.com/archives/C06A1M1507J/p1711368228779059?thread_ts=1711084967.121779&cid=C06A1M1507J

edgarrmondragon (approved):
Thanks @ReubenFrankel!

PR #2348, opened 2024-03-27
fix(targets): Added a condition to the `No schema for record field` warning

tobiascadee:
If schema flattening is enabled but a value for a key in the record which is flattened is None, this warning is returned because the record can not be flattened and then the key is not in the schema. This can be avoided by checking if the value is None. If it is None the record field actually does not exist so we do not need a warning.

----
📚 Documentation preview 📚: https://meltano-sdk--2348.org.readthedocs.build/en/2348/


edgarrmondragon (approved):
Thanks @tobiascadee!

Issue #2353, opened 2024-03-30
feat: prepare_table() makes O(1) calls to prepare_column(), instead of O(n) (n number of columns)


raulbonet:
### Feature scope

Taps (catalog, state, stream maps, tests, etc.)

### Description

Function `prepare_table()` in `SQLConnector` calls `prepare_column()`, for each column in a table. 
In turn, this function queries the database to first determine whether the column exists or not. Therefore we make O(n) calls to the database. 
However, in a single call we can retrieve all columns and leverage this information. 

The `target-postgres` already implemented this optimisation by overriding the  default methods `prepare_table()` and `prepare_columns()`.

The way `target-postgres` implemented it implied modifying the signature of the functions, my implementation is different. 
I take the responsibility of checking whether the column exists out of the `prepare_column()` function and encapsulate this in a new function called `prepare_table_columns()`. 

However, a thought: I think that in the end the underlying issue is that functions like `prepare_table()` and this new function `prepare_table_columns()` belong to the `SQLSink` class rather than the connector: a new sink is created every time there is a schema change. The class that knows the current schema and therefore can do these optimisations is the sink, not the connector maybe?

PR #2354, opened 2024-03-30
feat(targets): Added a new `prepare_table_columns` method to the `SQLConnector` base class

raulbonet:
ref #2353 

----
📚 Documentation preview 📚: https://meltano-sdk--2354.org.readthedocs.build/en/2354/


singer_sdk/connectors/sql.py
```diff
@@ -798,12 +798,40 @@ def prepare_table(
...
+        )
+
+    def prepare_table_columns(
+        self,
+        full_table_name: str,
+        schema: dict,
+    ) -> None:
+        """Adapt target table columns to provided schema if possible. The advantage of encapsulating this
+        in a separate method is that we reduce the number of calls to the database by calling get_table_columns()
+        only once per table.
```
edgarrmondragon, line 813:
```suggestion
        """Adapt target table columns to provided schema if possible.
        
        The advantage of encapsulating this in a separate method is that we reduce
        the number of calls to the database by calling get_table_columns() only once
        per table.
```

singer_sdk/connectors/sql.py
```diff
@@ -798,12 +798,42 @@ def prepare_table(
...
+
+        Args:
+            full_table_name: the target table name.
+            schema: the JSON Schema for the table.
+        """
+        columns = self.get_table_columns(
+            full_table_name=full_table_name,
+        )
         for property_name, property_def in schema["properties"].items():
-            self.prepare_column(
```
edgarrmondragon, line 802: Should we add a deprecation warning when this function is called?

singer_sdk/connectors/sql.py
```diff
@@ -798,12 +798,42 @@ def prepare_table(
             )
             return
 
+        self.prepare_table_columns(
+            full_table_name=full_table_name,
+            schema=schema,
+        )
+
+    def prepare_table_columns(
```
edgarrmondragon, line 806: I think we should tests for this. `tests/core/test_connector_sql.py` might a good place for that.

PR #2525, opened 2024-07-09
refactor!: Make `PyJWT` and `cryptography` dependencies optional

edgarrmondragon:
* A quick gathering of some data points from MeltanoLabs:

  - 1 package relies on `jwt`:
    - [tap-github](https://github.com/MeltanoLabs/tap-github/blob/1b41def4001c2111a8e7024aa94241759e9ba90e/tap_github/authenticator.py#L10), but it's already [an explicit dependency](https://github.com/MeltanoLabs/tap-github/blob/1b41def4001c2111a8e7024aa94241759e9ba90e/pyproject.toml#L27)
  - 1 package relies on `cryptography`:
    - [target-snowflake](https://github.com/MeltanoLabs/target-snowflake/blob/b357c52574380345e9253eace99989cd35787ca0/target_snowflake/connector.py#L8), but it's already [an explicit dependency](https://github.com/MeltanoLabs/target-snowflake/blob/b357c52574380345e9253eace99989cd35787ca0/pyproject.toml#L18)

* Matatika has 0 repos
* Automattic has 0 repos
* AutoIDM has 0 repos
* Hotglue on GitHub has 1 package that imports the transitive `jwt` dependency:
  - https://github.com/hotgluexyz/tap-stella-connect/pull/11
* Hotglue on GitLab is a harder nut to crack 😅

It's possible to get more data by going through the Hub, and downloading the source code of extractors and loaders.

----
📚 Documentation preview 📚: https://meltano-sdk--2525.org.readthedocs.build/en/2525/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ python = ">=3.8"
 backoff = { version = ">=2.0.0", python = "<4" }
 backports-datetime-fromisoformat = { version = ">=2.0.1", python = "<3.11" }
 click = "~=8.0"
-cryptography = ">=3.4.6"
 fs = ">=2.4.16"
 importlib-metadata = {version = "<9.0.0", python = "<3.12"}
 importlib-resources = {version = ">=5.12.0,!=6.2.0,!=6.3.0,!=6.3.1", python = "<3.9"}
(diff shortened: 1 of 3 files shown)


pyproject.toml
WillDaSilva, line 1: These dependencies are used in `singer_sdk/authenticators.py` within a function. Would it make sense to import them within a `try` block, and if there's an import error, provide a clear error message about needing to install the `jwt` extra to use this functionality?

edgarrmondragon: Makes sense: c396b6e99184c346b4ed96dd1bc18b5e3e3487c7

WillDaSilva (approved):

edgarrmondragon:
Given how uncommon jwt-based authentication seems to be in downstream packages, I feel confident merging this and calling it out in the release notes for the next version.

PR #2578, opened 2024-07-30
feat(mappers): Stream map expressions now have access to the `Faker` class, rather than just a faker instance

Mac-lp3:
I made [a post in the discussions](https://github.com/meltano/sdk/discussions/2537) about my motivation for this.

Note the Faker lib provides two ways to accomplish this:
1. How it is done in this PR

```python
from faker import Faker
Faker.seed('something')
faker = Faker()
```

The meltano mapper config would then look something like this:
```yaml
stream_maps:
    "*":
        FirstName: Faker.seed('something') or fake.first_name() # note lib not instance
    faker_config:
        seed: 'something' # kind of redundant in this example, but hey, still works
```

2. Create your faker instance using the `Faker.factory`. 
```python
from faker.factory import Factory
Faker = Factory.create
fake = Faker()
fake.seed(0)
```

With this approach, the meltano mapper config would have looked something like this:
```yaml
stream_maps:
    "*":
        FirstName: faker.seed('something') or fake.first_name() # note instance not lib
    faker_config:
        seed: 'something'
```

Option 2 is actually [what the faker docs recommend](https://faker.readthedocs.io/en/master/fakerclass.html).

Unfortunately, option 2 does not allow for multiple locales. Given that, I thought it best to go with option 1, rather than breaking other people's code.

----
📚 Documentation preview 📚: https://meltano-sdk--2578.org.readthedocs.build/en/2578/


edgarrmondragon:
Hi @Mac-lp3, thanks for the PR!

I've read through your post in https://github.com/meltano/sdk/discussions/2537 also the docs in and https://faker.readthedocs.io/en/master/fakerclass.html#upgrade-guide and I think this makes and is the easiest way to add support for reproducible masking.

I think it's worth documenting the new object that'll be available in the stream map context. Wdyt?

Mac-lp3:
Hi again @edgarrmondragon,

Docs are now updated. 

I added `Faker` the list of built-in variables. I also added a section about how to use faker and the different approaches. 

Hope this is up to standard, but just lmk if not 🤝

docs/stream_maps.md
```diff
@@ -249,6 +249,8 @@ can be referenced directly by mapping expressions.
 - `fake` - a [`Faker`](inv:faker:std:doc#index) instance, configurable via `faker_config`
   (see previous example) - see the built-in [standard providers](inv:faker:std:doc#providers)
   for available methods
+- `Faker` - the `Faker` library itself. This was made available to enable consistent data
```
edgarrmondragon, line 252:
```suggestion
- `Faker` - the [`Faker`](nv:faker:std:doc#fakerclass) class. This was made available to enable consistent data
```

docs/stream_maps.md
```diff
@@ -249,6 +249,8 @@ can be referenced directly by mapping expressions.
 - `fake` - a [`Faker`](inv:faker:std:doc#index) instance, configurable via `faker_config`
   (see previous example) - see the built-in [standard providers](inv:faker:std:doc#providers)
   for available methods
+- `Faker` - the [`Faker`](nv:faker:std:doc#fakerclass) class. This was made available to enable consistent data
```
edgarrmondragon, line 252:
```suggestion
- `Faker` - the [`Faker`](inv:faker:std:doc#fakerclass) class. This was made available to enable consistent data
```

edgarrmondragon (approved):

edgarrmondragon:
Superseded by #2598

PR #2582, opened 2024-08-01
fix(target): DDL for adding a column now uses a SQLAlchemy-compiled clause to apply proper quoting

haleemur:
# Description

call the `compile` method, specifying the sqlalchemy engine, this ensures that sqlalchemy is able to specify database specific object quoting.

----
📚 Documentation preview 📚: https://meltano-sdk--2582.org.readthedocs.build/en/2582/


edgarrmondragon (approved):
Thanks @haleemur!

PR #2583, opened 2024-08-02
fix(targets): Quote add-column-ddl with column starting with `_` (underscore) based on engine

haleemur:
# Description

Pass the compiled string output to the DDL element constructor.

This ensures that the compilation is engine specific and engine specific quoting rules are respected.

Passing the compiler element to the context `sa.DDL` does not yield proper engine-specific quoting.

Some databases (e.g. IBM Db2, and maybe others as well) require fields starting with underscore (e.g. `_id`) to be quoted, while other databases (postgres, sqlite) accept fields with leading underscores without quotes.

## Testing

Updated schema evolution test in `testing/target_test_streams` to ensure any downstream target is also tested for scenario: `add column starting with underscore`.

Added scenario: `add column starting with underscore` to the sqlite sample target test.

----
📚 Documentation preview 📚: https://meltano-sdk--2583.org.readthedocs.build/en/2583/


edgarrmondragon (approved):

edgarrmondragon:
Thanks @haleemur!

PR #2589, opened 2024-08-05
fix(targets): Make sink assertion compatible with stream maps

JCotton1123:
The change here is necessary to ensure the assertion utilizes the aliases stream name (vs the original stream name).

Given a stream_maps config with an `__alias__` configuration, see example below, a target will produce the following error:

```
singer_sdk.exceptions.RecordsWithoutSchemaException: A record for stream 'Account' was encountered before a corresponding schema. Check that the Tap correctly implements the Singer spec.
```

```
# meltano.yaml
...
  loaders:
  - config:
      athena_database: db
      s3_path: s3://bucket/data/
      stream_maps:
        Account:
          __alias__: Account_v2
    name: target-s3-parquet
```

----
📚 Documentation preview 📚: https://meltano-sdk--2589.org.readthedocs.build/en/2589/


edgarrmondragon:
@JCotton1123 thanks for the PR!

Can you expand on what problem this solves? The original code does smell like a bug, but I'd like to understand better what errors you're running into with the current implementation.

JCotton1123:
> @JCotton1123 thanks for the PR!
> 
> Can you expand on what problem this solves? The original code does smell like a bug, but I'd like to understand better what errors you're running into with the current implementation.

I just added a bit more detail. Let me know if I can provide more.

edgarrmondragon (approved):
Thanks @JCotton1123!

PR #2598, opened 2024-08-09
feat(mappers): Stream map expressions now have access to the `Faker` class, rather than just a faker instance

edgarrmondragon:
Copy of https://github.com/meltano/sdk/pull/2578 where I can push, since that one was created from a protected branch.

----
📚 Documentation preview 📚: https://meltano-sdk--2598.org.readthedocs.build/en/2598/


ReubenFrankel:
Maybe I'm missing something, but could this not have been achieved already using `seed_instance`? 

```yml
stream_maps:
  customers:
    # will always generate the same value for the same seed
    first_name: fake.seed_instance(_['first_name']) or fake.first_name()
faker_config:
  # IMPORTANT: `fake` and `Faker` names are only available if faker_config is defined.
  locale: en_US
```

https://faker.readthedocs.io/en/master/index.html#seeding-the-generator

edgarrmondragon:
Oh, right

<blockquote>

Each generator can also be switched to use its own instance of `random.Random`, separated from the shared one, by using the `seed_instance()` method, which acts the same way.

</blockquote>

If that's the case, we should deprecate `Faker` and recommend `seed_instance()`.

Thoughts @Mac-lp3?

Mac-lp3:
Happy to do some testing tomorrow and see if it provides the same
functionality. I'll respond here with the results and some pros/cons if any.

Mac-lp3:
@ReubenFrankel you are 100% correct, you can accomplish the exact same thing with `fake.seed_instance()`, so adding `Fake` was not necessary 🤦 

@edgarrmondragon, so it looks like the mapper.py changes can be reverted... i think it would at least be handy to keep the updated docs that explains how to get consistent hashing and why you may want to. Of course, it just needs to be updated again to use `fake.seed_instance()` rather than `Fake.seed()`, which I don't mind fixing since its my mistake...

Should I just submit a new PR for that? If so, should it just be for the documentation changes, or should it also include reverting `mapper.py`?

edgarrmondragon:
> @ReubenFrankel you are 100% correct, you can accomplish the exact same thing with `fake.seed_instance()`, so adding `Fake` was not necessary 🤦 
> 
> @edgarrmondragon, so it looks like the mapper.py changes can be reverted... i think it would at least be handy to keep the updated docs that explains how to get consistent hashing and why you may want to. Of course, it just needs to be updated again to use `fake.seed_instance()` rather than `Fake.seed()`, which I don't mind fixing since its my mistake...
> 
> Should I just submit a new PR for that? If so, should it just be for the documentation changes, or should it also include reverting `mapper.py`?

Thanks for the offer @Mac-lp3! Can you take a look at https://github.com/meltano/sdk/pull/2670 and comment if you think something's off or missing?

Mac-lp3:
@edgarrmondragon took a look and looks good to me!

ReubenFrankel:
@Mac-lp3 FWIW I think the docs change was super helpful! 😅

edgarrmondragon:
Thanks both!

Issue #390, opened 2022-05-20
Cast environment variables config values into their respective JSON schema type


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/393

Originally created by @edgarrmondragon on 2022-05-20 18:16:14

---

## Summary
[//]: # (Concisely summarize the feature you are proposing.)

Currently only arrays of strings are parsed: `value1,value2` -> `["value1", "value2"]`. A similar process for `number`, `boolean` and `integer` would be valuable.

## Proposed benefits
[//]: # (Concisely summarize the benefits this feature would bring to yourself and other users.)

- Users can fully rely on environment variables and `.env` to configure their taps
- Richer data types can be used for environment variables

## Proposal details
[//]: # (In as much detail as you are able, describe the feature you'd like to build or would like to see built.)

We already have some functions to check for data types in https://gitlab.com/meltano/sdk/-/blob/main/singer_sdk/helpers/_typing.py, so we could extend and leverage those when parsing environment variables.

## Best reasons not to build
[//]: # (Will this negatively affect any existing functionality? Do you anticipate any breaking changes versus what may already be working today? Make the counter-argument to your proposal here.)

Support for environment variables is not common in the Singer ecosystem.

MeltyBot:
[View 1 previous comment from the original issue on GitLab](https://gitlab.com/meltano/sdk/-/issues/393)

PR #2637, opened 2024-09-02
feat: Environment variables are now parsed for boolean, integer, array and object setting values

### LINKED ISSUE #1896: break: Support the same env var formats as Meltano
Currently, a setting of array type can be parsed from an env var with the following format:

```python
MY_ENV_VAR='a,b,c'
```

However, Meltano expects array settings to be JSON arrays:

```python
MY_ENV_VAR='["a","b","c"]'
```

This mismatch is surprising to users familiar with Meltano, and to developers using Meltano to test connectors during development. In the latter case, the env vars and `.env` file are incompatible.

Related:

* https://github.com/meltano/sdk/issues/390
* This mismatch was introduced early in SDK development: https://github.com/meltano/sdk/commit/e66de5d3b90f93288d1e84206eaf91a3116e2823

edgarrmondragon:
Related:

- Closes #1896 (but without a breaking change at this point)
- Closes https://github.com/meltano/sdk/issues/390

----
📚 Documentation preview 📚: https://meltano-sdk--2637.org.readthedocs.build/en/2637/


edgarrmondragon:
@dependabot rebase

PR #2665, opened 2024-09-11
fix: Mapped datetime values are now typed as `date-time` strings in the schema message

### LINKED ISSUE #1695: bug: Mapped datetime values are typed as strings in the schema message
### Singer SDK Version

0.26.0

### Is this a regression?

- [ ] Yes

### Python Version

NA

### Bug scope

Mapping (stream maps, flattening, etc.)

### Operating System

NA

### Description

When generating a datetime value from a string using stream map functionality, e.g.

```yaml
stream_maps:
  my_stream:
    created: datetime.datetime.from_timestamp(createdAt)
```

the detected schema type for the value falls back to string.

https://github.com/meltano/sdk/blob/33d64c8dc94d039368bee16da6ab0037e6672ff6/singer_sdk/mapper.py#L372-L379

### Code

_No response_

gregkoutsimp:
When generating a datetime value from a string using stream map functionality, the detected schema type for the value falls back to string.

In this implementation we extend _eval_type method of class CustomStreamMap from the mapper.py to be bale to process dates and datetimes data types in stream maps.

Closes #1695

----
📚 Documentation preview 📚: https://meltano-sdk--2665.org.readthedocs.build/en/2665/


edgarrmondragon (approved):
Thanks @gregkoutsimp!

PR #2755, opened 2024-11-11
fix: Safely compare UUID replication keys with state bookmarks

nikzavada:
This change will return a stringified value for an instance of uuid.UUID.

----
📚 Documentation preview 📚: https://meltano-sdk--2755.org.readthedocs.build/en/2755/


edgarrmondragon (approved):
Thanks @nikzavada!

PR #2785, opened 2024-11-30
fix: Use FS-specific `listdir` in folder tap

edgarrmondragon:
----
📚 Documentation preview 📚: https://meltano-sdk--2785.org.readthedocs.build/en/2785/

### DIFF
--- a/singer_sdk/testing/templates.py
+++ b/singer_sdk/testing/templates.py
@@ -103,7 +103,7 @@ class TestTemplate:
         Raises:
             ValueError: if Test instance does not have `name` and `type` properties.
         """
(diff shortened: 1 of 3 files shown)


singer_sdk/testing/templates.py
```diff
@@ -104,7 +104,7 @@ def run(
             ValueError: if Test instance does not have `name` and `type` properties.
         """
         if not self.name or not self.plugin_type:
```
edgarrmondragon, line 106:
```suggestion
         if not self.name or not self.plugin_type:  # pragma: no cover
```

PR #2793, opened 2024-12-02
perf(taps): Improved discovery performance for SQL taps

edgarrmondragon:
TODO:

- [x] Measure it 📈 
- [ ] Document it
- [x] Add a Codspeed benchmark? 🏎️

----
📚 Documentation preview 📚: https://meltano-sdk--2793.org.readthedocs.build/en/2793/

### DIFF
--- a/singer_sdk/connectors/sql.py
+++ b/singer_sdk/connectors/sql.py
@@ -878,15 +878,14 @@ class SQLConnector:  # noqa: PLR0904
     def discover_catalog_entry(
         self,
         engine: Engine,  # noqa: ARG002
-        inspected: Inspector,
+        inspected: Inspector,  # noqa: ARG002
         schema_name: str | None,
         table_name: str,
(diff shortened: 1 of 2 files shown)


singer_sdk/connectors/sql.py
```diff
@@ -969,21 +1039,30 @@ def discover_catalog_entries(self) -> list[dict]:
         result: list[dict] = []
         engine = self._engine
         inspected = sa.inspect(engine)
+        object_kinds = (
+            (reflection.ObjectKind.TABLE, False),
+            (reflection.ObjectKind.ANY_VIEW, True),
```
edgarrmondragon, line 1044: The regression in https://github.com/MeltanoLabs/tap-postgres/pull/528 was probably caused by ignoring views in the updated discovery process.

See https://github.com/MeltanoLabs/tap-postgres/issues/535#issuecomment-2513239541.

Issue #2818, opened 2024-12-27
feat: Add SHA256 encryption method


ben-schulz-mh:
### Feature scope

Inline mapping (stream maps, flattening, etc.)

### Description

Currently only MD5 is supported. SHA256 is a more secure encryption method that is used widely and is desired. For context, another ETL provider (Fivetran) use SHA256 and this would allow feature parity with that solution.

edgarrmondragon:
PRs welcome @ben-schulz-mh!

PR #2819, opened 2024-12-27
feat: add SHA256 encryption method to inline stream maps

ben-schulz-mh:
Implements https://github.com/meltano/sdk/issues/2818


edgarrmondragon (approved):
Thanks @ben-schulz-mh!

ben-schulz-mh:
Thanks for your generous help @edgarrmondragon

PR #2820, opened 2025-01-04
chore: Update templates

edgarrmondragon:
----
📚 Documentation preview 📚: https://meltano-sdk--2820.org.readthedocs.build/en/2820/


cookiecutter/tap-template/{{cookiecutter.tap_id}}/tox.ini
```diff
@@ -6,6 +6,8 @@ requires =
     tox>=4.19
 
 [testenv]
+pass_env =
+    TAP_*
```
edgarrmondragon, line 10: Maybe use a more specific wildcard

Issue #2783, opened 2024-11-29
JSON Schema vocabulary for Singer schemas


edgarrmondragon:
Stub.

---

For example OpenAPI defines its own vocabulary on top of Draft 2020-12

1. Custom data types and formats: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#data-types

2. Examples: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#schema-object-examples

3. https://spec.openapis.org/oas/3.1/dialect/base

```json
{
    "$id": "https://spec.openapis.org/oas/3.1/dialect/base",
    "$schema": "https://json-schema.org/draft/2020-12/schema",

    "title": "OpenAPI 3.1 Schema Object Dialect",
    "description": "A JSON Schema dialect describing schemas found in OpenAPI documents",

    "$vocabulary": {
        "https://json-schema.org/draft/2020-12/vocab/core": true,
        "https://json-schema.org/draft/2020-12/vocab/applicator": true,
        "https://json-schema.org/draft/2020-12/vocab/unevaluated": true,
        "https://json-schema.org/draft/2020-12/vocab/validation": true,
        "https://json-schema.org/draft/2020-12/vocab/meta-data": true,
        "https://json-schema.org/draft/2020-12/vocab/format-annotation": true,
        "https://json-schema.org/draft/2020-12/vocab/content": true,
        "https://spec.openapis.org/oas/3.1/vocab/base": false
    },

    "$dynamicAnchor": "meta",

    "allOf": [
        { "$ref": "https://json-schema.org/draft/2020-12/schema" },
        { "$ref": "https://spec.openapis.org/oas/3.1/meta/base" }
    ]
}
```

4. https://spec.openapis.org/oas/3.1/meta/base

```json

{
    "$id": "https://spec.openapis.org/oas/3.1/meta/base",
    "$schema": "https://json-schema.org/draft/2020-12/schema",

    "title": "OAS Base vocabulary",
    "description": "A JSON Schema Vocabulary used in the OpenAPI Schema Dialect",

    "$vocabulary": {
        "https://spec.openapis.org/oas/3.1/vocab/base": true
    },

    "$dynamicAnchor": "meta",

    "type": ["object", "boolean"],
    "properties": {
        "example": true,
        "discriminator": { "$ref": "#/$defs/discriminator" },
        "externalDocs": { "$ref": "#/$defs/external-docs" },
[...]

PR #1872, opened 2023-07-21
feat: Add sql-datatype to the SDK discovery and catalog

BuzzCutNorman:
This is an attempt to add the sql_datatype to the discovery process and will be reflected in the catalog.

Closes #1323

----
:books: Documentation preview :books:: https://meltano-sdk--1872.org.readthedocs.build/en/1872/

Issue #1323, opened 2023-01-13
[Feature]: Add `sql-datatype` to the SDK discovery and catalog. 


BuzzCutNorman:
### Feature scope

Taps (catalog, state, stream maps, etc.)

### Description

It would be very helpful if SDK based taps during the discovery process of a database grabbed column data types and placed them into the catalog metadata area for the column in the field of `sql-datatype`.  The `sql-datatype` field is a reserved keyword mentioned in the singer discovery documentation.

https://github.com/singer-io/getting-started/blob/master/docs/DISCOVERY_MODE.md#metadata

Keyword | Tap Type | Discoverable? | Description
-- | -- | -- | --
`sql-datatype` | database | discoverable | Represents the datatype of the database column

```
        {
          "breadcrumb": [
            "properties",
            "name"
          ],
          "metadata": {
            "sql-datatype": "varchar(255)",
            "selected-by-default": true
          }
        }
```

edgarrmondragon:
Related: #1903

edgarrmondragon:
Copying here my comment from https://github.com/meltano/sdk/pull/1872#issuecomment-2506661584:

One thing that unfortunately isn't very clear to me is how are targets supposed to consume this information since a target isn't aware of the tap's metadata. It might be better if this metadata lived in the schema, at least that way it would be emitted with `SCHEMA` messages and _could_ be used by the target.

The [official Singer docs](https://github.com/singer-io/getting-started/blob/master/docs/DISCOVERY_MODE.md#metadata) only explain the field as `Represents the datatype of a database column`, and pipelinewise's taps implement but the target don't seem to use it in any way:

https://github.com/search?q=org%3Atransferwise+sql-datatype+language%3APython&type=code&l=Python

Otherwise, who is the consumer of this metadata?

BuzzCutNorman:
I am guessing the consumer might be the Stich platform.  During this conversation I had in the Meltano Slack:

https://discuss.meltano.com/t/16288098/i-was-wondering-if-i-someone-could-point-me-in-the-proper-di

We came to the same conclusion that this would need to be a part of the `SCHEMA` message to become useful.  The alternative of using JSON Shema attributes to describe enough of the specifics to determine what the types are was mentioned.  This has been done in countless other taps and targets just there is not standardization of it.

knowing the database type (posgresql, sql server, oracle) and sql data type ( integer(), varchar(10) ) would be easier and maybe more accurate than the JSON Schema fingerprint method.

[...]

PR #2829, opened 2025-01-15
feat(targets): Support a `x-sql-datatype` JSON Schema annotation to let targets customize SQL type handling

### LINKED ISSUE #2102: feat(taps): Support declaring arbitrary SQLAlchemy type instances in SQL tap schemas
### Feature scope

Tap/target metadata.

### Description

Related to the original [Singer spec's `sql-datatype`](https://github.com/singer-io/getting-started/blob/master/docs/DISCOVERY_MODE.md#metadata) (see https://github.com/meltano/sdk/issues/1323, https://github.com/meltano/sdk/issues/1903), this new metadata object would work in an analogous manner to Python's logging module capability for importing and instantiating arbitrary callables and classes.

The story, specifically for vector/embedding fields for LLMs, would go something like:

1. As a user, I _know_ that one or more fields in a stream represent a vector. For example, `{"id": 1, "my_vector": [1, 2, 3]}`.

2. I would like to use [MeltanoLabs/target-postgres](https://github.com/MeltanoLabs/target-postgres) alongside [pgvector/pgvector-python](https://github.com/pgvector/pgvector-python?tab=readme-ov-file#sqlalchemy) to declare the SQLAlchemy type of `my_vector`: `pgvector.sqlalchemy.Vector`

3. I would like not to only declare the _type_ but also any arbitrary parameteres for it, like length, dimension, etc.:

   ```yaml
   # catalog metadata in Meltano syntax
   schema:
     my_stream:
       my_vector:
         sqlalchemy_type:
           (): "pgvector.sqlalchemy.Vector"
           dim: 3
   ```

   This would trigger the SDK to import `pgvector.sqlalchemy.Vector` and instantiate it as `Vector(dim=3)`.

4. The other requirement is that the user installs `pgvector-python` in the same virtual environment as target-postgres, which could be achieved with [package extras](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#install-extras), e.g. `target-postgres[vector]` or by documenting known postgres SQLAlchemy extensions in the target's readme.

### Potential issues

The biggest problem I can think off is figuring out the priority with which this type override is considered. For example, in target-postgres:

1. a few JSON schema types are mapped first to postgres-specific column types, e.g. `int -> BIGINT
2. then the SDK defaults are used

But by introducing this feature, we'd expect targets to resolve in the following order:

1. `sqlalchemy_type` overrides
2. custom target implementation overrides
3. SDK defaults

edgarrmondragon:
```[tasklist]
### Tasks
- [x] Add documentation
- [ ] Update JSON typing helpers?
- [ ] Use a JSON schema vocabulary in the SCHEMA messages (https://github.com/meltano/sdk/issues/2783)
``` 

### Docs

- [Updated SQL target guide](https://meltano-sdk--2829.org.readthedocs.build/en/2829/guides/sql-target.html#use-the-x-sql-datatype-json-schema-extension)
- [`JSONSchemaToSQL` reference docs](https://meltano-sdk--2829.org.readthedocs.build/en/2829/classes/singer_sdk.connectors.sql.JSONSchemaToSQL.html#singer_sdk.connectors.sql.JSONSchemaToSQL.register_sql_datatype_handler)

### Related

- #1323
- #1872
- Closes #2102
- #2783

### References

1. [JSON Schema blog: The Last Breaking Change](https://json-schema.org/blog/posts/the-last-breaking-change)
2. [JSON Schema blog: Custom Annotations Will Continue](https://json-schema.org/blog/posts/custom-annotations-will-continue)

----
📚 Documentation preview 📚: https://meltano-sdk--2829.org.readthedocs.build/en/2829/

### DIFF
--- a/docs/guides/sql-target.md
+++ b/docs/guides/sql-target.md
@@ -50,3 +50,56 @@ class MyConnector(SQLConnector):
         to_sql.register_format_handler("uri", URI)
         return to_sql
 ```
+
+### Use the `x-sql-datatype` JSON Schema extension
+
+You can register new type handlers for the `x-sql-datatype` extension:
+
+```python
+from my_sqlalchemy_dialect import URI
+
+
+class MyConnector(SQLConnector):
+    @functools.cached_property
+    def jsonschema_to_sql(self):
+        to_sql = JSONSchemaToSQL()
+        to_sql.register_sql_datatype_handler("smallint", sa.types.SMALLINT)
+        return to_sql
+```
+
+Then you can annotate the tap' catalog to specify the SQL type:
+
+````{tab} meltano.yml
+```yaml
+# https://docs.meltano.com/concepts/plugins/#schema-extra
+plugins:
+  extractors:
+  - name: tap-example
+    schema:
+      addresses:
+        number:
+          x-sql-datatype: smallint
+```
+````
+
+````{tab} JSON catalog
+```json
+{
+  "streams": [
+    {
+      "stream": "addresses",
+      "tap_stream_id": "addresses",
+      "schema": {
+        "type": "object",
+        "properties": {
+          "number": {
+            "type": "integer",
+            "x-sql-datatype": "smallint"
+          }
+        }
+      }
+    }
+  ]
+}
+```
+````
--- a/singer_sdk/_singerlib/schema.py
+++ b/singer_sdk/_singerlib/schema.py
@@ -42,6 +42,8 @@ STANDARD_KEYS = [
     "anyOf",
     "patternProperties",
     "allOf",
+    # JSON Schema extensions
+    "x-sql-datatype",
 ]
 
 
@@ -84,6 +86,9 @@ class Schema:
     contentMediaType: str | None = None  # noqa: N815
     contentEncoding: str | None = None  # noqa: N815
 
+    # JSON Schema extensions
+    x_sql_datatype: str | None = None
+
     def to_dict(self) -> dict[str, t.Any]:
         """Return the raw JSON Schema as a (possibly nested) dict.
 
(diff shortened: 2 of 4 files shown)


maxmarcon:
Hi @edgarrmondragon  and thanks for this.

Question: is this already working out of the box? I tried this in my tap:

```yml
 - name: tap-tbii-raw
    schema:
      tbii_raw-country_codes:
        id: 
          x-sql-datatype: integer         
```

I see the `x-sql-datatype` in the generated Singer JSON:

```json
 "stream": "tbii_raw-country_codes",
  "schema": {
    "properties": {
      "id": {
        "x-sql-datatype": "integer"
      },
 ...     
```

However, when I import it with `target-postgres`, `id` is created with type `text`.

Am I missing something?

Local plugin versions:

```
target-postgres v0.6.3, Meltano SDK v0.53.7
tap-postgres v0.8.3, Meltano SDK v0.53.7
```

Thanks!

edgarrmondragon:
Hey @maxmarcon!

I think this is a legit bug (to what extent it's a bug in target-postgres vs Meltano I haven't decided).

Can you try adding a dummy `type` to your schema override? That is

```yaml
 - name: tap-tbii-raw
   schema:
    tbii_raw-country_codes:
      id: 
        type: integer  # the value doesn't really matter
        x-sql-datatype: integer         
```

edgarrmondragon:
[target-postgres v0.7.0](https://github.com/MeltanoLabs/target-postgres/releases/tag/v0.7.0) has a fix for this.

Thanks again @maxmarcon for reporting!

maxmarcon:
@edgarrmondragon you're welcome! Thank you for fixing this. I can confirm it worked for me 👍

PR #2835, opened 2025-01-21
feat: Load project `.env` for built-in tests

ReubenFrankel:
In my VSCode config, I have

```
  "python.envFile": "",
```

to stop it auto-loading the contents of my project `.env` file into the environment. A drawback of doing this is that Pytest discovery fails for my SDK plugin projects  with `ConfigValidationError`s until I manually export the `.env` or temporarily provide sensitive config in `tests/test_core.py` and hope I don't accidentally commit it. :grimacing:

This change loads the project `.env` file before test discovery runs. I was only able to get in working in `tests/test_core.py` in the limited time that I had today (which means this would really only be beneficial for new SDK taps/targets), but let me know if it is possible somewhere that would mean I don't have to update all my plugin `tests/test_core.py` files manually (e.g. in `TapTestClassFactory.new_test_class`).

----
📚 Documentation preview 📚: https://meltano-sdk--2835.org.readthedocs.build/en/2835/

### DIFF
--- a/cookiecutter/tap-template/{{cookiecutter.tap_id}}/tests/test_core.py
+++ b/cookiecutter/tap-template/{{cookiecutter.tap_id}}/tests/test_core.py
@@ -2,9 +2,13 @@
 
 import datetime
 
+from dotenv import load_dotenv
 from singer_sdk.testing import get_tap_test_class
 
 from {{ cookiecutter.library_name }}.tap import Tap{{ cookiecutter.source_name }}
+
+load_dotenv()
+
 
 SAMPLE_CONFIG = {
     "start_date": datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d"),


edgarrmondragon:
I think I'm missing something since I can confirm that the `.env` is loaded when running tests:

https://github.com/meltano/sdk/blob/d1ba40003a516c8d0c42b9fd81ca8a826de3d392/singer_sdk/testing/factory.py#L99-L100

and then

https://github.com/meltano/sdk/blob/d1ba40003a516c8d0c42b9fd81ca8a826de3d392/singer_sdk/plugin_base.py#L191-L193

and

https://github.com/meltano/sdk/blob/d1ba40003a516c8d0c42b9fd81ca8a826de3d392/singer_sdk/plugin_base.py#L271-L284

If you run pytest with the `-s` flag, you should be able to see something like

```
2025-01-21 11:09:25,719 | INFO     | singer_sdk.configuration._dict_config | Parsing 'account_id' config from env variable 'TAP_CHECKLY_ACCOUNT_ID'.
2025-01-21 11:09:25,719 | INFO     | singer_sdk.configuration._dict_config | Parsing 'token' config from env variable 'TAP_CHECKLY_TOKEN'.
```

even though the env vars are **not** set in the shell where this is running.

edgarrmondragon:
It might also be worth for you to look at the `rootdir` value that's printed in the pytest header. The `.env` should be there for config to be loaded correctly.

```
=========================================================================== test session starts ===========================================================================
platform darwin -- Python 3.13.0, pytest-8.3.4, pluggy-1.5.0 -- /path/to/tap-checkly/.venv/bin/python
cachedir: .pytest_cache
rootdir: /path/to/tap-checkly
```

Issue #953, opened 2022-09-06
[Feature]: Automatically register stream classes as they are declared


edgarrmondragon:
### Feature scope

Taps (catalog, state, stream maps, etc.)

### Description

The current pattern we enforce through our documentation and cookiecutter for adding new streams to an HTTP tap is implement a custom `MyTap.discover_streams` that returns a list of the desired stream instances.

This pattern may provide some consistency but it could be mildly annoying for a developer to remember they have to:

1. import the new stream class, `from tap_example.streams import NewStream`
2. add the class to a module-level collection, `STREAM_CLASSES = [NewStream]`
3. implement (once) the `discover_streams` method:

   ```python
   class TapExample(Tap):
     def discover_streams(self):
       return [stream_class(self) for stream_class in STREAM_CLASSES]
   ```

---

A better approach might get inspiration from other OOP frameworks, like Django. There, database models/tables are defined as classes and whenever some abstraction is required for all models, it's [explicitly declared as _abstract_](https://docs.djangoproject.com/en/4.1/topics/db/models/#abstract-base-classes), so it doesn't map to a table in the database.

So, a similar thing in the SDK might look like this:

```python
class MyAPIStream(RESTStream, abstract=True):
  ...

class UsersStream(MyAPIStream):
  name = "users"
```

Under the hood

- `MyAPIStream` would **not** be registered, as it's declared to be _abstract_,
- `UsersStream` would be registered,
- The default implementation of `discover_streams` would read `UsersStream` from the underlying class registry and instantiate them appropriately.

aaronsteers:
@edgarrmondragon - I like this idea. Would we also need to be able to tackle `Stream` classes like `SQLStream` in which discovery can produce multiple stream instances of the same class?

What about designing `discover_catalog_entries()` as a class method on the base `Stream` class, adapted from [`SQLConnector.discover_catalog_entries()`](https://sdk.meltano.com/en/latest/classes/singer_sdk.SQLConnector.html#singer_sdk.SQLConnector.discover_catalog_entries)? Normal stream classes could simply return a single entry based on dynamic or static `schema` definition. And classes like `SQLStream` could implement this differently, in order to return multiple (or zero) catalog entries. In the case of the `SQLStream` class, the stream class could in theory call out to the `SQLConnector` method we have today. 

Side note: some stream types will require the `config` dictionary in order to detect schema dynamically, so if we go this direction, we may need to pass the `config` to the class method.

edgarrmondragon:
[...]

PR #2888, opened 2025-02-26
feat(taps): Abstract streams

edgarrmondragon:
## Related

- https://github.com/meltano/sdk/discussions/2880
- https://github.com/meltano/sdk/discussions/2846

## Summary by Sourcery

Add support for abstract streams, which are not included in the catalog.

----
📚 Documentation preview 📚: https://meltano-sdk--2888.org.readthedocs.build/en/2888/


singer_sdk/streams/core.py
```diff
@@ -1166,7 +1175,7 @@ def _sync_records(  # noqa: C901
                         )
                         raise
 
-                    if selected:
+                    if not self.__abstract__ and selected:
```
ReubenFrankel, line 1178:
```suggestion
                    if not selected:
```

singer_sdk/streams/core.py
```diff
@@ -1234,7 +1243,7 @@ def sync(self, context: types.Context | None = None) -> None:
             self._write_replication_key_signpost(context, signpost)
 
         # Send a SCHEMA message to the downstream target:
-        if self.selected:
+        if not self.__abstract__ and self.selected:
```
ReubenFrankel, line 1246:
```suggestion
        if not self.selected:
```

tests/core/test_streams.py
```diff
@@ -728,9 +728,13 @@ class UnselectedStream(SelectedStream):
...
         selected_by_default = False
 
+    class AbstractStream(RESTStream):
+        url_base = "https://example.com"
+        __abstract__ = True
+
     class MyTap(tap_class):
         def discover_streams(self):
-            return [SelectedStream(self), UnselectedStream(self)]
+            return [SelectedStream(self), UnselectedStream(self), AbstractStream(self)]
```
ReubenFrankel, line 737: Would it be possible to avoid this somehow? As in, auto-discover/instantiate any parent streams that are abstract?

singer_sdk/streams/core.py
ReubenFrankel: Personally, I would remove the redundant assignment

```py
        selected = self.selected
```

above for clarity. Then this becomes

 ```suggestion
                    if not self.selected:
```

as below.

File: singer_sdk/tap_base.py [hunk 4]
```diff
@@ -338,6 +338,7 @@ def _singer_catalog(self) -> Catalog:
         return Catalog(
             (stream.tap_stream_id, stream._singer_catalog_entry)  # noqa: SLF001
             for stream in self.streams.values()
+            if not stream.__abstract__
```
ReubenFrankel, line 341: Advantages/disadvantages of filtering out abstract streams here vs in `self.streams`?

tests/core/test_parent_child.py
```diff
@@ -264,3 +269,59 @@ def discover_streams(self):
...
+        """An abstract parent stream."""
+
+        __abstract__ = True
+
+        def get_records(self, context):  # noqa: ARG002
+            yield {"id": 1}
+            yield {"id": 2}
+
+        def generate_child_contexts(self, record, context):  # noqa: ARG002
+            yield {"pid": record["id"]}
```
ReubenFrankel, line 287: Wondering if there is a better way to support this... It would be good if this could be encapsulated into `generate_child_contexts` only, but that currently requires at least one record (hence `get_records` implementation).

tests/core/test_streams.py
edgarrmondragon: That's a very good point. I'll try to do that.

ReubenFrankel: Think this would relate to https://github.com/meltano/sdk/issues/953

PR #2903, opened 2025-03-17
fix: Don't warn about catalog schema not matching records if there are no records available to test against

ReubenFrankel:
https://github.com/ReubenFrankel/tap-f1/actions/runs/13897716477/job/38882050908?pr=111#step:7:21

## Summary by Sourcery

Bug Fixes:
- Fixes a bug where a warning was raised about catalog schema mismatches even when no records were available to test against.

----
📚 Documentation preview 📚: https://meltano-sdk--2903.org.readthedocs.build/en/2903/

### DIFF
--- a/singer_sdk/testing/tap_tests.py
+++ b/singer_sdk/testing/tap_tests.py
@@ -121,12 +121,17 @@ class StreamCatalogSchemaMatchesRecordTest(StreamTestTemplate):
 
     def test(self) -> None:
         """Run test."""
-        stream_transformed_keys = set(
+        if not self.stream_records:
+            return
+
+        stream_transformed_keys: set[str] = set(
             self.stream.stream_maps[-1].transformed_schema["properties"].keys(),
         )
-        stream_record_keys = set().union(*(d.keys() for d in self.stream_records))
+        stream_record_keys: set[str] = set().union(
+            *(d.keys() for d in self.stream_records)
+        )
         diff = stream_transformed_keys - stream_record_keys
-        if stream_record_keys and diff:
+        if diff:
             warnings.warn(
                 UserWarning(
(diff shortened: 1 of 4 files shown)


ReubenFrankel:
Could also use this to suppress [`No records were available to test` warnings](https://github.com/meltano/sdk/blob/fa14033fdacd9f697bc2beb50a07597fb3703048/singer_sdk/testing/templates.py#L243-L246) as a result of accessing `non_null_attribute_values`, when streams are explicitly ignored with `ignore_no_records_for_streams` suite config: 

https://github.com/ReubenFrankel/tap-f1/actions/runs/13897716477/job/38882050908?pr=111#step:7:24

i.e. 
```py
        if (
            not values
            and self.stream.name not in self.config.ignore_no_records_for_streams
        ):
            warnings.warn(
                UserWarning("No records were available to test."),
                stacklevel=2,
            )
```

ReubenFrankel:
Perhaps this run summary is a better example: https://github.com/ReubenFrankel/tap-f1/actions/runs/13898773243 Ideally, I don't want to see any of these warnings.

singer_sdk/testing/tap_tests.py
```diff
@@ -126,7 +126,7 @@ def test(self) -> None:
         )
         stream_record_keys = set().union(*(d.keys() for d in self.stream_records))
         diff = stream_transformed_keys - stream_record_keys
-        if diff:
+        if stream_record_keys and diff:
```
edgarrmondragon, line 129: Do you think returning early if `self.stream_records` is empty would also work here?

edgarrmondragon (approved):

edgarrmondragon:
Thanks @ReubenFrankel!

PR #3037, opened 2025-05-13
fix: Avoid aborting parent stream before child stream is finished in standard tap tests

edgarrmondragon:
## Summary by Sourcery

Bug Fixes:
- Only apply the dry-run abort record limit to streams without child streams, ensuring parent streams complete properly.


ReubenFrankel:
Amazing! Thanks @edgarrmondragon

PR #3040, opened 2025-05-14
fix: Avoid Emit empty state on target drain all

joaopamaral:
In some cases, a target may emit an empty state, which can lead to unexpected behavior when saving the state. Specifically, this could result in a previously valid state being overwritten, potentially causing state loss. This fix ensures that empty states will not be emitted.

### DIFF
--- a/singer_sdk/target_base.py
+++ b/singer_sdk/target_base.py
@@ -434,14 +434,14 @@ class Target(PluginBase, SingerReader, metaclass=abc.ABCMeta):
     def _process_state_message(self, message_dict: dict) -> None:
         """Process a state message. drain sinks if needed.
 
-        If state is unchanged, no actions will be taken.
+        If state is unchanged or empty, no actions will be taken.
 
         Args:
             message_dict: TODO
         """
         self._assert_line_requires(message_dict, requires={"value"})
         state = message_dict["value"]
-        if self._latest_state == state:
+        if not state or self._latest_state == state:
             return
         self._latest_state = state
 
@@ -511,7 +511,8 @@ class Target(PluginBase, SingerReader, metaclass=abc.ABCMeta):
         if is_endofpipe:
             for sink in self._sinks_active.values():
                 sink.clean_up()
-        self._write_state_message(state)
+        if state:  # avoid emit empty state
+            self._write_state_message(state)
         self._reset_max_record_age()
 
     @t.final


singer_sdk/target_base.py
```diff
@@ -530,10 +530,13 @@ def _write_state_message(self, state: dict) -> None:
         Args:
             state: TODO
         """
-        state_json = json.dumps(state)
-        self.logger.info("Emitting completed target state %s", state_json)
-        sys.stdout.write(f"{state_json}\n")
-        sys.stdout.flush()
+        if state:
```
edgarrmondragon, line 533: We could also make this change on line 493

```diff
- if self._latest_state is not None:
+ if self._latest_state:
```

would that work?

edgarrmondragon (commented):
Hey @joaopamaral 👋

I'm actually a bit on the fence about this change. An empty incoming STATE message could be legit. I'm curious which taps are emitting an empty state that need this treatment.

I also guess this is needed on your end now since we changed the implementation of https://github.com/meltano/meltano/pull/9258 to skip saving the target output state _iff_ it fails to parse it.

singer_sdk/target_base.py
joaopamaral: Oh, my PR was outdated! 🤦 I didn't see this change! Yep, I think `if self._latest_state:` should work. Let me change it.

joaopamaral: Changed in https://github.com/meltano/sdk/pull/3040/commits/bfb6d84546ea4a3f611e3f8c8b37da05fc5c8f6c

joaopamaral:
> I'm actually a bit on the fence about this change. An empty incoming STATE message could be legit. I'm curious which taps are emitting an empty state that need this treatment.

At the moment, I’m only encountering this issue with a custom version of tap-airbyte-wrapper, where we’re not using Docker. Instead, we’re relying on Yarn to start remote containers. It’s a bit hacky—we’re exchanging messages through mounted folders/files since we can’t access container logs directly. However, I’m not sure if these modifications are the cause of the empty states.

I did notice that the main tap-airbyte-wrapper repo includes some examples with empty states (e.g., [this test fixture](https://github.com/MeltanoLabs/tap-airbyte-wrapper/blob/e874473c2b0806bd6cc5b86d0001f041db7699da/tests/fixtures/KPHX.singer#L367)), so it might be somewhat expected. I still need to dig into that part of the code to be sure.

> I also guess this is needed on your end now since we changed the implementation of https://github.com/meltano/meltano/pull/9258 to skip saving the target output state iff it fails to parse it.

Yep, this fixes one of the issues (thanks for merging it) where external clients that we use in target sometimes send an error message before failing in meltano container (not logged in singer format), and Meltano was trying to interpret it as a message—causing a state read error and resulting in an empty state.

joaopamaral:
> An empty incoming STATE message could be legit

I’m still trying to think of a case where an empty incoming STATE message would be legitimate. Typically, if we need to force a clean state, we’d use the meltano state CLI, right? And if the tap emits a new state, Meltano will just persist that. So would this only happen if the tap was updated to a version that no longer save states at all?

joaopamaral:
I can bump the sdk version of my target to verify if https://github.com/meltano/sdk/pull/3034 will fix this issue and reopen if not. Wdyt?

edgarrmondragon:
> I can bump the sdk version of my target to verify if #3034 will fix this issue and reopen if not. Wdyt?

Let's keep this open in case that patch is not enough.

edgarrmondragon (approved):
Thinking about this more, it makes that sense to me to skip empty states. If the user manages state manually, no state output simply means not updating the state file. Otherwise if there's any sort of automation around the state file (e.g. with Meltano), then there are other and better ways of clearing bookmarks.

Thanks @joaopamaral!

PR #3108, opened 2025-06-16
fix(taps): Suppress unmapped warning for additional properties

ReubenFrankel:
Similar fix to #3031/#3048 for taps.

## Summary by Sourcery

Suppress warnings for unmapped properties when additionalProperties is enabled and update tests to verify no warnings are emitted for additional properties

Bug Fixes:
- Only track unmapped properties when additionalProperties is disabled in _conform_record_data_types

Tests:
- Wrap the conformity call in caplog and assert no warnings are logged for additionalProperties


ReubenFrankel:
@edgarrmondragon I had another test to reaffirm behaviour with no additional properties: #3109

PR #3109, opened 2025-06-16
test: Test unmapped properties warning still shows when no additional properties specified

ReubenFrankel:
Follow-up #3108

## Summary by Sourcery

Add an autouse fixture to clear the unmapped-properties warning cache between tests and add a parametrized test to ensure that `conform_record_data_types` emits a warning and filters out extra fields when `additional_properties` is set to `False` or `None`.

Tests:
- Introduce an autouse pytest fixture to reset the `_warn_unmapped_properties` cache after each test.
- Add a parametrized test verifying that unmapped object properties trigger a warning and are dropped when `additional_properties` is `False` or `None`.

### DIFF
--- a/tests/core/test_typing.py
+++ b/tests/core/test_typing.py
@@ -14,7 +14,6 @@ import sqlalchemy as sa
 from singer_sdk.helpers._typing import (
     TypeConformanceLevel,
     _conform_primitive_property,
-    _warn_unmapped_properties,
     conform_record_data_types,
 )
 from singer_sdk.typing import (
@@ -31,12 +30,6 @@ if t.TYPE_CHECKING:
     from pytest_benchmark.fixture import BenchmarkFixture
 
 logger = logging.getLogger("log")
-
-
-@pytest.fixture(autouse=True)
-def reset__warn_unmapped_properties_log_cache():
-    yield
-    _warn_unmapped_properties.cache_clear()
 
 
 def test_simple_schema_conforms_types():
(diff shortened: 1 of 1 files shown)


ReubenFrankel:
> Thanks for following up @ReubenFrankel, let me know when this is ready for a review.

Good to go now I think :+1:

tests/core/test_typing.py
```diff
@@ -32,6 +33,12 @@
 logger = logging.getLogger("log")
 
 
+@pytest.fixture(autouse=True)
+def reset__warn_unmapped_properties_log_cache():
+    yield
+    _warn_unmapped_properties.cache_clear()
```
ReubenFrankel: This was necessary to get subsequent parametrized runs of `test_conform_object_no_additional_properties` to pass. Maybe there's a better way e.g. make the stream name unique for the parametrized request to avoid a cache hit.

edgarrmondragon: > Maybe there's a better way e.g. make the stream name unique for the parametrized request to avoid a cache hit.

That seems like a more elegant solution. Do you think the effort is small enough to do it in this PR? OK if not.

ReubenFrankel: Yeah, I'll do that this evening.

ReubenFrankel: https://github.com/meltano/sdk/pull/3109/commits/23322bfe1deb451fe3bc31dcf1ba82e22d1821df

edgarrmondragon (approved):
Thanks @ReubenFrankel

PR #3141, opened 2025-06-25
feat: Support disabling stream pagination logic

ReubenFrankel:
Support disabling pagination for a stream by returning `None` from `get_new_paginator`. Useful when a client class defines `get_new_paginator` that isn't applicable for some child classes (i.e. endpoints do not support pagination).

### DIFF
--- a/cookiecutter/tap-template/{{cookiecutter.tap_id}}/{{cookiecutter.library_name}}/rest-client.py
+++ b/cookiecutter/tap-template/{{cookiecutter.tap_id}}/{{cookiecutter.library_name}}/rest-client.py
@@ -139,7 +139,7 @@ class {{ cookiecutter.source_name }}Stream({{ cookiecutter.stream_type }}Stream)
 {%- endif %}
         return {}
 
-    def get_new_paginator(self) -> BaseAPIPaginator:
+    def get_new_paginator(self) -> BaseAPIPaginator | None:
         """Create a new pagination helper instance.
 
         If the source API can make use of the `next_page_token_jsonpath`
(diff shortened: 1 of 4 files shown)


singer_sdk/pagination.py
```diff
@@ -446,3 +446,19 @@ def get_next(self, response: requests.Response) -> TPageToken | None:
                 the end of pagination.
         """
         return self.stream.get_next_page_token(response, self.current_value)
+
+
+class NoOpPaginator(BaseAPIPaginator):
```
ReubenFrankel: Flexible on this name.

edgarrmondragon: More curious about the practical different between this and `SinglePagePaginator`.

ReubenFrankel: Oh, my bad! Didn't realise that was already an available paginator -- in fact, that was one of the alternative names I had in mind for this class! 😅

I guess the only difference is the `has_next` implementation, meaning that it exits the `advance` method a little earlier on. We can pretty easily add that to `SinglePagePaginator` and just use that class instead.

edgarrmondragon: > We can pretty easily add that to `SinglePagePaginator` and just use that class instead.

That makes sense to me!

cookiecutter/tap-template/{{cookiecutter.tap_id}}/{{cookiecutter.library_name}}/rest-client.py
```diff
@@ -150,7 +150,8 @@ def get_new_paginator(self) -> BaseAPIPaginator:
         other approaches, please read the guide: https://sdk.meltano.com/en/v0.25.0/guides/pagination-classes.html.
 
         Returns:
-            A pagination helper instance.
+            A pagination helper instance, or ``None`` to indicate pagination
+            is not supported.
```
edgarrmondragon, line 154: Thanks for also updating the template

edgarrmondragon (approved):

PR #3200, opened 2025-07-31
feat: Expose new variable `__original_stream_name__` in stream maps

### LINKED ISSUE #3199: feat: Access original stream name in stream maps
### Feature scope

Taps (catalog, state, tests, etc.)

### Description

The SDK exposes a `__stream_name__` variable in stream maps. If used in conjunction with `__alias__` to modify the stream name, `__stream_name__` in other mapping expressions references the alias - not the original stream name.

```yml
      stream_maps:
        '*':
          __alias__: disposition
          original_stream_name: __stream_name__
          data: record
          __else__: __NULL__
```

```json
{"type":"RECORD","stream":"disposition","record":{"original_stream_name":"disposition","data":{...},"time_extracted":"2025-07-30T19:36:10.735039+00:00"}
```

It would be useful to have a new variable to reference the original stream name (not quite sure if this a feature or bug with how `__stream_name__` is currently expected to work).

ReubenFrankel:
> The original name of the stream. Useful when `__alias__` is specified but you want to reference the original (unaliased) stream name elsewhere (e.g. property mapping). If `__alias__` is not specified, this will reference the same value as `__stream_name__`.

Closes #3199

## Summary by Sourcery

Add support for an __original_stream_name__ variable in stream maps by passing the original stream name into the mapper, exposing it in the transform context, and including tests and documentation updates

New Features:
- Expose __original_stream_name__ variable in stream mapping context to reference the original (unaliased) stream name

Documentation:
- Document the new __original_stream_name__ variable and update __stream_name__ description

Tests:
- Add tests for mapping scenarios with and without __alias__ to validate __original_stream_name__

### DIFF
--- a/singer_sdk/mapper.py
+++ b/singer_sdk/mapper.py
@@ -254,6 +254,7 @@ class CustomStreamMap(StreamMap):
         key_properties: t.Sequence[str] | None,
         map_transform: dict,
         flattening_options: FlatteningOptions | None,
+        *,
         stream_name: str | None = None,
     ) -> None:
         """Initialize mapper.


singer_sdk/mapper.py
```diff
@@ -254,17 +254,19 @@ def __init__(
         key_properties: t.Sequence[str] | None,
         map_transform: dict,
         flattening_options: FlatteningOptions | None,
+        stream_name: str | None = None,
```
ReubenFrankel, line 258: Made this a kwarg to avoid a breaking change. Logically I think it should be higher up next to `stream_alias` in the method signature.

edgarrmondragon: I think we can safely make these all keyword-only, but no need to do that in this PR.
```suggestion
        *,
        stream_name: str | None = None,
```

PR #3228, opened 2025-08-13
refactor: Move exception when schema is not provided to the stream outside of the constructor

edgarrmondragon:
## Summary by Sourcery

Refactor schema handling in Stream class to defer validation until the schema property is accessed and improve exception documentation

Enhancements:
- Change internal schema attribute to allow None and defer missing-schema validation to the schema property
- Move schema presence check out of the constructor and into the schema getter
- Enhance exception messages and update constructor docstrings to clarify ValueError and FileNotFoundError conditions

### DIFF
--- a/singer_sdk/streams/core.py
+++ b/singer_sdk/streams/core.py
@@ -130,8 +130,9 @@ class Stream(metaclass=abc.ABCMeta):  # noqa: PLR0904
             name: Name of this stream.
 
         Raises:
-            ValueError: TODO
-            FileNotFoundError: TODO
+            ValueError: If the stream name is not provided.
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
> FYI, I think this change is the root cause of [MeltanoLabs/tap-hubspot#180](https://github.com/MeltanoLabs/tap-hubspot/issues/180)

Thanks for investigating @ReubenFrankel, and sorry about the misguidance about your tap-hubspot PR!

I guess this change surfaced exactly the sort of thing I'd like to avoid in implementations (side effects in `__init__` methods).

ReubenFrankel:
All good -- agreed the side effect was pretty odd anyway 😅

PR #3340, opened 2025-10-22
fix(targets): Track table preparation to prevent data loss on schema changes

edgarrmondragon:
## Summary by Sourcery

Prevent data loss when target schemas change mid-stream by tracking table preparation state and only dropping tables on the first overwrite, then incrementally adding new columns thereafter.

Bug Fixes:
- Fix unexpected table drops and data loss on schema changes in OVERWRITE mode

Enhancements:
- Introduce a preparation cache to ensure tables are only dropped once per run and subsequent schema updates add columns incrementally

Tests:
- Add unit tests to verify tables aren’t dropped and data is preserved during schema evolution
- Add an end-to-end SQLite integration test for successive schema changes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,7 +51,7 @@ repos:
   - id: check-readthedocs
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.4
+  rev: v0.14.5
   hooks:
   - id: ruff-check
     args: [--fix, --exit-non-zero-on-fix, --show-fixes]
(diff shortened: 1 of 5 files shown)


edgarrmondragon:
Thanks @kgpayne!

I think this now actually addresses the issue.

> 1. Only drop-and-recreate once per Stream. I.e. set target load method to `TargetLoadMethods.APPEND_ONLY` or `UPSERT` on subsequent instances of SQLSink for a given stream.

This does that effectively, though not by changing the load method on the fly.

edgarrmondragon:
My plan is to merge this tomorrow and make a release in the coming days.

PR #3448, opened 2026-01-14
docs: Fix typos in parent-child streams

ralichkov:
Hi! I noticed that the code example for [Parent-Child Streams](https://sdk.meltano.com/en/latest/parent_streams.html) page has a typo in the property name so it was acting as a no-op for me.

This simply changes `ignore_parent_replication_keys` to `ignore_parent_replication_key` and adds a missing whitespace earlier in the document.

## Summary by Sourcery

Correct documentation for parent-child streams to fix a misleading configuration example.

Documentation:
- Fix the documented child stream configuration flag name for ignoring parent replication keys.
- Tidy up formatting in the parent-child streams documentation by adding missing whitespace.

### DIFF
--- a/docs/parent_streams.md
+++ b/docs/parent_streams.md
@@ -19,7 +19,7 @@ from a parent record each time the child stream is invoked.
 1. If the number of _parent_ items is very large (thousands or tens of thousands), you can
    optionally set [`state_partitioning_keys`](singer_sdk.Stream.state_partitioning_keys) on the child stream to specify a subset of context keys to use
    in state bookmarks. (When not set, the number of bookmarks will be equal to the number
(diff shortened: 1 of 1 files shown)


ralichkov:
I noticed that building the docs hasn't modified the files in `.http_cache`, but I'm not sure of the details there.

edgarrmondragon:
> I noticed that building the docs hasn't modified the files in `.http_cache`, but I'm not sure of the details there.

If you're curious, that cache is used for integration testing: https://github.com/search?q=repo%3Ameltano%2Fsdk%20CachedSession&type=code. So we don't hammer those third-party APIs on each run of the tests.

PR #3517, opened 2026-02-26
feat(taps): Allow REST streams to ignore certain errors and continue on to the next stream/context

### LINKED ISSUE #1689: feat: SkipAPIError to compliment RetriableAPIError and FatalAPIError

edgarrmondragon:
SSIA

## Summary by Sourcery

Add support for treating certain REST stream errors as ignorable so affected requests are skipped without failing the sync.

Enhancements:
- Log a warning and skip further processing when an ignorable error occurs during REST stream record requests.

Tests:
- Add tests covering propagation of IgnorableAPIError from REST response validation and verifying that REST streams skip records and log appropriately on ignorable errors.


ReubenFrankel:
This will be nice to replace our custom `ResumableAPIError` implementation that we have in a couple of taps with! 😁 (thanks for following up even after I forgot to open an issue)

PR #3565, opened 2026-03-13
perf(mappers): Optimize stream map expression evaluation

edgarrmondragon:
## Summary

After the simpleeval compatibility fix was merged in #3595, this PR contains stream map performance optimizations — a **34% improvement** on `test_bench_simple_map_transforms` (547 ms → 407 ms).

### Optimizations

**1. Pre-build static `names` dict at init time**

Entries that never change across records — `config`, `__stream_name__`, `__original_stream_name__`, and `fake` — are assembled once in `__init__` into `_static_eval_names`. Previously these were re-inserted into a fresh dict inside `_eval` on every single property of every single record.

**2. Build the `names` dict once per record, not once per field**

The `transform_fn` closure now calls `_build_eval_names(record)` once and passes the result to every `_eval` call for that record. `_eval` only patches the `self` key (the current property's original value) before each evaluation. Previously a full `record.copy()` plus several dict assignments happened inside `_eval` for every transformed property.

**3. Pre-evaluate constant expressions at init time**

`_is_constant_expression` inspects the parsed AST of each mapping expression. If all `Name` nodes refer to known function names (and therefore not to record variables), the expression is evaluated once during `_init_functions_and_schema` and its result is cached. For every subsequent record the cached value is used directly, bypassing simpleeval entirely.

## Related

- Closes https://github.com/meltano/sdk/issues/3561
- https://github.com/danthedeckie/simpleeval/pull/176
- #3595 (simpleeval 1.0.5 compatibility, already merged to `main`)
- #3605 / #3606 (debug-log removal, merged separately — no measured performance impact per CodSpeed)

## Summary by Sourcery

Optimize stream map expression evaluation in the mapper for improved performance while preserving existing behavior.

Enhancements:
- Precompute a static evaluation context for config and stream metadata and reuse it across record evaluations.
- Build the expression evaluator names dictionary once per record and reuse it for all property transformations within that record.
- Introduce detection and pre-evaluation of constant mapping expressions so their results can be reused without re-invoking the expression evaluator.
- Extend the mapper function type aliasing to support simpleeval module wrappers alongside callables.

### DIFF
--- a/singer_sdk/mapper.py
+++ b/singer_sdk/mapper.py
@@ -39,6 +39,8 @@ if t.TYPE_CHECKING:
     from singer_sdk.helpers._flattening import FlatteningOptions
     from singer_sdk.singerlib.catalog import Catalog
 
+FunctionsDict: t.TypeAlias = dict[str, t.Callable | simpleeval.ModuleWrapper]
+
 
 MAPPER_ELSE_OPTION = "__else__"
 MAPPER_FILTER_OPTION = "__filter__"
@@ -348,6 +350,20 @@ class CustomStreamMap(StreamMap):
         self.faker_config = faker_config
         self.stream_name = stream_name
 
+        self.expr_evaluator = _MapperEval(functions=self.functions)
+        self.fake = self._init_faker_instance()
+
+        # Pre-build the static portion of the names dict (entries that never change
+        # across records).  Record-specific entries (_/record/self) are merged in at
+        # eval time.
+        self._static_eval_names: dict[str, t.Any] = {
+            "config": self.map_config,
+            "__stream_name__": self.stream_alias,
+            "__original_stream_name__": self.stream_name,
+        }
+        if self.fake:
+            self._static_eval_names["fake"] = self.fake
+
         self._transform_fn: t.Callable[[dict], dict | None]
         self._filter_fn: t.Callable[[dict], bool]
         (
@@ -355,9 +371,6 @@ class CustomStreamMap(StreamMap):
             self._transform_fn,
             self.transformed_schema,
         ) = self._init_functions_and_schema(stream_map=map_transform)
-        self.expr_evaluator: simpleeval.EvalWithCompoundTypes = _MapperEval(
-            functions=self.functions,
-        )
         self.fake = self._init_faker_instance()
 
     def transform(self, record: dict) -> dict | None:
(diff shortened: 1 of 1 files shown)


edgarrmondragon:
@ReubenFrankel I think we'll have to accept the performance degradation. For context, the test applies a simple stream map to a stream with 1000 records:

https://github.com/meltano/sdk/blob/3e13f8f6bfadae920de39a983c7582ae64d25a88/tests/core/test_mapper.py#L1047-L1070

I managed to get it down from ~40% to 17-20% in https://github.com/danthedeckie/simpleeval/pull/176. Replacing the modules with individual functions I believe only wins us ~2%, so probably not worth the disruption.

What do you think of this in the context of folks using this feature in production?

ReubenFrankel:
> What do you think of this in the context of folks using this feature in production?

If this is across all stream maps expressions, a -20% performance hit is not great. Does that compound for multiple expressions too? If the current exposure of `datetime` and `json` modules doesn't pose any security risk to the user, then maybe we just sit on `<1.0.5` for a while and figure out a better approach...?

> or subclass SimpleEval to bypass it

Did you try this? That might allow us to workaround the performance overhead of `ModuleWrapper`.

ReubenFrankel:
> > or subclass SimpleEval to bypass it

I tried this in https://github.com/meltano/sdk/pull/3565/commits/558f48774de1fc3fd025f196aa8600e03396be35 but the performance was substantially worse. Reverted now.

edgarrmondragon:
I managed to improve performance somewhat in 33cd0d93dea075980aaa5b71da80b325d4af0ef2 by doing some caching and avoiding repeated `.copy()`s of the record dict, and it resulted in offsetting the regression introduced by `SimpleEval._check_disallowed_items()`.

I'm tempted to backport some of that.

ReubenFrankel:
> it resulted in offsetting the regression introduced by SimpleEval._check_disallowed_items()

Nice! 🔥

edgarrmondragon:
This PR now focuses entirely on performance optimizations, after the compatibility fix was implemented in

- https://github.com/meltano/sdk/pull/3595

I had Claude split the 4 performance optimizations into separate commits. I might open individual PRs for those, but the most urgent item of fixing compatibility while not regressing on performance has been addressed.

PR #3659, opened 2026-06-01
feat(targets): Report all record validation errors instead of only the first

### LINKED ISSUE #2516: feat: Log all JSON schema validation errors in sinks/targets
### Feature scope

### Description

The default JSON schema validation implementation uses the `.validate()`[^1] method. This means that only _one_ schema validation failure is logged to the user:

https://github.com/meltano/sdk/blob/6638a908736d42887178f5645b913b9ec3aa662d/singer_sdk/sinks/core.py#L120-L123

It might be helpful to instead display all validation errors, and we could use the `.iter_errors()` method for that, raising an appropriate `InvalidRecord` exception.

[^1]: https://python-jsonschema.readthedocs.io/en/stable/validate/

Lemonnycodes:
### What
`JSONSchemaValidator.validate()` used `validate()`, which raises on the first schema error only. This switches to `iter_errors()` so every violation in a record is collected and reported together in a single `InvalidRecord`.

### Why
Closes #2516. When a record breaks the schema in multiple places, users currently have to fix-and-rerun repeatedly to discover each error. Aggregating them is much friendlier for debugging targets.

### How
- Replace the single `validate()` call with `sorted(iter_errors(...))` and join the messages.
- Errors are sorted for deterministic output.
- Added `test_jsonschema_validator_reports_all_errors` covering a record with three independent violations.

## Summary by Sourcery

Aggregate all JSON schema validation errors for a record instead of stopping at the first failure.

Bug Fixes:
- Ensure record validation raises an InvalidRecord containing all schema violations found in the record, rather than only the first error.

Tests:
- Add a test verifying that JSONSchemaValidator surfaces all field-level schema errors in a single InvalidRecord, including preserving the original record.


Lemonnycodes:
Thanks for the review! Addressed both points: switched to a structured `(path, message)` sort key instead of relying on the error's string repr, and each message is now prefixed with its `json_path` so it's clear which field failed. Also tightened the test to assert the full ordered message.

edgarrmondragon (approved):
Thanks @Lemonnycodes!

PR #3665, opened 2026-06-03
fix(targets): reuse SQLAlchemy inspector reflection cache

### LINKED ISSUE #3478: bug: SQLAlchemy reflection cache is defeated by creating new Inspector instances per call
### Singer SDK Version

0.53.5

### Is this a regression?

- [ ] Yes

### Python Version

NA

### Bug scope

### Operating System

_No response_

### Description

Three methods in `SQLConnector` create a new `sa.inspect(self._engine)` on every call. Since SQLAlchemy's reflection cache (`info_cache`) is an instance-level dictionary on `Inspector`, each new instance starts with an empty cache and the previous one is garbage collected — completely defeating dialect-level caching optimizations.

#### Affected methods

- [`table_exists()`](https://github.com/meltano/sdk/blob/c3176a4ec22872993c5542c5b7443f62012650c1/singer_sdk/sql/connector.py#L1206)
- [`schema_exists()`](https://github.com/meltano/sdk/blob/c3176a4ec22872993c5542c5b7443f62012650c1/singer_sdk/sql/connector.py#L1217)
- [`get_table_columns()`](https://github.com/meltano/sdk/blob/c3176a4ec22872993c5542c5b7443f62012650c1/singer_sdk/sql/connector.py#L1235) — most critical, called frequently during data loading

For contrast, [`discover_catalog_entries()`](https://github.com/meltano/sdk/blob/c3176a4ec22872993c5542c5b7443f62012650c1/singer_sdk/sql/connector.py#L1116) already correctly creates a single Inspector and reuses it throughout the method.

#### How SQLAlchemy's reflection cache works

SQLAlchemy stores reflection results in `Inspector.info_cache`, a plain dictionary. The `@reflection.cache` decorator on dialect methods (e.g., Snowflake's `_get_schema_columns()`) uses this dict to avoid redundant queries. **New Inspector = new empty dict = all cached results lost.**

#### Real-world impact (Snowflake)

The Snowflake SQLAlchemy dialect optimizes by querying **all columns in a schema** at once (`SELECT ... FROM information_schema.columns WHERE table_schema = :schema`) and caching the result. Subsequent `get_columns()` calls for other tables in the same schema should return from cache with zero queries.

Because the SDK creates a throwaway Inspector each time, this optimization is completely negated.

This affects any SQL dialect that uses `@reflection.cache` for optimization, not just Snowflake.

#### Suggested fix

Cache the Inspector instance at the connector level (e.g., a lazy `_inspector` property) so the `info_cache` persists across calls. A `clear_cache()` escape hatch can be provided for edge cases where schema changes during a sync.

### Link to Slack/Linen

_No response_

akashmalbari:
Closes #3478.

## Summary

This updates `SQLConnector` to reuse a cached SQLAlchemy `Inspector` instance for repeated reflection calls instead of creating a new `sa.inspect(self._engine)` for each call.

The affected methods now share the connector-level Inspector:

- `table_exists()`
- `schema_exists()`
- `get_table_columns()`

This preserves SQLAlchemy's `Inspector.info_cache` across reflection calls, allowing dialect-level reflection caching optimizations to work as intended.

## Cache invalidation

Because the Inspector now persists across calls, this also adds a reflection cache clearing helper and clears the cache after connector-managed DDL operations such as:

- creating schemas
- creating tables
- adding columns
- renaming columns
- altering column types

This avoids stale reflection after the SDK itself changes the target schema.

## Tests

Added regression coverage for:

- reusing a single Inspector across repeated SQLConnector reflection calls
- clearing the cached Inspector reflection cache
- avoiding stale reflection state after connector-managed DDL

## Validation

- `uv run pytest tests/sql/test_connector.py -q`
- `uv run pytest tests/sql -q`
- `uv sync --all-groups --all-extras --all-packages`
- `uv run pytest tests -q`
- `uvx nox -t typing`
- `uvx pre-commit run --all-files`

## Summary by Sourcery

Cache and reuse a SQLAlchemy Inspector within SQLConnector to enable reflection caching and invalidate it after connector-managed schema changes.

Bug Fixes:
- Ensure SQLConnector reuses a single SQLAlchemy Inspector instance for reflection so dialect-level reflection caches are preserved across calls.
- Clear the Inspector reflection cache after connector-managed DDL operations (schema/table creation, column add/rename/type change, table recreation) to avoid stale metadata.

Documentation:
- Add AGENTS.md with contributor-facing guidance on development setup, commands, code style, exceptions, deprecation policy, architecture, testing, and project structure.

Tests:
- Add regression tests covering Inspector reuse across reflection methods, reflection cache clearing behavior, and cache invalidation after DDL operations.


edgarrmondragon:
Thanks for the PR @akashmalbari!

The problem of stale reflection makes me hesitant to accept this as-is. Developers can override many of the DDL methods here where you're calling `self.clear_reflection_cache()`, so we could easily be feeding them a stale cache and producing subtle bugs unless we go and change their code after we release this 🙃.

I'll try to think about a different approach, but in the meantime wdyt?

edgarrmondragon (changes requested):
See https://github.com/meltano/sdk/pull/3665#issuecomment-4626594140.

akashmalbari:
Thanks for the review, @edgarrmondragon.

I updated the approach so cache invalidation is no longer dependent on individual `SQLConnector` DDL helper methods calling `clear_reflection_cache()`.

The connector now registers an engine-level `after_cursor_execute` listener and clears the cached inspector whenever DDL is executed through the connector engine (`CREATE`, `ALTER`, or `DROP`). This means subclass overrides that execute DDL through `_engine` / `_connect()` will still invalidate reflection cache even if they do not call `clear_reflection_cache()` directly.

I also added a regression test covering an overridden `_create_empty_column()` implementation that omits `clear_reflection_cache()` but still gets fresh reflection after DDL.

PR #3668, opened 2026-06-05
feat: Handle GraphQL error responses

### LINKED ISSUE #1421: [Feature]: Handle GraphQL errors
### Feature scope

### Description

As documented in https://www.apollographql.com/docs/react/data/error-handling, GraphQL errors are included as an array in the response:

```json
{
  "errors": [
    {
      "message": "Cannot query field \"nonexistentField\" on type \"Query\".",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED",
        "exception": {
          "stacktrace": [
            "GraphQLError: Cannot query field \"nonexistentField\" on type \"Query\".",
            "...additional lines..."
          ]
        }
      }
    }
  ],
  "data": null
}
```

(getting JSON-RPC flashbacks 💀)

See https://github.com/MeltanoLabs/tap-github/pull/181

Dexter2099:
Closes #1421

## Summary
- detect GraphQL response payloads containing a non-empty `errors` field during response validation
- raise `FatalAPIError` with the GraphQL error message instead of treating the response as successful
- add regression coverage for GraphQL error payloads returned with HTTP 200

## Testing
- `uv run pytest tests/core/test_streams.py::test_graphql_validate_response_raises_for_errors -q`
- `uv run pytest tests/core/test_streams.py -q`
- `nox -rs tests-3.11 -- tests/core/test_streams.py`
- `pre-commit run --files singer_sdk/streams/graphql.py tests/core/test_streams.py`

## Summary by Sourcery

Handle GraphQL error payloads returned with HTTP 200 by validating the response body and surfacing errors as fatal API failures.

Bug Fixes:
- Treat GraphQL responses containing a non-empty errors field as failures and raise a FatalAPIError instead of treating them as successful responses.

Tests:
- Add regression tests covering GraphQL error payloads with HTTP 200, non-JSON and non-dict bodies, empty or missing errors fields, and alternate error structures.


singer_sdk/streams/graphql.py
```diff
@@ -31,6 +34,41 @@ class GraphQLStream(RESTStream, abc.ABC, t.Generic[_TToken]):
...
+
+        Raises:
+            FatalAPIError: If the GraphQL response contains errors.
+        """
+        super().validate_response(response)
+
+        try:
+            data = response.json()
+        except ValueError:
+            return
```
edgarrmondragon: Returning here would mean processing of the response continues as if it was valid. Should we raise `FatalAPIError(...) from ...` here?

Dexter2099: I replaced the malformed-response silent returns with `FatalAPIError`; the JSON decode path now uses `raise ... from e` so the original parsing failure is preserved.

I kept valid GraphQL payloads without errors (`missing-errors` / `empty-errors`) as non-fatal, and moved only malformed bodies into the raises test.

PR #3672, opened 2026-06-08
feat(taps): Respect rate-limit headers in the default REST backoff

### LINKED ISSUE #2012: feat: Respect `RateLimit` headers in default REST backoff implementation
### Feature scope

### Description

This [IETF draft](https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/) defines a series of `RateLimit` headers:

- [`X-RateLimit-Limit`](https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html#name-limit-keyword)
- [`X-RateLimit-Remaining`](https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html#name-remaining-keyword)
- [`X-RateLimit-Reset`](https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html#name-reset-keyword)

Taps built with the SDK around REST APIs that respond with these headers would benefit from a default `RESTStream` backoff implementation that understands and respects them - namely `X-RateLimit-Remaining`.

Nazib65:
## What & why

Closes #2012.

The default `RESTStream.backoff_wait_generator` returned `backoff.expo(factor=2)` and
never looked at the response, so taps hitting `429 Too Many Requests` ignored the
server's own `Retry-After` / `X-RateLimit-Reset` headers and waited a guessed
exponential interval instead.

Per the discussion with @ReubenFrankel and @edgarrmondragon, this implements
header-aware backoff **within the existing mechanism** (the native-retries refactor is
left for later given the open `RetriableAPIError` questions), respecting rate-limit
headers when present and falling back to exponential backoff when not.

## Changes

- `backoff_wait_generator` is now a generator using backoff's
  prime-then-`.send(exception)` protocol. It reads the response defensively
  (`getattr(exception, "response", None)`) so connection/timeout errors that carry no
  response fall back cleanly instead of raising `AttributeError`.
- New overridable **`get_wait_time_from_response(response)`** hook reads `Retry-After`
  (precedence; seconds or HTTP-date) then `X-RateLimit-Reset`.
- New `_parse_retry_after` helper for the two `Retry-After` value forms.

## Decisions worth a look

- **`X-RateLimit-Reset` is treated as seconds-until-reset (delta)** per the IETF
  RateLimit header fields draft referenced in the issue, not a Unix epoch timestamp.
  Happy to switch to epoch (or support both) if you'd prefer — it's a small change.
- `Retry-After` takes precedence over `X-RateLimit-Reset`; `0`-second waits are honoured.
- Fully backwards compatible — taps overriding `backoff_wait_generator` are unaffected.

## Tests

Added to `tests/core/rest/test_failure.py`:
- `get_wait_time_from_response` parsing: Retry-After seconds + HTTP-date, X-RateLimit-Reset,
  precedence, and unparsable/missing → exponential fallback.
- End-to-end generator drive: header → respected wait; `ConnectionError` (no response) →
  exponential fallback **without crashing**; header-less `429` → exponential continues.

Full `tests/core/rest/` passes (89); `ruff`, `ruff format`, `ty`, and `mypy` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Make the default REST retry backoff honor server-provided rate-limit delays while retaining exponential fallback behavior.

New Features:
- Respect `Retry-After` and `X-RateLimit-Reset` response headers when determining REST retry delays.
- Provide an overridable response wait-time hook supporting `Retry-After` seconds or HTTP dates.

Bug Fixes:
- Prevent connection and timeout errors without responses from crashing the retry backoff generator.
- Honor zero-second rate-limit waits and fall back safely when headers are missing or invalid.

Tests:
- Add coverage for rate-limit header parsing, precedence, date handling, clamping, invalid values, and end-to-end backoff behavior.


Nazib65:
Thanks @sourcery-ai 🙏

**Test suggestions** — both added: zero/negative clamping in `dbd417d8`, and the
generator's unparsable-header fallback (plus naive-date coverage) in `f17ce328`.

**Design points** — I made deliberate choices here, but they're genuinely open and
I'd defer to maintainer preference:

1. **`X-RateLimit-Reset` delta vs epoch** — I implemented delta-seconds per the IETF
   RateLimit-Headers draft cited in #2012. Auto-detecting delta-vs-epoch is ambiguous
   (a large delta is indistinguishable from a near-future epoch, and guessing wrong
   risks very long waits), so I kept it spec-literal and made
   `get_wait_time_from_response` overridable for epoch-style APIs. Happy to add explicit
   epoch support if preferred.

2. **Passing `response` vs the full exception** — I pass the `Response` (guaranteed
   non-`None`; the generator does the no-response guarding). This was partly to avoid the
   footgun behind connection-level crashes where a hook assumed `exception.response` was
   always populated. Glad to switch to passing the exception and document the
   `None`-response case if that flexibility is worth more.

@edgarrmondragon @ReubenFrankel — any preference on these two?

PR #3690, opened 2026-06-30
feat: support nested replication keys via dotted path

### LINKED ISSUE #1198: docs: Document how to deal with nested replication keys
### Feature scope

#### Option 1

Use a JSONPath expression to declare the possibly nested field:

```python
class MyStream(Stream):
  replication_key = "$attributes.updated"

  schema = PropertiesList(
    Property("id", IntegerType),
    Property("attributes", ObjectType(
      Property("created", DateTimeType),
      Property("updated", DateTimeType),
    )),
  ).to_dict()
```

### Option 2

Public methods to get the replication value from each record:

```python
class Stream:
  def get_replication_key_value(self, record):
    if self.replication_key:
      return record.get[self.replication_key]
    else:
      None

  def get_replication_key_schema(self):
    if self.replication_key:
      return return self.schema["properties"][self.replication_key]
    else:
      None

class MyStream(Stream):
  def get_replication_key_value(self, record):
    return record["attributes"]["updated"]

  def get_replication_key_schema(self):
    return self.schema["properties"]["attributes"]["properties"]["updated"]
    # or using toolz (https://toolz.readthedocs.io/)
    # return get_in(["properties", "attributes", "properties", "updated"], record)
```

### Current workaround

1. Declare the denested field as the new replication key and include at the top level it in the schema:

   ```python
   class MyStream(Stream):
     replication_key = "updated"

     schema = PropertiesList(
       Property("updated", DateTimeType),
     ).to_dict()
   ```

1. Override [`Stream.post_process`](https://sdk.meltano.com/en/latest/classes/singer_sdk.Stream.html#singer_sdk.Stream.post_process):

   ```python
   def post_process(row, context):
    row["updated"] = row["attributes"].pop("updated")
    return row
   ```

daria-hrabar:
## What does this PR do?

Adds support for nested replication keys in the Singer SDK by introducing a `get_nested_value()` helper in `singer_sdk/helpers/_util.py` that traverses a record dictionary level by level using a dotted key path (e.g., `"attributes.updated"`). The `_increment_stream_state()` method and `is_timestamp_replication_key` property in `singer_sdk/streams/core.py` are updated to use this traversal instead of a flat dictionary lookup, enabling tap developers to point replication keys at timestamp fields nested inside sub-objects rather than only at top-level fields.

## Why was this PR needed?

Many real-world APIs — including HubSpot, Salesforce, and GitHub — return timestamps nested inside sub-objects. Before this fix, setting `replication_key = "attributes.updated"` caused the SDK to perform a flat lookup (`record.get("attributes.updated")`), which returned `None` because no top-level key with that literal name exists. As a result, the state was never updated, and the tap silently re-synced all records on every run.

Investigation traced the root cause to two locations in `singer_sdk/streams/core.py`:

- `_increment_stream_state()` passed the raw record directly to `increment_state()`, which performed a flat key lookup returning `None` for dotted paths
- `is_timestamp_replication_key` called `schema.get("properties", {}).get(replication_key)` which only checked the schema's top level, incorrectly raising `InvalidReplicationKeyException` even when the field existed at a deeper level

The fix keeps `singer_sdk/streams/_state.py` unchanged by flattening the extracted value into a flat record before passing it to the state manager, limiting the scope of the change and reducing regression risk.

## What are the relevant issue numbers?

Closes #1198

## Does this PR meet the acceptance criteria?

- [x] Tests added for new/changed behavior
- [x] All tests passing
- [x] Follows project style guide
- [x] No breaking changes introduced

## Summary by Sourcery

Enable nested replication keys throughout stream state advancement and replication-key validation.

New Features:
- Support dotted replication keys that resolve timestamp and other replication values within nested record objects.

Bug Fixes:
- Fix incremental state updates and timestamp replication-key validation for nested schema fields, preventing repeated full resynchronization when bookmarks are nested.

Enhancements:
- Add a reusable helper for traversing nested mappings with dotted key paths and distinguishing missing paths when needed.

Tests:
- Add coverage for flat, multi-level nested, missing, invalid, and non-mapping replication-key paths, including state-manager integration.


daria-hrabar:
@edgarrmondragon I've addressed the Sourcery bot feedback, improved test coverage, and updated the PR title to follow the semantic PR conventions. Would appreciate your review and feedback when you have a chance.

edgarrmondragon (commented):
Thanks @daria-hrabar, what happens to the

- `bookmark_properties` in the SCHEMA message

   https://github.com/meltano/sdk/blob/d6279758a3090e64d8b60de382b2e3f17ff55712/singer_sdk/singerlib/encoding/simple.py#L162

- `valid_replication_keys` in the standard metadata

  https://github.com/meltano/sdk/blob/d6279758a3090e64d8b60de382b2e3f17ff55712/singer_sdk/singerlib/catalog.py#L171

?

daria-hrabar:
Hi @edgarrmondragon ,

Thank you for your comment. I hadn't considered these two fields yet. 

For `bookmark_properties`, the dotted path `"attributes.updated"` would be written as-is into the SCHEMA message since the SDK constructs it directly from `self.replication_key`. A target receiving `bookmark_properties: ["attributes.updated"]` may not know how to interpret a dotted path as a nested field, since it's just a string from the target's perspective.

Similarly, `valid_replication_keys` in the catalog metadata would list `"attributes.updated"` as a flat string, which may confuse targets or orchestrators that try to validate it against the schema's top-level properties.

I think this is a limitation of my current implementation—it handles the extraction and state tracking correctly on the tap side, but doesn't address how the dotted key is communicated to targets via SCHEMA messages or catalog metadata.

Would you suggest I leave this as a known limitation with documentation, or would you like me to attempt to handle these cases as well? I want to make sure the fix is complete before merging.

daria-hrabar:
Hi @edgarrmondragon,
I am following up on my previous comment regarding your notes on `bookmark_properties` and `valid_replication_keys`. Please let me know if I need to make any changes to my PR. Thank you for your feedback.

daria-hrabar:
Hi @edgarrmondragon,
Once again, I am kindly following up on my previous comment regarding your questions about `bookmark_properties` and `valid_replication_keys`. Please let me know if I need to make any changes to my PR. Thank you.

daria-hrabar:
Hi @edgarrmondragon,
Once again, I am kindly following up on my previous comment regarding your questions about `bookmark_properties` and `valid_replication_keys`. Please let me know if I need to make any changes to my PR. Thank you.

edgarrmondragon:
> I think this is a limitation of my current implementation—it handles the extraction and state tracking correctly on the tap side, but doesn't address how the dotted key is communicated to targets via SCHEMA messages or catalog metadata.
> 
> Would you suggest I leave this as a known limitation with documentation, or would you like me to attempt to handle these cases as well? I want to make sure the fix is complete before merging.

The problem is that you're not tying this to a particular tap where this is needed, and the current solution brings complexity for dubious benefit (with a broken SCHEMA message and catalog). I'm not sure the https://github.com/meltano/sdk/issues/1198 is something we should address, since a simple workaround exists and maybe we just need to improve the documentation with a simple focused guide.

daria-hrabar:
> > I think this is a limitation of my current implementation—it handles the extraction and state tracking correctly on the tap side, but doesn't address how the dotted key is communicated to targets via SCHEMA messages or catalog metadata.
> > Would you suggest I leave this as a known limitation with documentation, or would you like me to attempt to handle these cases as well? I want to make sure the fix is complete before merging.
> 
> The problem is that you're not tying this to a particular tap where this is needed, and the current solution brings complexity for dubious benefit (with a broken SCHEMA message and catalog). I'm not sure the #1198 is something we should address, since a simple workaround exists and maybe we just need to improve the documentation with a simple focused guide.

@edgarrmondragon so you recommend I do not change anything, and we stick to the existing workaround?

edgarrmondragon:
> @edgarrmondragon so you recommend I do not change anything, and we stick to the existing workaround?

Yes

PR #3697, opened 2026-07-10
fix: Make the `Sink.clean_up()` method a no-op by default

edgarrmondragon:
This is to avoid forcing users to call `super().clean_up()` in their implementation.

TIL that this called the ["Template Method"](https://refactoring.guru/design-patterns/template-method), and it is something we should probably do in most of our explicitly public APIs.

## Summary by Sourcery

Introduce a finalized sink shutdown step that preserves metrics behavior while making `clean_up()` a no-op by default.

Enhancements:
- Add a final `_clean_up()` method on sinks to handle end-of-stream cleanup and metric exit calls.
- Change target draining logic to invoke the new sink `_clean_up()` method instead of `clean_up()`.

Build:
- Update Ruff pre-commit hook and required Ruff version to 0.15.21 in tooling configuration.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,7 @@ repos:
   - id: check-readthedocs
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.15.20
+  rev: v0.15.21
   hooks:
   - id: ruff-check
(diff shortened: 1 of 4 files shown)


edgarrmondragon:
@sourcery-ai review

PR #3703, opened 2026-07-15
feat: Allow `record_counter` without a stream and tag only provided values

edgarrmondragon:
SSIA

## Summary by Sourcery

Make the `stream` parameter optional in `metrics.record_counter` and update tests accordingly.

New Features:
- Allow `metrics.record_counter` to be used without specifying a stream, optionally tagging by endpoint only.

Enhancements:
- Add conditional use of typing `override` for Python 3.12+ and apply it to custom test helpers.

Tests:
- Expand `test_record_counter` with parametrized cases to cover usage with no parameters, with stream only, and with endpoint only.

### DIFF
--- a/singer_sdk/metrics.py
+++ b/singer_sdk/metrics.py
@@ -376,28 +376,29 @@ def get_metrics_logger() -> logging.Logger:
 
 
 def record_counter(
-    stream: str,
+    stream: str | None = None,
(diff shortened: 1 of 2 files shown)


edgarrmondragon:
@sourcery-ai title

Issue #106, opened 2021-04-15
Add settings 'write-back' capability (Singer SDK)


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/sdk/-/issues/106

Originally created by @aaronsteers on 2021-04-15 21:08:05

---

Related to similar issue for Meltano: https://gitlab.com/meltano/meltano/-/issues/2710

Some tap and target implementations use a writeback method ([we think?](https://gitlab.com/meltano/meltano/-/issues/2710#note_659612129)) to store settings (such as auth tokens) back in the `config.json` file for future use.

There's currently no similar way for a tap developer using the SDK to write back settings for use in future invocations.

## Wanted:

For this issue, we are currently gaging feedback and asking for developers to provide use cases if this is a blocker for them. If this affects you, please post to the comments.

## Related

* https://github.com/meltano/meltano/issues/2660

MeltyBot:
[View 2 previous comments from the original issue on GitLab](https://gitlab.com/meltano/sdk/-/issues/106)

pnadolny13:
@edgarrmondragon I implemented a version of this with the SDK in tap-gohighlevel.

- I had to [override the tap init](https://github.com/MeltanoLabs/tap-gohighlevel/blob/1ff44d38831114398139d848fcdb74d3544dcf18/tap_gohighlevel/tap.py#L47) to capture the config file path and save it to the tap instance. I also chose to raise an exception if the configs are passed in outside of a config file. 
- Then when my authenticator is being instantiated by pull the config path variable out and [pass it in](https://github.com/MeltanoLabs/tap-gohighlevel/blob/1ff44d38831114398139d848fcdb74d3544dcf18/tap_gohighlevel/auth.py#L61). The way I access that var is definitely not nice but it works. 
- I overrode update_access_token to catch the new refresh token
- Then call a method [_write_back_to_config](https://github.com/MeltanoLabs/tap-gohighlevel/blob/1ff44d38831114398139d848fcdb74d3544dcf18/tap_gohighlevel/auth.py#L107C1-L108C1) that overwrites the file.

I honestly didnt spend a lot of time thinking through the cleanest way to pass this stuff around so let me know if theres a better approach.

edgarrmondragon:
Thanks for sharing the implementation @pnadolny13!

[...]

PR #3727, opened 2026-08-06
fix: Retain returned OAuth refresh tokens

ryanduguid:
## Summary
- retain a non-empty `refresh_token` returned by an OAuth token endpoint
- preserve the current refresh token when a provider omits one
- add regression coverage for both cases

## Motivation
Some OAuth providers rotate refresh tokens. Without retaining the response value, the next refresh in a long-running process can reuse an invalidated token.

This deliberately updates only the in-memory authenticator; it does not attempt config write-back (tracked separately in #106).

## Verification
- `uv run --no-sync pytest tests/core/rest/test_authenticators.py -q --benchmark-skip` (45 passed)
- `uvx ruff check singer_sdk/authenticators.py tests/core/rest/test_authenticators.py`
- `uvx ruff format --check singer_sdk/authenticators.py tests/core/rest/test_authenticators.py`
- `uv run --no-dev --with ty ty check singer_sdk/authenticators.py`

## Summary by Sourcery

Ensure OAuth authenticators retain valid refresh tokens returned by providers and add regression coverage for refresh token handling.

Bug Fixes:
- Preserve existing OAuth refresh tokens when token endpoint responses omit a refresh_token.
- Retain new non-empty OAuth refresh tokens returned by providers during token refresh.

Tests:
- Add regression tests verifying refresh tokens are updated when returned and preserved when omitted in OAuth token responses.

### DIFF
--- a/singer_sdk/authenticators.py
+++ b/singer_sdk/authenticators.py
@@ -639,8 +639,7 @@ class OAuthAuthenticator(APIAuthenticatorBase):
 
         token_json = token_response.json()
         self.access_token = token_json["access_token"]
-        refresh_token = token_json.get("refresh_token")
-        if refresh_token:
+        if refresh_token := token_json.get("refresh_token"):
             self.refresh_token = refresh_token
         expiration = token_json.get("expires_in", self._default_expiration)
         self.expires_in = int(expiration) if expiration else None
--- a/singer_sdk/authenticators.py
+++ b/singer_sdk/authenticators.py
@@ -640,7 +640,7 @@ class OAuthAuthenticator(APIAuthenticatorBase):
         token_json = token_response.json()
         self.access_token = token_json["access_token"]
         refresh_token = token_json.get("refresh_token")
-        if isinstance(refresh_token, str) and refresh_token:
+        if refresh_token:
             self.refresh_token = refresh_token
         expiration = token_json.get("expires_in", self._default_expiration)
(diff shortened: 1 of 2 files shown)


singer_sdk/authenticators.py
```diff
@@ -639,6 +639,9 @@ def update_access_token(self) -> None:
 
         token_json = token_response.json()
         self.access_token = token_json["access_token"]
+        refresh_token = token_json.get("refresh_token")
+        if isinstance(refresh_token, str) and refresh_token:
+            self.refresh_token = refresh_token
```
edgarrmondragon, line 643: can we skip the instance check and just see retain the token if it's truthy?

ryanduguid: Done in 879fbd9d — now retain a returned refresh token whenever it is truthy. I also ran the focused regression test successfully (2 passed).

singer_sdk/authenticators.py
```diff
@@ -639,6 +639,9 @@ def update_access_token(self) -> None:
 
         token_json = token_response.json()
         self.access_token = token_json["access_token"]
+        refresh_token = token_json.get("refresh_token")
+        if refresh_token:
+            self.refresh_token = refresh_token
```
edgarrmondragon:
```suggestion
        if refresh_token := token_json.get("refresh_token"):
            self.refresh_token = refresh_token
```

PR #3759, opened 2026-08-31
SPAM

### LINKED ISSUE #3694: feat: Expose UUID generation and parsing in stream map expressions
### Feature scope

Taps (catalog, state, tests, etc.)

## Problem

Working with UUIDs in stream map expressions can be useful in specific cases — generating surrogate keys, producing synthetic IDs, or validating that an incoming field is a real UUID. There's no first-class support for this today: the `uuid` module isn't available in the stream map function namespace, and it can't simply be added, since `simpleeval` blocks module objects outright (`FeatureNotAvailable: Sorry, modules are not allowed`).

Some of this can be done today, but only through awkward workarounds:

- **Validation** is possible with pure string manipulation — checking length, hyphen positions, and stripping the hex/hyphen character set — but the resulting expressions are long, hard to read, and easy to get subtly wrong. A canonical-shape check ends up looking like:

```yaml
  __filter__: >-
    len(value) == 36
    and value[8] == '-' and value[13] == '-'
    and value[18] == '-' and value[23] == '-'
    and value.strip('0123456789abcdef-') == ''
```

  It works, but it can't express stricter notions like "is a real v4 UUID," and the intent isn't obvious to a reader.

- **Generation** has no equivalent workaround — there's no way to produce a random UUID from within an expression.

First-class UUID callables would replace the string manipulation with something readable, and make generation possible at all.

## Proposal

Register a small set of UUID callables in the stream map function namespace. At minimum:

- `uuid4()` — generate a random UUID (returned as a `str`), for surrogate/synthetic keys.
- `UUID(value)` — parse/validate a UUID string (`uuid.UUID`), enabling validation and canonicalization.

```python
import uuid

functions = {
    # ...existing entries...
    "uuid4": lambda: str(uuid.uuid4()),
    "UUID": uuid.UUID,
}
```

Returning `uuid4()` as a `str` avoids leaking a `UUID` object into the record (which downstream JSON serialization / target type handling may not expect).

## Example usage

Generate a surrogate key:

```yaml
stream_maps:
  my_stream:
    surrogate_id: uuid4()
```

Validate / filter to real UUIDs (pairs well with a `try_cast`-style helper so bad input doesn't raise):

```yaml
    __filter__: try_cast(value, UUID) is not None
```

Canonicalize an incoming UUID string (lowercased, hyphenated) via round-trip:

```yaml
    id: str(UUID(id))
```

## Design considerations

- **Validation is already possible, just awkwardly** — the string-manipulation approach above works today, so for validation this is largely about ergonomics and correctness; generation, by contrast, has no workaround at all.
- **Module can't be exposed directly** — `simpleeval` rejects module objects, so functionality must be registered as named callables (as above).
- **Return `str`, not `UUID` objects** — keeps records JSON-serializable and consistent with how targets expect string keys.
- **`UUID()` raises on bad input** — on its own it can't be used safely in an expression (no `try`/`except`). It's most useful alongside an exception-handling helper; see the related request for a `try_cast` helper.
- **Determinism** — `uuid4()` is random, so it's non-reproducible across re-runs. If a stable/deterministic ID is desirable, a `uuid5(namespace, name)`-style callable could be considered as a follow-up.

---

Follows [this Slack discussion](https://meltano.slack.com/archives/C069CQNHDNF/p1783571869802079) - would go hand-in-hand with #3693.

🤖 Generated with [Claude](https://claude.ai)

ryanduguid:
## Summary

- expose `UUID()` for parsing and validating UUID values in stream map expressions
- expose `uuid4()` for generating JSON-serializable UUID strings
- document both helpers and cover generation and canonicalization

Closes #3694.

## Testing

- `uvx nox -s tests-3.12`
- `uvx pre-commit run --files singer_sdk/mapper.py tests/core/test_mapper.py docs/stream_maps.md`
- `uvx nox -s typing-3.10 -- singer_sdk/mapper.py`
- `uvx nox -s docs`

## Summary by Sourcery

Expose UUID helpers for validating, canonicalizing, and generating UUID values in stream map expressions.

New Features:
- Expose UUID parsing, validation, canonicalization, and random UUID generation in stream map expressions.

Documentation:
- Document the UUID() and uuid4() stream map helpers.

Tests:
- Add coverage for UUID canonicalization and version 4 UUID generation in stream maps.


ryanduguid:
Fair call on the retitles, and sorry for the noise.

#3759 duplicated #3732, which had been open since 11 August, and it exposed `UUID()` directly after you had said on #3694 in July that you would rather have a `check_uuid` function. I opened it without reading either thread. #3734 and #3735 were unrequested changes to sample code and CI that I closed myself on 26 August without saying why. All three were prepared with agent tooling and I reviewed them less carefully than #3727.

I will not open anything further here unless it starts from an issue labelled `accepting pull requests` and I have asked on the issue first. Thanks for the review time on #3727.

